PR #3, opened 2020-06-15
Actually fix links

bobcatfish:
without committing straight to master O.O


dlorenc:
Merging manually, since we don't have OWNERS yet :)

PR #6, opened 2020-06-16
Initial controller code.

dlorenc:
This stands up a controller using ko. The controller does nothing
right now, other than register watchers on TaskRuns and log
reconciliation events.


font:
@dlorenc Is this mostly leveraged from `tektoncd/pipeline`? Would it be possible to split out the commit into generated and non-generated code?

dlorenc:
> @dlorenc Is this mostly leveraged from `tektoncd/pipeline`? Would it be possible to split out the commit into generated and non-generated code?

So everything outside of vendor/ is non-generated. I think we can add a github config to hide these by default. I'll look into it.

dlorenc:
> So everything outside of vendor/ is non-generated. I think we can add a github config to hide these by default. I'll look into it.

OK, I added the .gitattributes file to mark those as generated. Github should hide those now.

font:
Thanks @dlorenc ! I was thinking for making it easier to review as GitHub won't show the entire diff, only the first 3000 changed files. But it looks like there's only 3 files (GitHub was smart enough to conveniently include them at the top) and the rest is generated.

/approve
/lgtm

PR #9, opened 2020-06-24
Add and wire up the signing interfaces.

dlorenc:
This defines an interface to convert a TaskRun to a payload format. Right
now the only supported format is just native Tekton (a passthrough of the TaskRun).
We can add other formats, like InToto and Grafeas next.

This also adds a few other small functions so we can wire this up to the controller:
- MarkSigned, to mark the TaskRun as signed.
- IsSigned, to tell if the TaskRun has already been signed.

Note: Payloads are not currently stored anywhere.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -16,6 +16,7 @@ package controller
 import (
 	"context"
 
+	"github.com/tektoncd/chains/pkg/signing"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	informers "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/reconciler"
@@ -34,6 +35,20 @@ type Reconciler struct {
 // handleTaskRun handles a changed or created TaskRun.
 // This is the main entrypoint for chains business logic.
 func (r *Reconciler) handleTaskRun(ctx context.Context, tr *v1beta1.TaskRun) error {
+	// Check to make sure the TaskRun is finished.
+	if !tr.IsDone() {
+		r.Logger.Infof("taskrun %s/%s is still running", tr.Namespace, tr.Name)
+		return nil
+	}
+	// Check to see if it has already been signed.
+	if signing.IsSigned(tr) {
+		r.Logger.Infof("taskrun %s/%s has already been signed", tr.Namespace, tr.Name)
+		return nil
+	}
+
+	if err := signing.SignTaskRun(r.Logger, r.PipelineClientSet, tr); err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -59,6 +74,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
 
 	r.Logger.Infof("Sending update for %s/%s (uid %s)", namespace, name, tr.UID)
 
-	// Call the actual handler with the fetched TaskRun.
-	return r.handleTaskRun(ctx, tr)
+	// Call the actual handler with a copy of the fetched TaskRun.
+	return r.handleTaskRun(ctx, tr.DeepCopy())
 }
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -4,11 +4,17 @@ import (
 	"context"
 	"testing"
 
+	"github.com/tektoncd/chains/pkg/signing"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	versioned "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	informers "github.com/tektoncd/pipeline/pkg/client/informers/externalversions/pipeline/v1beta1"
 	fakepipelineclient "github.com/tektoncd/pipeline/pkg/client/injection/client/fake"
 	faketaskruninformer "github.com/tektoncd/pipeline/pkg/client/injection/informers/pipeline/v1beta1/taskrun/fake"
+	"github.com/tektoncd/pipeline/pkg/reconciler"
+	"go.uber.org/zap"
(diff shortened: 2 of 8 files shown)


pkg/signing/signing.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// SignTaskRun signs a TaskRun, and marks it as signed.
+// Use var for mocking
+var SignTaskRun = func(logger *zap.SugaredLogger, ps versioned.Interface, tr *v1beta1.TaskRun) error {
+	// First sign the overall TaskRun with all the configured payloads
+	payloads := generatePayloads(logger, tr)
+	logger.Infof("Generated payloads: %v for %s/%s", payloads, tr.Namespace, tr.Name)
+
+	// TODO: Store those payloads in all the configured storage systems.
```
mpeters, line 56: Do we have a plan for how these storage systems are going to be configured? I imagine it's going to pretty similar to how we pick which payload formats are used as well.

pkg/signing/signing.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Now mark the TaskRun as signed
+	return MarkSigned(tr, ps)
+}
+
+func generatePayloads(logger *zap.SugaredLogger, tr *v1beta1.TaskRun) []interface{} {
```
mpeters, line 62: Is the plan to generate all possible payload types? Or will this be configurable about which ones are needed?

dlorenc: Great question! I imagined it as configurable. I added a section to the design doc on how we might do that: https://docs.google.com/document/d/10MDaivmc-mOcst0zRYWBxQuSQ5BPnjLXN8PEr7hMBC4/edit#

pkg/signing/signing.go
dlorenc: I added a section on how I was imaging this to work in the design doc here: https://docs.google.com/document/d/10MDaivmc-mOcst0zRYWBxQuSQ5BPnjLXN8PEr7hMBC4/edit#

pkg/signing/signing.go
```diff
@@ -0,0 +1,73 @@
...
+	ChainsAnnotation = "chains.tekton.dev/signed"
+)
+
+// IsSigned determines whether a TaskRun has already been signed.
+func IsSigned(tr *v1beta1.TaskRun) bool {
+	signature, ok := tr.ObjectMeta.Annotations[ChainsAnnotation]
+	if !ok {
+		return false
+	}
+	return signature != ""
```
font, line 34: Should we allow any non-empty string to signify that the taskrun is signed or only allow "true" since that's what we explicitly set?

pkg/signing/signing_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}{
+		{
+			name:       "signed",
+			want:       true,
+			annotation: "true",
+		},
+		{
+			name:       "signed with other string",
+			want:       true,
+			annotation: "baz",
```
font, line 72: Should `IsSigned` return true for this case?

pkg/controller/controller_test.go
```diff
@@ -66,3 +72,92 @@ func setupData(ctx context.Context, t *testing.T, trs []*v1beta1.TaskRun) inform
...
+			r := &Reconciler{
+				Base: &reconciler.Base{
+					PipelineClientSet: c,
+				},
+				Logger: logtesting.TestLogger(t),
+			}
+			if err := r.handleTaskRun(ctx, tt.tr); err != nil {
+				t.Errorf("Reconciler.handleTaskRun() error = %v", err)
+			}
+			if m.signed != tt.shouldSign {
```
font, line 140: Any reason not to use `IsSigned` here?

pkg/controller/controller.go
```diff
@@ -34,6 +35,20 @@ type Reconciler struct {
 // handleTaskRun handles a changed or created TaskRun.
 // This is the main entrypoint for chains business logic.
 func (r *Reconciler) handleTaskRun(ctx context.Context, tr *v1beta1.TaskRun) error {
+	// Check to make sure the TaskRun is finished.
+	if !tr.IsDone() {
+		r.Logger.Infof("taskrun %s/%s is still running", tr.Namespace, tr.Name)
+		return nil
+	}
+	// Check to see if it has already been signed.
+	if signing.IsSigned(tr) {
```
font, line 44: Does Tekton have a concept of an immutable `TaskRun`? Since users could still modify the TaskRun to manipulate this annotation after it has been added and after they modify the TaskRun, should we concern ourselves with making sure this is immutable?

dlorenc: Tekton/Kubernetes do not really allow immutability to be enforced. I think we'll need to rely on RBAC here.

pkg/signing/signing.go
dlorenc: Good point. I'll tighten this up for now.

pkg/signing/signing_test.go
dlorenc: Changed with the above change.

pkg/controller/controller_test.go
dlorenc: I mocked out the function that actually does the signing (it would require a fake API server and more boilerplate.

pkg/controller/controller.go
font: I'm not really sure it's worth investigating at this point, but I'm thinking of the use of admission controllers to enforce immutability for certain TaskRuns that require it e.g. ones that have already been signed.

pkg/controller/controller_test.go
font: I was thinking `IsSigned` because it just takes a `TaskRun`. I think this makes more sense in an e2e test, but since we're faking the signing anyway, it just seemed like `IsSigned` could be used here to check the annotations that are already configured in this test's setup, or have the mock signing function add the annotation itself.

PR #10, opened 2020-06-24
Add skelterjohn to the OWNERS file.

dlorenc:
He was in the original chains proposal, but wasn't an org member yet
so couldn't get added.


dlorenc:
Looks like we need to wait for peribolos to sync.

Issue #12, opened 2020-06-29
in-toto integration 


lukehinds:
Placeholder issue for in-toto integration tasks

@SantiagoTorres , hopefully you can see this.

- [x] Await on https://github.com/tektoncd/chains/pull/9 landing first
- [ ] Create simple python package, place onto test pypi
- [ ] Construct tasks and pipeline
- [ ] How will we query the link file?

Will need their own issues.
- [ ] Storage adapter
- [ ] Support more signing keys

dlorenc:
OK, I'm going to get started on a skeleton PR for in-toto formatting next!

dlorenc:
- [ ] Add some glue in chains to detect when a python package is built and sign/upload that to pip

SantiagoTorres:
/reopen

I think this is still going on?

PR #13, opened 2020-07-07
WIP: Add an intoto formatter.

dlorenc:
This translates TaskRuns to an InToto link format. Resources are mapped
to Products/Materials, and the TaskRun information is mapped to Environment.

Ref #12


dlorenc:
This needs tests and documentation still, marking as WIP.

lukehinds:
hey @dlorenc , that was quick!

Is there a reason for importing in-toto wholesale and not as a dependency in go.sum? (I am sure there must be, so curious).

dlorenc:
> Is there a reason for importing in-toto wholesale and not as a dependency in go.sum? (I am sure there must be, so curious).

Good question! There are two main modes to work with go modules, vendoring and non vendoring. We're using vendoring here, so the code still gets committed.

I prefer vendoring, but it's not something every project does.

shibumi:
> Good question! There are two main modes to work with go modules, vendoring and non vendoring. We're using vendoring here, so the code still gets committed. I prefer vendoring, but it's not something every project does.

Vendoring was useful before go.mod, right now it's fully ok to let go mod generate the vendor directory on demand and just push the go.mod and go.sum files to the project.

Btw, I am working on in-toto run support on in-toto-golang. This might be useful for you: https://github.com/in-toto/in-toto-golang/pull/56

I will definitely have a look on this PR here, too :)

shibumi:
@dlorenc  do you know the state of this PR here? Is this PR deprecated?

dlorenc:
> @dlorenc do you know the state of this PR here? Is this PR deprecated?

Yeah, this one is deprecated, it's being replaced by the ITE-6 formatter. The PR hasn't been opened yet, but you can see the branch and discussion in #65. I think @kommendorkapten is going to open it early next week!

kommendorkapten:
@dlorenc @shibumi what Dan said, next week we should have a PR.

PR #16, opened 2020-07-11
Add and wire up a storage interface.

dlorenc:
This adds a storage interface to store signed payloads. Right now
the only supported backend is "Tekton", which stores them in the
annotations field of a Tekton k8s object.

### DIFF
--- a/pkg/signing/signing_test.go
+++ b/pkg/signing/signing_test.go
@@ -14,11 +14,17 @@ limitations under the License.
 package signing
 
 import (
+	"errors"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/chains/pkg/signing/storage"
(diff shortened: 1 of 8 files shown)


pkg/signing/signing_test.go
```diff
@@ -93,3 +99,115 @@ func TestIsSigned(t *testing.T) {
 		})
 	}
 }
+
+func TestTaskRunSigner_SignTaskRun(t *testing.T) {
+
```
jonjohnsonjr, line 104: Extra newline here

pkg/signing/signing_test.go
```diff
@@ -93,3 +99,115 @@ func TestIsSigned(t *testing.T) {
...
+	// SignTaskRun does two main things:
```
jonjohnsonjr, line 105: This is three things

dlorenc: This is why we keep you around.

PR #21, opened 2020-07-20
Actually sign payloads!

dlorenc:
This commit hooks up the secret and our PGP signing library to the reconciler.
It also tweaks the storage backend interface a bit - we need to take two parameters instead of one:
- The signed body (in byte form)
- The signature itself

Since we need to store both of these.

I also added some extra debugging to the e2e tests.


pkg/signing/signing.go
```diff
@@ -66,19 +69,49 @@ var getBackends = storage.InitializeBackends
...
+	// Sign all the payloads with all the signing strategies (right now just pgp)
+	// TODO: Currently, this reads secrets from disk every time.
+	// This could be optimized to instead watch the secret for changes.
+	signer, err := pgp.NewSigner(ts.SecretPath, ts.Logger)
+	if err != nil {
+		return err
+	}
+
+	// Signing the payload objects forces them to be marshalled into bytes.
+	// Use this marshaled form from now on, instead of the object itself.
```
kimsterv, line 85: "Use this marshalled form from now on"

PR #22, opened 2020-07-21
Add a config store that can be used for formats, signing, storage, etc.

dlorenc:
This adds a config store to watch a config map. That config map can contain
keys and data governing how chains should operate. There is currently one
field "formats.enabled-formats", which is not hooked up yet.


pkg/config/store.go
```diff
@@ -0,0 +1,73 @@
...
+	EnabledFormats []string
+}
+
+const (
+	enabledFormatsKey = "formats.enabled-formats"
+)
+
+func parse(data map[string]string) Config {
+	cfg := Config{}
+	if formats, ok := data[enabledFormatsKey]; ok {
```
lukehinds, line 24: Do we want an `else` here to catch any failures, something like the code example for  "comma, ok" idiom https://golang.org/doc/effective_go.html#interface_conversions

dlorenc: I don't think we need one here - that would just indicate the user left this as a default value, which would be equivalent to an empty string or no enabled formats.

I added a comment explaining why we need the check at all - unfortunately strings.Split is a little tricky here.

PR #23, opened 2020-07-21
Add boilerplate for e2e tests.

dlorenc:
This should be enough to get them setup in prow/plumbing.

Ref https://github.com/tektoncd/plumbing/pull/489

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/tektoncd/plumbing v0.0.0-20200615101855-f56d60092af6
 	go.uber.org/zap v1.15.0
 	golang.org/x/crypto v0.0.0-20200323165209-0ec3e9974c59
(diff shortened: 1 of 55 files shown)


dlorenc:
OK https://github.com/tektoncd/plumbing/pull/489 is in, I think it gets auto-applied to our cluster then I can iterate here until there's an empty test running and reporting results.

vdemeester:
> OK [tektoncd/plumbing#489](https://github.com/tektoncd/plumbing/pull/489) is in, I think it gets auto-applied to our cluster then I can iterate here until there's an empty test running and reporting results.

Yep it does, might take a few minutes before applied but it should be there soon

dlorenc:
Yesss! It's alive!

image

dlorenc:
> Why does your editor strip newlines at EOF?

I think it was a copy/paste thing

Issue #27, opened 2020-07-22
Proposal: Config Model


dlorenc:
PR #22 will add a config store we can use to configure the chains system. Here's a rough sketch for what that config might look like.

Roughly in chains, we have a few subsystems:

- Things to watch and sign
- Metadata formats to sign (in-toto, grafeas, etc.)
- Signing strategies (PGP, x509, etc.)
- Storage systems to store signatures in

Some storage systems and metadata formats may cross artifact types, but others won't. Here's my proposal for how to organize config, starting with things to watch and sign.

Things to sign will include taskruns, oci images, binaries, wheels, etc.

```
taskruns.formats.enabled-formats=in-toto,tekton,etc.
taskruns.signing.enabled-strategies=pgp,x509,etc.
taskruns.storage.backends=tekton,gcs,oci,etc.
```

We can then repeat this for the other artifact types:

```
oci.formats.enabled-formats=in-toto
oci.signing.enabled-strategies=pgp
oci.storage.backends=oci
```

to indicate we want to sign OCI images using PGP, with metadata in the in-toto format, and store that in an OCI registry.

Some of these may require their own custom sections as well:

```
storage.gcs.location=gs://foo/bar
```

would indicate where we want to store signatures in GCS.

For OCI, we could either store signatures next to the image we sign (notary v2 style), or we'd want to specify a repository to store signatures in (for other types):

```
storage.oci.repository=gcr.io/foo/bar
```

dlorenc:
For things like wheels on pip, we'll have the Task/Run upload the wheel itself. The chains system will then generate and store a signature that will also be stored in pypi. That could look like:

```
wheels.storage.backends=pypi
storage.pypy.credentials-secret=$name of secret with creds for pypi
```

PR #29, opened 2020-07-24
More config store improvements.

dlorenc:
Nest the keys following the outline in https://github.com/tektoncd/chains/issues/27
Add a second config list for enabled storage backends.

These will be hooked up next.

### DIFF
--- a/pkg/config/store.go
+++ b/pkg/config/store.go
@@ -12,22 +12,55 @@ import (
 )
 
 type Config struct {
-	EnabledFormats []string
+	Artifacts Artifacts
+}
+
+type Artifacts struct {
+	TaskRuns TaskRuns
+}
+
+type TaskRuns struct {
+	Formats         Formats
+	StorageBackends StorageBackends
+}
+
+type Formats struct {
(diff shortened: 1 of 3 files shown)


pkg/config/store.go
```diff
@@ -12,22 +12,50 @@ import (
...
+	taskrunEnabledStoragesKey = "artifacts.taskrun.storage.enabled"
 )
 
 func parse(data map[string]string) Config {
 	cfg := Config{}
-	if formats := data[enabledFormatsKey]; formats != "" {
+
+	// Start with artifact-specific configs
+	// TaskRuns
+	setStringSlice(&cfg.Artifacts.TaskRuns.Formats.EnabledFormats, taskrunEnabledFormatsKey, data)
```
dekkagaijin, line 45: `cfg.Artifacts.TaskRuns.Formats.EnabledFormats = stringSlice(taskrunEnabledFormatsKey, data)`?

dlorenc: D'oh. Much cleaner.

PR #31, opened 2020-07-29
Readme documentations updates and additions

lukehinds:
This change fixes a small typo around the shell prompt and adds
some futher tweaks based on my onw testing / run through

Note that line 81 requires the root namespace, e.g:

`kubectl get taskrun.tekton.dev/taskrun home-is-set-rwhzs`


dlorenc:
Thanks!

PR #33, opened 2020-07-31
Hook config up!

dlorenc:
This commit hooks up the configstore object for backends and payload formats!

The only supported ones are still Tekton, but this will make it easier to add more.


dlorenc:
/hold

I think I'd like to get in one more refactor before this goes in.

dlorenc:
Closing this out in favor of #39

PR #47, opened 2020-08-26
Add an e2e test for OCI images.

dlorenc:


dlorenc:
This needs the nightly pipeline release in #48

PR #50, opened 2020-09-02
Store signatures in an OCI registry.

dlorenc:


pkg/signing/storage/oci/oci.go
```diff
@@ -0,0 +1,131 @@
...
+type Backend struct {
+	logger *zap.SugaredLogger
+	tr     *v1beta1.TaskRun
+	cfg    config.Config
+	auth   remote.Option
+}
+
+// NewStorageBackend returns a new OCI StorageBackend that stores signatures in an OCI registry
+func NewStorageBackend(logger *zap.SugaredLogger, tr *v1beta1.TaskRun, cfg config.Config) (*Backend, error) {
+	kc, err := k8schain.NewInCluster(k8schain.Options{Namespace: "tekton-pipelines", ServiceAccountName: "tekton-pipelines-controller"})
```
dlorenc, line 51: I think you can get the namespace from `system.Namespace()`

PR #53, opened 2020-12-10
Add x509 signing interfaces and tests.

dlorenc:
These are not hooked up yet.

### DIFF
--- a/pkg/signing/x509/x509.go
+++ b/pkg/signing/x509/x509.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2020 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


pkg/signing/x509/x509.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Signer exposes methods to sign payloads using x509
+type Signer struct {
+	key    interface{}
+	logger *zap.SugaredLogger
+}
+
+// NewSigner returns a configured Signer
+func NewSigner(secretPath string, logger *zap.SugaredLogger) (*Signer, error) {
+	privateKeyPath := filepath.Join(secretPath, "x509.pem")
```
lukehinds, line 40: Any reason for not passing in the file name as a variable? I am sure you have a good reason, but thought it worth checking.

lukehinds (approved):
small comment, but code looks good

pkg/signing/x509/x509.go
dlorenc: Good question, and it's a little hard to see because this isn't fully hooked up!

This will actually be set by a Kubernetes secret mounted in a specific location with a specific syntax, so the path is fully under our control. Users will basically set whatever key they want to use in the k8s secret at a specific location, then this part can just be a constant.

The next PR will hopefully make this more obvious.

PR #66, opened 2021-05-03
Add a documentdb storage backend.

dlorenc:
This allows for flexible storage and querying of fully-nested types. Full lookups
on each property in the ITE-6 provenance payload, for example.


dlorenc:
You can query all the properties of the TaskRun or ITE-6 format with this:
image

pkg/chains/storage/docdb/docdb.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// StorePayload implements the Payloader interface.
+func (b *Backend) StorePayload(signed []byte, signature string, key string) error {
+	// We need two object names: the signature and the payload. We want to make these unique to the UID, but easy to find based on the
+	// name/namespace as well.
+	// $bucket/taskrun-$namespace-$name-$uid/$key.signature
+	// $bucket/taskrun-$namespace-$name-$uid/$key.payload
+	root := fmt.Sprintf("taskrun-%s-%s-%s", b.tr.Namespace, b.tr.Name, b.tr.UID)
+	sigName := path.Join(root, fmt.Sprintf("%s.signature", key))
```
kommendorkapten, line 75: `sigName` is never used except for the log message. In the signed document, `Name` is set to `key` (which is the last part of `sigName`), that's not where `sigName` was meant to go?

kommendorkapten: Also two names are mentioned in the comment, but only one (`sigName`) is generated.

dlorenc: Thanks! copy/paste mistake from the GCS storage one :)

priyawadhwa (approved):
LGTM! The annotations in the README [here](https://github.com/tektoncd/chains/blob/main/README.md#usage) just need to be updated to include the UID

font (approved):
Nice!

/lgtm

Issue #62, opened 2021-04-23
cosign signature spec compat


dlorenc:
https://github.com/sigstore/cosign/blob/main/SPEC.md

When we sign OCI artifacts with "simplesigning" format we should make sure it works with this spec. I think it does already, should just double check and add a test.

PR #70, opened 2021-05-05
Replace OCI storage backend with cosign upload

priyawadhwa:
This change uses cosign.Upload for the OCI storage backend. The only big change is that I **removed** support for storing TaskRuns in OCI registries (basically it'll return an error if the thing it's trying to store isn't an image). But, I can add it back in if people think that's better!

Also added an integration test that pushes to an insecure registry in the cluster

this will ensure & close https://github.com/tektoncd/chains/issues/62


dlorenc (approved):

dlorenc:
The tekton/sigstore synergy is beautiful.

PR #76, opened 2021-05-10
Add priyawadhwa@ as an approver

priyawadhwa:


dlorenc:
I think you have to get the org one in first

priyawadhwa:
the bot doesn't seem to be picking up that I'm in the org now, opened https://github.com/tektoncd/chains/pull/78 to replace this one!

Issue #73, opened 2021-05-07
Run the integration tests that require a registry


priyawadhwa:
Right now, we're skipping these tests because we can't access GCR. Some options we have to fix this:
1. Use an insecure registry running in the cluster (this can only be accessed from pods because the LoadBalancer doesn't work)
2. Figure out a way to inject a secret with GCR credentials into the test cluster

PR #80, opened 2021-05-10
e2e testing: port forward registry for external access

priyawadhwa:
This will allow us to start running tests the need access to a registry! This one is still an insecure registry so it would be nice to eventually figure out how to plumb in credentials to use GCR

closes https://github.com/tektoncd/chains/issues/73

### DIFF
--- a/test/clients.go
+++ b/test/clients.go
@@ -34,12 +34,12 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


afrittoli:
@priyawadhwa Thank you for your PR! 
I'm not sure if the `chains` project has a different policy, but in the `tektoncd` org we do not automatically squash commits before merge, so I would ask you if you could do so to keep a clean history 🙏  You might want to rebase too, as that might unlock `tide`.
Thank you for your patience :)

Issue #79, opened 2021-05-10
Support cosign generated private keys for signing


priyawadhwa:
This would probably be a type of x509 signing, which supports the COSIGN ENCRYPTED PRIVATE KEY type

dlorenc:
+1! Nice.

PR #82, opened 2021-05-11
Add support for signing with cosign generated keys

priyawadhwa:
And adds an integration test for it!

closes https://github.com/tektoncd/chains/issues/79


test/testdata/cosign.password
```diff
@@ -0,0 +1 @@
+p
```
dlorenc, line 1: lol

dlorenc (approved):

Issue #71, opened 2021-05-07
Better defaults/validation for the configmap


dlorenc:
We should probably clean up the configmap code a bit, the defaults are set in yaml which means they get overwritten every time you deploy.

We can probably do this better with actual defaults, and maybe even use reflection!

PR #83, opened 2021-05-11
Add defaults and validation to configmap

priyawadhwa:
closes https://github.com/tektoncd/chains/issues/71


dlorenc (commented):
Should we also drop (or comment out) the config values in the deployment.yaml?

priyawadhwa:
Good idea! Commented them out

PR #88, opened 2021-05-17
in-toto attestations (ITE-6) as a formatter.

kommendorkapten:
This PR implements the basic support for https://github.com/tektoncd/chains/issues/65. 
The formatter uses type hinting to be able to extract information to prepare the in-toto attestation.

Issue #65, opened 2021-04-30
ITE-6 as a formatter


dlorenc:
https://github.com/tektoncd/chains/compare/main...solarwinds:ite6 is a draft branch.

The ITE-6 code is here: https://github.com/in-toto/in-toto-golang/tree/ite-6

dlorenc:
@kommendorkapten - are you working on this still from the draft branch?

dlorenc:
Hey @kommendatorkapten - I rebased your commit then updated it to the latest ite-6 formats (from your other commit):
https://github.com/in-toto/in-toto-golang/tree/ite-6


My branch is here: https://github.com/dlorenc/chains-1/pull/new/ite6

It's hard to merge back into yours because I rebased...

dlorenc:
Early feedback on the ITE-6 draft branch:

* use PURL to encode references to container images
* The build ID should be configurable when chains is installed/setup, not use the pod name

dlorenc:
There was a bigger question around the "entrypoint" field too we still need to work out.

PR #89, opened 2021-05-17
in-toto attestations (ITE-6) as a formatter.

kommendorkapten:
This PR deprecates https://github.com/tektoncd/chains/pull/88. 

This PR implements the basic support for https://github.com/tektoncd/chains/issues/65.
The formatter uses type hinting to be able to extract information to prepare the in-toto attestation.

### DIFF
--- a/pkg/chains/formats/intotoite6/intotoite6.go
+++ b/pkg/chains/formats/intotoite6/intotoite6.go
@@ -1,0 +1,353 @@
+package intotoite6
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/tektoncd/chains/pkg/chains/formats"
+	"github.com/tektoncd/chains/pkg/config"
+
+	"github.com/in-toto/in-toto-golang/in_toto"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+
+	"github.com/google/go-containerregistry/pkg/name"
+)
+
+const (
+	tektonID = "https://tekton.dev/attestations/chains@v1"
+)
+
+var ErrNoBuilderID = errors.New("no builder id configured")
+
+const (
+	commitParam        = "CHAINS-GIT_COMMIT"
+	urlParam           = "CHAINS-GIT_URL"
+	ociDigestResult    = "IMAGE_DIGEST"
+	chainsDigestSuffix = "_DIGEST"
+)
+
+type artifactType int
+
+const (
+	artifactTypeUnknown artifactType = iota
+	artifactTypeOCI
+)
+
+type InTotoIte6 struct {
+	builderID string
+}
+
+type artifactResult struct {
+	ParamName  string
+	ResultName string
+	Type       artifactType
+}
+
+type vcsInfo struct {
+	Commit string
+	URL    string
+}
+
+func NewFormatter(cfg config.Config) (formats.Payloader, error) {
+	if cfg.Builder.ID == "" {
+		return nil, ErrNoBuilderID
+	}
+
+	return &InTotoIte6{builderID: cfg.Builder.ID}, nil
+}
+
+func (i *InTotoIte6) CreatePayload(obj interface{}) (interface{}, error) {
+	var tr *v1beta1.TaskRun
+
+	switch v := obj.(type) {
+	case *v1beta1.TaskRun:
+		tr = v
+	default:
+		return nil, fmt.Errorf("unsupported type: %s", v)
+
+	}
+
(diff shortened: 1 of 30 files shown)


kommendorkapten:
I'm working on how to proceed with the CLA with my Employer.

dlorenc:
/ok-to-test
Yay!

lukehinds:
Is this ready for review now @kommendorkapten , or still a WIP?

kommendorkapten:
@lukehinds It's ready for review. Just consider that the functionality provided is MVP. There are some fields in the in-toto attestation that we don't have a good way of populating now based from the information seen in the `TaskRun` struct. Expect follow up work to improve what fields we can populate (e.g when "Hermekton" is finished, there are a set of attributes in the attestation related to hermeticity in builds we should populate).

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -0,0 +1,351 @@
...
+			Type:          in_toto.StatementInTotoV01,
+			PredicateType: in_toto.PredicateProvenanceV01,
+		},
+		Predicate: in_toto.ProvenancePredicate{
+			Builder: in_toto.ProvenanceBuilder{
+				ID: i.builderID,
+			},
+			Recipe: in_toto.ProvenanceRecipe{
+				Type:       tektonID,
+				EntryPoint: tr.Spec.TaskRef.Name,
```
priyawadhwa, line 96: Do you think it would be useful to include `Arguments` and `Environment` to the recipe as well? I'm not sure if we need people to be able to rebuild the TaskRun from the annotation, but we'd probably want them in there if that's the goal.

priyawadhwa (commented):
Hey @kommendorkapten this looks great!

dlorenc:
This is awesome! I'm excited to get this in!

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// Here we translate a Tekton TaskRun into an InToto ite6 attestation.
+	// At a high leevel, the  mapping looks roughly like:
+	// Podname -> Builder.Id
+	// Results with name *_DIGEST -> Subject
+	// Step containers -> Materials
+	// Params with name CHAINS-GIT_* -> Materials and recipe.materials
+	// tekton-chains -> Recipe.type
+	// Taskname -> Recipe.entry_point
+	att := in_toto.ProvenanceStatement{
```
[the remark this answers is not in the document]
kommendorkapten: Fixed!.

pkg/config/store.go
```diff
@@ -137,6 +145,9 @@ func parse(data map[string]string) Config {
 
 	cfg.Signers.KMS.KMSRef = valueOrDefault(kmsSignerKMSRef, data)
 
+	// Build config
+	cfg.Builder.ID = data[builderIDKey]
```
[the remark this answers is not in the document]
kommendorkapten: Fixed.

pkg/chains/formats/intotoite6/intotoite6.go
kommendorkapten: Great comment. For `Environment` I think we should skip it now, as by the in-toto spec states that any included environment value should be _only_ the part of the "build API", and from Chain's perspective, it's hard to understand that in general (such semantic would be defined in the tasks which I don't know of a good and reliable way for Chains to interpret).

For arguments it's a slightly similar story, what goes in the `Arguments` value is "external input that influenced the build", and better would be to rely on the `Entrypoint` alone (which the spec also recommends). 
The problem with the PR in the current shape as the value for the `Entrypoint` is not fully correct, currently it sets the value to the TaskName.

So with that in mind, my wish would be to have an annotation or some type hinting that would allow the task to specify the _real_ arguments to the build, and the `Recipe.Type`. As an example, we have a `maven` task, that accepts parameter `GOALS`, where `GOALS` might be `[install, -DskipTests]`. And with that we should produce an attestation with `Recipe.Type: maven`, `Entrypoint`: `install -DskipTests` or similar, but I don't have spent too much time thinking of such a solution yet. But I will soon :)

priyawadhwa: SGTM, thanks for explaining!

priyawadhwa:
Hey @kommendorkapten this looks great! Would you mind squashing the commits, and then we can merge?

priyawadhwa (approved):

dlorenc:
It's happening!

dlorenc:
You might still need to squash this all the way down to one commit to get it to merge

PR #94, opened 2021-05-19
Add kaniko Task and TaskRun example

priyawadhwa:
I figured it would be nice to have a kaniko task that works with Chains as an example in the repo, and for any blog posts/tutorials we may write.

This task basically:
- Creates a dockerfile as the first step
- builds + pushes the image with kaniko
- writes the IMAGE_URL so that Chains picks up the image that was built and signs it

It would be nice if we added the IMAGE_URL to the official kaniko task as well so that it would just work with chains


dlorenc (approved):
🚀 this is awesome

Issue #75, opened 2021-05-07
Figure out ITE-5/SSL Signing Spec integration


dlorenc:
See here: https://github.com/in-toto/in-toto-golang/pull/101

When generating ITE-6 payloads we can use this format (and the JSON wrapper) for signatures.

It might mean we have to change the interfaces around a bit, but now's the time!

dlorenc:
https://github.com/secure-systems-lab/signing-spec/blob/master/protocol.md

PR #109, opened 2021-05-30
Support the ITE-5 signing spec

dlorenc:
This is now on by default for the intoto payloads.

Ref #75

### DIFF
--- a/pkg/chains/signing/wrap.go
+++ b/pkg/chains/signing/wrap.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2021 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package signing
+
+import (
+	"context"
+	"crypto"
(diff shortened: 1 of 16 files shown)


pkg/chains/signing/x509/x509.go
```diff
@@ -37,18 +41,70 @@ type Signer struct {
...
+type sslSigner struct {
+	s *ssl.EnvelopeSigner
+}
+
+func (s *sslSigner) PublicKey(ctx context.Context) (crypto.PublicKey, error) {
+	panic("unimplemented")
+}
+
+func (s *sslSigner) Sign(ctx context.Context, payload []byte) ([]byte, []byte, error) {
+	env, err := s.s.SignPayload("application/vnd.in-toto+json", payload)
```
kommendorkapten, line 99: The payload type is defined in in-toto here: https://github.com/in-toto/in-toto-golang/blob/ite-6/in_toto/model.go#L45 to avoid raw strings.

pkg/chains/signing/wrap.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (w *sslAdapter) Verify(keyID string, data, sig []byte) (bool, error) {
+	panic("unimplemented")
+}
+
+// sslSigner converts the EnvelopeSigners back into our types, after wrapping.
+type sslSigner struct {
+	wrapper *ssl.EnvelopeSigner
+	typ     string
```
priyawadhwa, line 64: nit: should this be `type`?

priyawadhwa (approved):
looks great! just one nit

pkg/chains/signing/wrap.go
dlorenc: Hah, no this is intentional to annoy Jon and because type is a reserved word :)

PR #111, opened 2021-06-03
Set up examples directory for testing taskruns

priyawadhwa:
Sets up an `examples/taskruns` directory that we can easily add in new tests for -- right now `TestExamples` only validates that the intoto attestation is correct, but I was planning to include other checks in later.


docs/config.md
```diff
@@ -25,7 +25,7 @@ Supported keys include (scroll right for defaults):
 
 | Key | Description | Supported Values | Default |
 | --- | --- | --- | --- |
-| `artifacts.taskrun.format` | The format to store `TaskRun` payloads in. | `tekton` | `tekton` |
+| `artifacts.taskrun.format` | The format to store `TaskRun` payloads in. | `tekton`, `in-toto-ite6` | `tekton` |
```
dlorenc, line 28: Maybe just call it "in-toto?

priyawadhwa: 👍🏽

PR #119, opened 2021-06-08
WIP: Fulcio signing in Chains.

dlorenc:


dlorenc:
TODO:

- [ ] make the auth mechanism configurable (this is hardcoded to use GCP identity tokens for now). we should at least support SPIFFE ones as well
- [ ] tests
- [ ] maybe cache certs and refresh on demand, rather than every time?
- [ ] Timestamping/tlogs

priyawadhwa:
merged in #147!

PR #121, opened 2021-06-15
Transparency log uploads in chains!

dlorenc:
TODO:

- [ ] More docs!
- [ ] ITE-6 integration
- [ ] Support PGP as well

Signed-off-by: Dan Lorenc <dlorenc@google.com>


pkg/chains/signing.go
```diff
@@ -224,3 +244,41 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
 	// Now mark the TaskRun as signed
 	return MarkSigned(tr, ts.Pipelineclientset)
 }
+
+type rekor struct {
```
priyawadhwa, line 248: We could move the rekor stuff into `pkg/chains/rekor` so it's easy to find

priyawadhwa: or, actually, if that's annoying for testing maybe just move it into a `rekor.go` file!

pkg/chains/signing_test.go
```diff
@@ -201,7 +204,74 @@ func TestTaskRunSigner_SignTaskRun(t *testing.T) {
...
+		},
+	}
+	if _, err := ps.TektonV1beta1().TaskRuns(tr.Namespace).Create(ctx, tr2, metav1.CreateOptions{}); err != nil {
+		t.Errorf("error creating fake taskrun: %v", err)
+	}
+	if err := ts.SignTaskRun(ctx, tr2); err != nil {
+		t.Errorf("TaskRunSigner.SignTaskRun() error = %v", err)
+	}
+
+	if len(rekor.entries) != 1 {
```
priyawadhwa, line 269: i think i'm missing something here, but shouldn't this still be 0 since the `pgp` signer was configured above??

pkg/chains/signing.go
dlorenc: Done!

pkg/chains/signing_test.go
dlorenc: Heh, there was some trickery. It should be all correct now.

priyawadhwa (approved):
Nice!!

PR #122, opened 2021-06-15
Update text ʕ•́ᴥ•̀ʔ

bobcatfish:
I read this and thought there was some missing cosign command that we'd
run but looks like it's just a typo and we do want to verify with gpg

lemme know if that's not a good link for gpg! :D


priyawadhwa (approved):
TY for fixing this!

PR #123, opened 2021-06-16
Add SPIRE verification to chains

priyawadhwa:
I might write docs for this later, once we've figured out how we want to provide the entrypointer image with SPIRE support -- I don't think anyone else can really use this right now.


dlorenc:
Do you want to keep this on a feature branch or something until we figure out the right strategy for experimenting with the entrypointer?

priyawadhwa:
SGTM!

dlorenc (approved):

PR #137, opened 2021-07-08
Fix flaky unit test

priyawadhwa:
We parse through a map when getting subjects, so the order isn't always the same.

Sort the list before we compare it against expected output.


priyawadhwa:
/test pull-tekton-chains-integration-tests
--


<br class="Apple-interchange-newline">

PR #138, opened 2021-07-09
Structure the reconciler in a more consistent fashion.

mattmoor:
The general layout followed by Knative-style controllers is to have:
  `pkg/reconciler/{resource name}/`

Then within that the reconciler logic lives within `{resource name}.go` and the logic to wrap that in a `controller.Impl` lives within `controller.go`

This PR should be purely organizational without any semantic changes.

/assign @dlorenc 
/kind cleanup

### DIFF
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -1,4 +1,17 @@
-package controller
+/*
+Copyright 2020 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


mattmoor:
/retest

If this fails again, I'm going to switch the `reflect.Equal` check to `cmp.Diff` so it has a better error message, but otherwise I'm doing to try and keep this semantically small.

mattmoor:
Passed on retry 🎉

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1,4 +1,17 @@
-package controller
+/*
```
priyawadhwa, line 1: nice catch!

priyawadhwa (approved):
Thanks for fixing this @mattmoor !

PR #139, opened 2021-07-09
Migrate chains to use genreconciler

mattmoor:
This leverages `genreconciler` to dramatically reduce the boilerplate needed for tekton/chains and dramatically increase the code coverage.

Much of the churn in vendor is from `go mod tidy && go mod vendor`, which I ran to pull in a few previously pruned packages from `knative.dev/pkg`, and it seems to have drifted a bit.

WIP until the baseline change: https://github.com/tektoncd/chains/pull/138 lands.
/assign @dlorenc 
/kind cleanup

PR #140, opened 2021-07-09
Enable HA

mattmoor:
Now that this is on genreconciler, it should be safe to enable HA.

WIP until https://github.com/tektoncd/chains/pull/139 lands.


mattmoor:
I'm like 90% sure this is because the controller doesn't have "lease" permissions, but that's a wild guess.  I'll dig in.

mattmoor:
Bingo:
```
User "system:serviceaccount:tekton-chains:tekton-chains-controller" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "tekton-chains"
```

mattmoor:
Alright passing, just needs another `/lgtm` @dlorenc

PR #150, opened 2021-07-21
Migrate the ConfigStore to use standard infrastructure.

mattmoor:
Knative has a ton of utilities for managing config storage, watching configmaps, and snapshotting those configmaps onto context during each reconciliation.

This is really just a first chunk, but migrates the bulk of the config package to support the standard interfaces, use the libraries for validation, and leverage the shared configmap watching libraries.

In a subsequent change, it would make the most sense to have the signer extract the config store from the context it is passed instead of having its own reference, but this change has grown overly large already.

/kind cleanup

cc @priyawadhwa @dlorenc


mattmoor:
Not sure what's going on here with go-license 😞

mattmoor:
I pushed the rest of the change here as a separate commit, but need to sort out the go-license things still, which I'll probably need some pointers on.

mattmoor:
Thanks to @dlorenc for spotting the missing newline in `./hack/tools.go` 🤦

mattmoor:
I screwed something up, I added and removed the follow up commit here, and I want to clean that up.

mattmoor:
ok it should be fixed.

mattmoor:
Finally green 😅

PR #159, opened 2021-07-24
Drop the config-logging volume mount.

mattmoor:
This is leftover cruft from older versions of sharedmain that read the initial logging configuration from a volume.

/kind cleanup

/assign @priyawadhwa @dlorenc

### DIFF
--- a/config/100-deployment.yaml
+++ b/config/100-deployment.yaml
@@ -73,8 +73,6 @@ spec:
       - name: tekton-chains-controller
         image: ko://github.com/tektoncd/chains/cmd/controller
         volumeMounts:
-        - name: config-logging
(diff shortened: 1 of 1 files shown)


priyawadhwa (commented):
should we also remove [config-logging.yaml](https://github.com/tektoncd/chains/blob/420adfcdf225326605f2b2c2264b42a2f7b86e4e/config/config-logging.yaml)?

mattmoor:
@priyawadhwa no, it's pulled in via the configmap watcher.

priyawadhwa (approved):

priyawadhwa:
Thanks for explaining!

PR #160, opened 2021-07-24
Fix format directive.

mattmoor:
I noticed in the logs that the `%v` wasn't being resolved (it is superfluous) so this drops it.

/kind cleanup

/assign @priyawadhwa


priyawadhwa (approved):
thank you!

Issue #161, opened 2021-07-24
NewStorageBackend has a smell


mattmoor:
I'm talking about this code in particular:
https://github.com/tektoncd/chains/blob/288242f3ca5ff100eca254192ca715e48394614c/pkg/chains/storage/oci/oci.go#L50-L51

There are two things about this that are bugging me:
1. `NewInCluster` should be avoided, and this should probably be calling `New` with a kube client passed through (this will also make testing easier because you can pass a fake client though).
2. Hard-coding the service account name (`"tekton-chains-controller"`) feels wrong to me for a variety of reasons (what if I want to reconfigure things to run differently?  Why is this using the controller's SA vs. the task's SA?)

Figured I'd open an issue for discussion rather than just slinging a PR for this one.

cc @dlorenc @priyawadhwa

dlorenc:
Not sure I understand the second part - are you saying the chains controller would assume the identity of the task in order to store the image info with impersonation?

mattmoor:
@dlorenc the second item is talking about a bunch of things.  Impersonating the identity of the task is one way to satisfy my concerns (which is the last piece of that bullet).

As far as the "reconfiguration" bit, `mink` is an example where I've done this for much of knative and tekton by folding most things into a single controller process / identity (it would make more sense to use the mink controller identity in that context, but that's not well supported given the hardcoding).  I have other cases in mind as well which are more extreme, but understanding the motivation for which identity is used here might help me reason about things.  Is there a reason this is distinct from the task's identity?

dlorenc:
In general we assume chains has more/different permissions from the task itself, but for this specific case where the signatures are stored directly next to the images in the same repository, the task identity should be fine.

mattmoor:
Would y'all be cool if I put together a PR that:
1. Plumbed through a kube client (New vs. NewInCluster), and
2. Used the TaskRun's identity instead of the chains controller's identity.

Not sure who all would need to sign off on that change.  cc @dlorenc @priyawadhwa

dlorenc:
Go for it! Let's just make sure to document that change in the release notes and README, it will probably make things easier to configure overall but it means people should be able to drop permissions from the chains service account which is a nice benefit.

PR #162, opened 2021-07-26
Plumb `kubernetes.Interface` through storage backends.

mattmoor:
Rather than synthesizing a `k8schain.NewInCluster`, use `k8schain.New` and plumb through a `kubernetes.Interface` from our context injection logic.

In practice this should be roughly semantically equivalent, although it will have some small side effects like respecting the shared `rest.Config` for things like client-side rate-limiting.  The effects of this are probably all desirable and unlikely to be consequential to any existing usage.

Related: https://github.com/tektoncd/chains/issues/161


mattmoor:
cc @dlorenc @priyawadhwa 

This is mostly cleanup.  In a subsequent change I'll change the service account this uses and reflect that in release-notes.

PR #163, opened 2021-07-26
Change the ServiceAccount identity used for OCI storage backend.

mattmoor:
Rather than hardcoding the chains controller as the service account identity used for OCI-based signing, use the task's service account identity.

This should just simplify things because users are already configuring the Task with credentials to push the image, but currently need to grant access to the chains controller as well.

Fixes: #161

```release-notes
The chains controller now uses the taskrun's serviceAccountName for credentials when using OCI storage, instead of the chains controller's service account.
```

WIP until https://github.com/tektoncd/chains/pull/162 lands


mattmoor:
cc @priyawadhwa @dlorenc 

Not sure if I did the release-notes bit right for this repo, so I'd appreciate a spot-check!

priyawadhwa (approved):
LGTM!

PR #179, opened 2021-08-05
Enable dependabot

naveensrinivasan:
Enable dependabot for notification when one of the dependencies has a vulnerability.

This is just the config file. The setting has to be enabled in the Security setting within the repository.


priyawadhwa (approved):

priyawadhwa:
Thank you for your contribution!  😄

PR #180, opened 2021-08-05
Fixed missing dependency and shell lint check

naveensrinivasan:
* Fixed the missing dependency go-licenses
   * Fixed the shell lint check


tools.go
```diff
@@ -0,0 +1,19 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tools
+
+import (
+	_ "github.com/google/go-licenses/licenses"
```
priyawadhwa, line 18: I think we can put this in this file instead:

https://github.com/tektoncd/chains/blob/main/hack/tools.go

naveensrinivasan: @priyawadhwa  Thanks moved it.

naveensrinivasan:
ping @priyawadhwa, let me know if something else is missing. Thanks

PR #192, opened 2021-08-13
Add Chains API service

priyawadhwa:
This PR adds an API server to Chains, which will be responsible for receiving large pieces of data from TaskRuns (e.g. SPIRE SVIDs, signatures, and perhaps SBOMs).

The grpc files can be generated via:

```
./hack/proto/generate-proto.sh
```

The server is specified with a proto/GRPC, but also works across HTTP.

TODO:
- right now data is stored in the container itself, which isn't persistent. i'll add in support for our other storage backends in another PR since it'll require a bigger refactor.
- add in a test to make sure the client generated code works
- merge this into the spire branch at first?


pkg/server/entry.go
```diff
@@ -0,0 +1,58 @@
...
+
+func addEntry(e *proto.Entry) error {
+	log.Printf("Adding entry %v...", e)
+	contents, err := json.Marshal(e)
+	if err != nil {
+		return errors.Wrap(err, "marshal binary")
+	}
+	f := path(e.Uid)
+	fmt.Println(f)
+	if err := ioutil.WriteFile(f, contents, 0644); err != nil {
```
naveensrinivasan, line 49: The `ioutil.WriteFile` makes it hard to unit test. Could this be instead a https://pkg.go.dev/io#Writer? which would be easier to mock and test?

pkg/server/entry.go
```diff
@@ -0,0 +1,58 @@
...
+		return errors.Wrap(err, "writing file")
+	}
+	log.Print("Successfully added entry...")
+	return nil
+}
+
+func path(uid string) string {
```
naveensrinivasan, line 56: Could this be inlined instead of a function?

pkg/server/server.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/tektoncd/chains/proto"
+	"google.golang.org/grpc"
+)
+
+var (
+	cfg Config
+)
+
+const (
+	port = 9000
```
naveensrinivasan, line 34: The `port` could be part of the `Config` with default's being `9000` instead of being in a `const`

hack/proto/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM golang:1.16 AS generate-files
```
naveensrinivasan, line 1: It is recommended to pin dependencies based on `SHA` https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies instead of `tags` because tags can be moved.

pkg/server/server.go
```diff
@@ -0,0 +1,96 @@
...
+			otherHandler.ServeHTTP(w, r)
+		}
+	})
+}
+
+func Start(ctx context.Context, config Config) error {
+	gwmux := runtime.NewServeMux()
+	cfg = config
+
+	proto.RegisterChainsServiceHandlerServer(ctx, gwmux, &server{})
```
naveensrinivasan, line 69: Error return value of `proto.RegisterChainsServiceHandlerServer` is not checked

pkg/server/entry.go
```diff
@@ -0,0 +1,58 @@
...
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/tektoncd/chains/proto"
+)
+
+func getEntry(uid string) (*proto.Entry, error) {
+	log.Printf("Getting entry with uid %s...", uid)
+	f := path(uid)
+	contents, err := ioutil.ReadFile(f)
```
naveensrinivasan, line 30: The `ioutil.ReadFile` makes it hard to unit test. Could this instead be https://pkg.go.dev/io#Reader?

pkg/server/entry.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func addEntry(e *proto.Entry) error {
+	log.Printf("Adding entry %v...", e)
+	contents, err := json.Marshal(e)
+	if err != nil {
+		return errors.Wrap(err, "marshal binary")
+	}
+	f := path(e.Uid)
+	fmt.Println(f)
```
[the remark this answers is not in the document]
priyawadhwa: leftover from debugging, thanks for catching that!

pkg/server/entry.go
priyawadhwa: i added in a unit test which shouldn't require changing this!

pkg/server/entry.go
priyawadhwa: it could, but since it's used in two different functions i figured i'd separate it out!

pkg/server/entry.go
priyawadhwa: resolved as i've added the unit test :)

dlorenc:
+1 on merging into the spire branch at first!

dlorenc:
LGTM for merging into the prototype branch

PR #193, opened 2021-08-14
Add label dependabot PRs ok-to-test

naveensrinivasan:
* Apply this label in any case so we can also pre-add it during PR creation


priyawadhwa (approved):
thank you!

PR #194, opened 2021-08-14
Codeql Check

naveensrinivasan:
[Codeql](https://codeql.github.com/docs/) check for the codebase.


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,38 @@
+name: "CodeQL"
+on:
+  push:
```
dlorenc, line 3: I actually haven't found codeql to be useful, it's never found an issue on the repos where I've used it, but it consumes an enormous amount of resources. If we include it at all I'd rather have it post-merge only on the cron.

naveensrinivasan: Sounds good. I will change it to merge.

naveensrinivasan: Removed to PR to main. Now it is merges to `main`

naveensrinivasan:
@dlorenc  ping

Issue #191, opened 2021-08-12
Retry uploading up to 3x, otherwise mark as failed


priyawadhwa:
Right now, we will continuously retry if some part of the reconcile loops fails. We could retry up to 3 times, and track the number of retries via annotations on the TaskRun:

```
annotations:
  chains.tekton.dev/retries: 2
```

If all three retries fail, we can set the signed annotation on the TaskRun to “failed”:

```
annotations:
  chains.tekton.dev/signed: failed
```

If we set it to "failed", we should leave the "retry" annotation so that it's clear all three retries occurred and failed.

PR #198, opened 2021-08-17
Implemented annotations for retries and failed.

### LINKED ISSUE #204: Chains finalizer can block deletion indefinitely
# Expected Behavior

Deleting a namespace containing taskruns completes successfully.

# Actual Behavior

I am seeing the chains finalization failing because the service account gets cleaned up, which results in a reconciliation error, and the TaskRun sits around forever (keeping the namespace forever).

I noticed this because I was running the pipelines e2e tests, which clean a lot of things up by deleting namespaces, and when I dug into why so much stuff was being left around, it looks a lot like this.

# Additional Info

Perhaps the finalization path should log, but not return errors? 🤷 

cc @priyawadhwa @dlorenc

naveensrinivasan:
* Fixes https://github.com/tektoncd/chains/issues/191
  * Implemented retries and failed annotations.

fixes #204

### DIFF
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -47,9 +47,26 @@ const (
 	ChainsAnnotation             = "chains.tekton.dev/signed"
 	ChainsRetriesAnnotation      = "chains.tekton.dev/retries"
 	ChainsTransparencyAnnotation = "chains.tekton.dev/transparency"
-	Failed                       = "failed"
+	Fail                         = "failed"
+	Pass                         = "true"
+	Retry                        = "retries"
 	MaxRetries                   = 3
 )
+
+type State int
+
+const (
+	Passed State = iota
+	Failed
+	Retrying
+	Unknown
+)
+
+var signedStates = map[string]State{
+	Pass:  Passed,
+	Fail:  Failed,
+	Retry: Retrying,
+}
 
 type Signer interface {
 	SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) error
@@ -61,39 +78,49 @@ type TaskRunSigner struct {
 	SecretPath        string
 }
 
-// IsSigned determines whether a TaskRun has already been signed.
-func IsSigned(tr *v1beta1.TaskRun) (bool, string) {
-	const signed = "true"
+// SignedState returns TaskRun state.
+func SignedState(tr *v1beta1.TaskRun) State {
 	signature, ok := tr.ObjectMeta.Annotations[ChainsAnnotation]
 	if !ok {
-		return false, ""
+		return Unknown
 	}
-	return signature == signed, signature
+
+	if value, ok := signedStates[signature]; !ok {
+		return Unknown
+	} else {
+		return value
+	}
 }
 
 // returns true if:
 // the signed annotation isn't there, the failed annotation isn't there, and retries<3
-func RetryAgain(tr *v1beta1.TaskRun) (bool, string) {
-	retries, ok := tr.ObjectMeta.Annotations[ChainsRetriesAnnotation]
-	if !ok {
-		return false, ""
-	}
-	return true, retries
+func RetryAgain(tr *v1beta1.TaskRun) bool {
+	_, ok := tr.ObjectMeta.Annotations[ChainsRetriesAnnotation]
+	return ok
 }
 
 // MarkSigned marks a TaskRun as signed.
-func MarkSigned(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string,
-	state string) error {
-	return markannotation(tr, ps, annotations, ChainsAnnotation, state)
+func MarkSigned(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string) error {
+	return markannotation(tr, ps, annotations, ChainsAnnotation, Pass)
+}
+
+// MarkFailed marks a TaskRun as signed.
+func MarkFailed(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string) error {
+	return markannotation(tr, ps, annotations, ChainsAnnotation, Fail)
 }
 
 // AddRetries marks a TaskRun as retrying.
-func AddRetries(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string,
-	value int) error {
-	if value > MaxRetries {
-		return fmt.Errorf("The retries cannot be greater than %d", MaxRetries)
+func AddRetries(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string) error {
+	s := "1"
+	if RetryAgain(tr) {
+		s = tr.ObjectMeta.Annotations[ChainsRetriesAnnotation]
+		if v, err := strconv.Atoi(s); err != nil {
+			return fmt.Errorf("invalid value for chain retries %s", s)
+		} else {
+			s = strconv.Itoa(v + 1)
+		}
 	}
-	return markannotation(tr, ps, annotations, ChainsRetriesAnnotation, strconv.Itoa(value))
+	return markannotation(tr, ps, annotations, ChainsRetriesAnnotation, s)
 }
 
 func markannotation(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string,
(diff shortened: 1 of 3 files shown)


pkg/chains/signing.go
```diff
@@ -58,21 +62,45 @@ type TaskRunSigner struct {
 }
 
 // IsSigned determines whether a TaskRun has already been signed.
-func IsSigned(tr *v1beta1.TaskRun) bool {
+func IsSigned(tr *v1beta1.TaskRun) (bool, string) {
 	signature, ok := tr.ObjectMeta.Annotations[ChainsAnnotation]
 	if !ok {
```
priyawadhwa, line 67: I don't think we need this check -- if it doesn't exist then signature will be an empty string and we'll get `false` back anyway

pkg/chains/signing.go
```diff
@@ -58,21 +62,45 @@ type TaskRunSigner struct {
...
-		return false
+		return false, ""
 	}
-	return signature == "true"
+	return signature == "true", signature
+}
+
+// IsRetrying determines whether a TaskRun is retrying state.
+func IsRetrying(tr *v1beta1.TaskRun) (bool, string) {
```
priyawadhwa, line 74: Instead of this, a function that just tells us if we should retry might be simpler to work in:

```go
// returns true if:
// the signed annotation isn't there, the failed annotation isn't there, and retries<3
func RetryAgain(tr *v1beta1.TaskRun) bool {}

File: pkg/chains/signing.go [hunk 3]
```diff
@@ -252,10 +280,30 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
 			}
 		}
 		if merr.ErrorOrNil() != nil {
+			ok, s := IsRetrying(tr)
```
priyawadhwa, line 283: Let's move all of this into a separate function so the current function doesn't get too long

pkg/chains/signing.go
```diff
@@ -58,21 +62,45 @@ type TaskRunSigner struct {
...
 
 // MarkSigned marks a TaskRun as signed.
-func MarkSigned(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string) error {
+func MarkSigned(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string,
+	state string) error {
+	return markannotation(tr, ps, annotations, ChainsAnnotation, state)
+}
+
+// MarkRetries marks a TaskRun as retrying.
+func MarkRetries(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string,
```
priyawadhwa, line 89: Instead of just marking retries, we could call this `AddRetry`, and handle increasing the count here along with updating the annotation

priyawadhwa (commented):
Thanks for fixing this! I've left a few comments.

pkg/chains/signing.go
naveensrinivasan: Addressed it.

pkg/chains/signing.go
naveensrinivasan: Changed it to RetryAgain.

pkg/chains/signing.go
naveensrinivasan: Moved it to a separate function.

pkg/chains/signing.go
naveensrinivasan: Changed it to AddRetry.

pkg/chains/signing.go
priyawadhwa: In that case let's create a `State` type and instead have a function that returns the state --

```go
type State int

const (
    Passed State = iota
    Failed
    Retrying
)

// returns the appropriate State
func State(tr *v1beta1.TaskRun) State
```

pkg/chains/signing.go
```diff
@@ -58,21 +62,47 @@ type TaskRunSigner struct {
...
-		return false
+		return false, ""
 	}
-	return signature == "true"
+	return signature == signed, signature
+}
+
+// returns true if:
+// the signed annotation isn't there, the failed annotation isn't there, and retries<3
+func RetryAgain(tr *v1beta1.TaskRun) (bool, string) {
```
priyawadhwa, line 76: This can just return `true/false`, we don't need to return the string of retries.

Instead, we can get the value of that string in `AddRetries` and then add to it there.

pkg/chains/signing.go
```diff
@@ -58,21 +62,47 @@ type TaskRunSigner struct {
...
 // MarkSigned marks a TaskRun as signed.
-func MarkSigned(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string) error {
+func MarkSigned(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string,
+	state string) error {
+	return markannotation(tr, ps, annotations, ChainsAnnotation, state)
+}
+
+// AddRetries marks a TaskRun as retrying.
+func AddRetries(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string,
+	value int) error {
```
priyawadhwa, line 92: As i mentioned in the above comment, we can remove `value int` from the header of this function, and read it from the annotations within the function

pkg/chains/signing.go
```diff
@@ -252,10 +282,34 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
...
+	if err != nil {
+		return err
+	}
+
+	if count <= MaxRetries {
+		return AddRetries(tr, ts.Pipelineclientset, extraAnnotations, count+1)
+	}
+
+	// Marking it failed after retrying MaxRetries
+	return MarkSigned(tr, ts.Pipelineclientset, extraAnnotations, Failed)
```
priyawadhwa, line 314: I think it would be more readable if we added a `MarkFailed` function and used that instead

naveensrinivasan: Addressed it. Thanks

pkg/chains/signing.go
naveensrinivasan: Took care of it. Thanks

pkg/chains/signing.go
naveensrinivasan: Took care of it. Thanks

pkg/chains/signing.go
```diff
@@ -44,9 +45,29 @@ import (
...
+	MaxRetries                   = 3
 )
 
+type State int
+
+const (
+	Passed State = iota
+	Failed
+	Retrying
+	Unknown
```
priyawadhwa, line 62: i don't think we need the Unknown state

pkg/chains/signing.go
```diff
@@ -57,22 +78,58 @@ type TaskRunSigner struct {
 	SecretPath        string
 }
 
-// IsSigned determines whether a TaskRun has already been signed.
-func IsSigned(tr *v1beta1.TaskRun) bool {
+// SignedState returns TaskRun state.
+func SignedState(tr *v1beta1.TaskRun) State {
```
priyawadhwa, line 82: this function should return:

Passed: if `ChainsAnnotation=true`
Failed: if `ChainsAnnotation=failed`
Retrying: if `ChainsRetriesAnnotation<3`

in that order

pkg/chains/signing.go
```diff
@@ -57,22 +78,58 @@ type TaskRunSigner struct {
...
+	} else {
+		return value
 	}
-	return signature == "true"
+}
+
+// returns true if:
+// the signed annotation isn't there, the failed annotation isn't there, and retries<3
+func RetryAgain(tr *v1beta1.TaskRun) bool {
+	_, ok := tr.ObjectMeta.Annotations[ChainsRetriesAnnotation]
```
priyawadhwa, line 98: this function isn't doing anything except for checking if ChainsRetriesAnnotation exists -- we want it to return true if ChainsRetriesAnnotation<3 OR if ChainsRetriesAnnotation doesn't exist yet

pkg/chains/signing.go
```diff
@@ -57,22 +78,58 @@ type TaskRunSigner struct {
...
+	return markannotation(tr, ps, annotations, ChainsAnnotation, Pass)
+}
+
+// MarkFailed marks a TaskRun as signed.
+func MarkFailed(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string) error {
+	return markannotation(tr, ps, annotations, ChainsAnnotation, Fail)
+}
+
+// AddRetries marks a TaskRun as retrying.
+func AddRetries(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string) error {
```
priyawadhwa, line 113: let's write this as:

```go
func AddRetries(tr *v1beta1.TaskRun, ps versioned.Interface, annotations map[string]string) error {
	retries := tr.ObjectMeta.Annotations[ChainsRetriesAnnotation]
	if retries == "" {
		return markannotation(tr, ps, annotations, ChainsRetriesAnnotation, "0")
	}
	v, err := strconv.Atoi(s)
	if err != nil {
		return errors.Wrapf(err, "invalid value for chain retries %s", s)
	}

	return markannotation(tr, ps, annotations, ChainsRetriesAnnotation, fmt.Sprintf("%d", v +1)
}
```

priyawadhwa: we don't need to check `RetryAgain` because we already did before calling `AddRetries`

pkg/chains/signing.go
```diff
@@ -252,10 +309,34 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
...
 	}
 
 	// Now mark the TaskRun as signed
 	return MarkSigned(tr, ts.Pipelineclientset, extraAnnotations)
 }
+
+// HandleRetries adds retry annotation to keep track of retries and marks the job as failed
+// if retries have reached MaxRetries.
+func HandleRetries(tr *v1beta1.TaskRun, ts *TaskRunSigner, extraAnnotations map[string]string) error {
+	ok := RetryAgain(tr)
```
priyawadhwa, line 326: I think this function just has to say:

```go
	if RetryAgain(tr) {
		return AddRetries(tr, ts.Pipelineclientset, extraAnnotations)

	}
	// otherwise, mark as failed
	return MarkFailed(tr, ts.Pipelineclientset, extraAnnotations)
```

pkg/chains/signing.go
naveensrinivasan: What happens when there is no `ChainsAnnotation`? Also, what is wrong with the present implementation?

pkg/chains/signing.go
naveensrinivasan: We need the unknown state if there isn't an annotation.

priyawadhwa:
closing in favor of #207 , as discussed with @naveensrinivasan offline :)

PR #199, opened 2021-08-18
Included cosign into tools.go

naveensrinivasan:
Included cosign into hack/tools.go


dlorenc:
What's this used for?

naveensrinivasan:
> What's this used for?

It is used for generating cosign keypair https://github.com/tektoncd/chains/blob/main/docs/signing.md#cosign

dlorenc:
Sorry - I know what cosign is used for, I don't know why we would need to add it to the tools.go file. When would that get installed or used?

naveensrinivasan:
> Sorry - I know what cosign is used for, I don't know why we would need to add it to the tools.go file. When would that get installed or used?

If anyone would like to contribute to the project would need `cosign` locally, with it being in the `tools.go` it would be installed https://github.com/tektoncd/chains/blob/main/test/presubmit-tests.sh#L34 and helps contributors installing tools/dependencies with specific versions. That is the idea behind this. 

Let me know your thoughts. I am open to the idea of closing it if you think it doesn't add value.

priyawadhwa:
We discussed moving away from `tools.go` on slack, which I think is a good idea. It probably makes sense to close this!

PR #200, opened 2021-08-18
Doc-removed gen x509 Keypair ,fixed cosign gen

naveensrinivasan:
* Removed the section of generate x509 which had broken link to `gen-keys.sh`
  * Fixed the command for generation of cosign keys.


priyawadhwa (approved):
thanks!

PR #230, opened 2021-09-16
Fix another bug in materials and add integration test

priyawadhwa:
if resources are specified, we don't check if the git params are set. we should check for git params first, and then go through resources if they exist! 

also adds some debug logs we can use in the future to diagnose test failures


dlorenc:
want to squash and then we should be good to merge

PR #243, opened 2021-09-28
Switch DSSE provider to go-securesystemslib

adityasaky:
The DSSE implementation added to in-toto-golang has now been split into its own package which lives at https://github.com/secure-systems-lab/go-securesystemslib/. The next step is the removal of this code from in-toto, so I'm submitting changes to dependent packages. :)

See: https://github.com/in-toto/in-toto-golang/pull/122


priyawadhwa (approved):
Thank you!

PR #245, opened 2021-09-30
Add new signing and storage features

rgreinho:
Adds a set of new features to the signing and storage packages to allow
extracting chains information and verify the signature.


priyawadhwa (commented):
Thanks for opening the draft PR! This is a great start. WDYT of creating a `Verifier` interface analogous to the [Signer](https://github.com/tektoncd/chains/blob/main/pkg/chains/signing.go#L41) interface we already have?

I'm kinda imagining something like this, but haven't fully thought through if it would be easier to implement this or do like a `TaskRunSignerVerifier` type thing instead :)

```go
type Verifier interface {
    VerifyTaskRun(ctx context.Context, tr *v1beta1.TaskRun) error)
}

type TaskRunVerifier struct {
         // use this to read the configmap and get details for the signer&storage backend
	KubeClient        kubernetes.Interface

         // looks like we need this for getting signers
       	SecretPath        string
}
```

pkg/chains/signing.go
```diff
@@ -48,6 +49,21 @@ type TaskRunSigner struct {
 	SecretPath        string
 }
 
+type Verifier interface {
```
priyawadhwa, line 52: could we move all verification code to `pkg/chains/verify.go`? might just make it easier to find

pkg/chains/storage/gcs/gcs.go
```diff
@@ -28,6 +29,10 @@ import (
 
 const (
 	StorageBackendGCS = "gcs"
+	// taskrun-$namespace-$name/$key.signature
+	SignatureNameFormat = "taskrun-%s-%s/%s.signature"
```
priyawadhwa, line 33: nice !

pkg/chains/storage/gcs/gcs.go
```diff
@@ -61,6 +67,7 @@ func (b *Backend) StorePayload(rawPayload []byte, signature string, opts config.
 	// name/namespace as well.
 	// $bucket/taskrun-$namespace-$name-$uid/$key.signature
 	// $bucket/taskrun-$namespace-$name-$uid/$key.payload
+	// Note(rgreinho): it looks like theere is no UUID in the object names :/
```
priyawadhwa, line 70: i think i removed the uuid in the object name because it is part of `opts.Key`, which gets included.

pkg/chains/storage/oci/oci.go
```diff
@@ -42,6 +42,7 @@ const (
 
 // Backend is a storage backend that stores signed payloads in the TaskRun metadata as an annotation.
 // It is stored as base64 encoded JSON.
+// Note(rgreinho): I don't think this is entirely true when the backend is OCI 🤔...
```
priyawadhwa, line 45: yup that comment can probably be removed :)

pkg/chains/signing.go
priyawadhwa: For sure! So to verify I believe all we should need are:
- `cfg config.Config`
- `kc kubernetes.Interface`, which we'll use to actually read the ConfigMap into `cfg`

We can use `kc` to get the ConfigMap object from the cluster, and then we can pass the data into [NewConfigFromConfigMap](https://github.com/tektoncd/chains/blob/main/pkg/config/config.go#L202) to get `cfg`. Once we have `cfg`, I believe that's everything we need for verification.

I don't think we need the `PipelineClientset` for anything, just the kubernetes Client should be sufficient.

Does this make sense?

rgreinho:
Yes it does! Thank you for the pointers,

So I updated a few tests. I actually needed the `PipelineClientset` for the storage function I added for the `tekton` storage ( `retrieveAnnotationValue`). I created a helper function to verify the signature. Let me know what you think.

Side question: when are the e2e tests being executed? For instance I could not test the GCS implementation as it requires an account, so being able to see it run in a CI somewhere would have been  fantastic.

pkg/chains/storage/gcs/gcs.go
```diff
@@ -122,7 +128,44 @@ type realGCSWriter struct {
 	bucket string
 }
 
+type gcsReader interface {
+	GetReader(object string) (io.ReadCloser, error)
+}
+
+type realGCSReader struct {
```
priyawadhwa, line 135: i think we can just call this `reader`, since it's in the `gcs` package it'll imply that it's a gcs reader

pkg/chains/storage/oci/oci.go
```diff
@@ -173,3 +171,26 @@ func (b *Backend) uploadAttestation(attestation in_toto.Statement, rawPayload []
 func (b *Backend) Type() string {
 	return StorageBackendOCI
 }
+
+func (b *Backend) RetrieveSignature(opts config.StorageOpts) (string, error) {
+	return "", fmt.Errorf("not implemented")
```
priyawadhwa, line 176: could you open an issue to implement this just so we don't forget about it?

test/test_utils.go
```diff
@@ -218,3 +223,54 @@ func printDebugging(t *testing.T, ns, taskRunName string) {
...
+	if err != nil {
+		t.Errorf("error retrieving tekton chains configmap: %s", err)
+	}
+	cfg, err := config.NewConfigFromConfigMap(chainsConfig)
+	if err != nil {
+		t.Errorf("error creating tekton chains configuration: %s", err)
+	}
+
+	// Prepare the logger.
+	logger, _ := zap.NewProduction()
```
priyawadhwa, line 239: there's a testing logger we can use [here](https://github.com/tektoncd/chains/blob/24c560f488ddc2b60bb6a4d2d6059ceea2db90db/pkg/chains/formats/intotoite6/intotoite6_test.go#L392)

pkg/chains/storage/gcs/gcs.go
rgreinho: Done ✅ To be consistent, I also renamed `realGCSWriter` to `writer`.

priyawadhwa:
Awesome! This is looking pretty good to me, let's try and run the tests and hopefully they're happy 🤞🏽

rgreinho:
Absolutely:

* Update the `Signer` interface to also handle verifying a signature.
* Add new `TaskRunVerifier` struct to verify the signature of a task run. 
* Update the `Backend` interface to add functions allowing to easily retrieve a payload and a signature.
* Implement the new functions of the `Backend` interface for the `tekton`, `gcs`,  and `docdb` backends.


@priyawadhwa there is no changelog file for this project, should we start one?

priyawadhwa:
We could try updating the `github.com/tektoncd/pipeline` dep here to `v0.28.0` to match what we have in `cli`, might fix the dependency issue 🤞🏽

priyawadhwa:
> Not sure if chains wants to adopt this or not, just thought it was worth mentionning 😉

That sounds really good! I'll look into it :)

rgreinho:
> We could try updating the `github.com/tektoncd/pipeline` dep here to `v0.28.0` to match what we have in `cli`, might fix the dependency issue 🤞🏽

I moved this problem to another issue: https://github.com/tektoncd/chains/issues/269

PR #264, opened 2021-10-12
refactor: upload attestation to rekor once

sbose78:
Do we need to attempt an upload to rekor for every signable type?

```
	enabledSignableTypes := []artifacts.Signable{
		&artifacts.TaskRunArtifact{Logger: logger}, <---- This is what helps produce the attestation and should be pushed to rekor, I think
		&artifacts.OCIArtifact{Logger: logger},
	}
```


priyawadhwa:
I believe the rekor attestations are working correctly here -- we have one upload for the oci signature, and one for the in-toto attestation. Are you seeing different behavior?

sbose78:
I see, so we need them twice at least, thanks!

PR #281, opened 2021-11-11
Fix URL for entry within annotation

### LINKED ISSUE #280: URL to entry in rekor within annotation is incorrect
https://github.com/tektoncd/chains/blob/2b6fc3903e2a71e303ada2b563848e3cb78c6857/pkg/chains/signing.go#L218

This should be `https://rekor.sigstore.dev/api/v1/log/entries?logIndex=12345` instead of `https://rekor.sigstore.dev/12345`

bobcallaway:
This ensures the URL has the correct path to fetch the entry directly.

Fixes #280

Signed-off-by: Bob Callaway <bob.callaway@gmail.com>


priyawadhwa (approved):
Thanks for fixing this @bobcallaway !

PR #220, opened 2021-09-03
Fix some bugs in uploading attestations

priyawadhwa:

PR #284, opened 2021-11-15
Implement missing features to OCI backend

rgreinho:
Implements the features allowing to retrrieve the payloads and the
signatures stored in an OCI backend.

Fixes #220


rgreinho:
I did attempt to add some unit tests here, but I am not exactly sure how to do it, and it looks like there would be a ton of things to mock. Would you have any pointers by chance? @priyawadhwa @dlorenc 

Otherwise, e2e tests should be fine as the use case would be covered there.

rgreinho:
Oh I need to remove the skeleton I wrote for the unit tests. 

Regarding the e2e tests, it looks like some cluster issues.

rgreinho:
Yep, it definitely looks like there are Kubernetes cluster issues. :slightly_frowning_face:

pkg/chains/storage/oci/oci.go
```diff
@@ -179,10 +188,156 @@ func (b *Backend) Type() string {
...
+		}
+		envelopes = append(envelopes, envs...)
+		b.logger.Debugf("extractEnvelopes: %d envelopes retrieved\n", len(envelopes))
+	}
+	return envelopes, nil
+}
+
 func (b *Backend) RetrieveSignature(opts config.StorageOpts) (string, error) {
-	return "", fmt.Errorf("not implemented")
+	envelopes, err := b.extractEnvelopes(opts)
```
priyawadhwa, line 286: I'm wondering if we can take more advantage of the `cosign/pkg/oci` package in the cosign repository here. Do you think something like this would work?

```go
	// given the TaskRun, retrieve the signature
	images := artifacts.ExtractOCIImagesFromResults(b.tr, b.logger)
	if len(images) != 1 {
		return "", errors.New("multiple images found, unsure which to provide signature for")
	}
	// get signature for a specific image
	ref, ok := images[0].(name.Digest)
	if !ok {
		return "", errors.New("error parsing image")
	}
	img, err := ociremote.SignedImage(ref)
	if err != nil {
		return "", err
	}
	sigImage, err := img.Signatures()
	if err != nil {
		return "", err
	}
	sigs, err := sigImage.Get()
	if err != nil {
		return "", err
	}
	for _, s := range sigs {
		if sig, err := s.Base64Signature(); err == nil {
			return sig, nil
		}
	}
	return "", fmt.Errorf("failed to find a signature")
```

and this should require a small change for getting the payload instead:

```
	for _, s := range sigs {
		if payload, err := s.Payload(); err == nil {
			return string(payload), nil
		}
	}
```


(as mentioned in the other comment, I do think we should change the function definition here, but this is just meant to be an example!)

priyawadhwa (commented):
Hey @rgreinho thanks for fixing this! I have a few thoughts:

1. When given a TaskRun, it can have multiple images , each of which can have multiple signatures. We should change these function definitions to reflect that:

```go
   // RetrievePayload maps [image name]:[payload] for a TaskRun
	RetrievePayload(opts config.StorageOpts) (map[string]string, error)
  // RetrieveSignature maps [image name]:[list of signatures] for a TaskRun
	RetrieveSignature(opts config.StorageOpts) (map[string][]string, error)
```

this should make it easy for clients to get all the info they need for verification!

I'm also wondering if something like this could be useful in the CLI for later:

```
$ tkn chains list artifacts

# image1
# image2

$ tkn chains payload --image [image name]
```

but I think we can consider that later! Just throwing it in here so I don't forget 😄

pkg/chains/storage/oci/oci.go
rgreinho: Oh sure! I'll give it a try 👍

rgreinho: @priyawadhwa Where did you get the `ociremote` package from? From version 1.2.1, it looks like it cannot be imported (`use of internal package github.com/sigstore/cosign/internal/oci/remote not allowed`), and if I upgrade the dependency to 1.3.1, it breaks a bunch of stuff.

priyawadhwa:
Hey @rgreinho think this just needs a rebase, but the cosign library upgrade should be in so you should be unblocked now 😄

rgreinho:
Awesome! Thanks a lot @priyawadhwa 👍

rgreinho:
@priyawadhwa Quick update: that almost works! I just lost one part of the behavior I wanted, but I'm sure we can fix it with the new packages you created. I'm going to keep digging around, but I wanted to keep you updated.

If I attempt to retrieve the signature I get what I want:

```bash
╭ cli on  chains-cli via 🐹 v1.17.3
╰❯ ./tkn chains signature $(tkn tr describe --last -o json | jq -r '.metadata.name')
E1202 15:33:11.781098    6531 aws_credentials.go:77] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
MEUCIQDwYrr0XAp/jp5vDv6TEBZvj8lwPE6f1fI/pY8609GVMwIgMQ6PQZYio/VKypjKEqfumo7it4I6it5BzDRG86uEtgw=
```

But if I attempt to retrieve the payload, I actually don't get the payload:

```bash
 cli on  chains-cli via 🐹 v1.17.3
╰❯ ./tkn chains payload $(tkn tr describe --last -o json | jq -r '.metadata.name')
Clients created successfully.
Taskrun created successfully.
E1202 15:33:26.186332    6549 aws_credentials.go:77] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
Backend initialized successfully.
{"Critical":{"Identity":{"docker-reference":"ttl.sh/b5e53772774f48bf80dd6b528e5715a8/slsapoc"},"Image":{"Docker-manifest-digest":"sha256:f82fe2b635e304c7d8445c0117a4dbe35dd3c840078a39e21c88073a885c5e0f"},"Type":"Tekton container signature"},"Optional":{}}
```

Which is similar to the output of `cosign verify`:

```bash
╭ cli on  chains-cli via 🐹 v1.17.3
╰❯ cosign verify --key "${COSIGN_KEY}" ttl.sh/b5e53772774f48bf80dd6b528e5715a8/slsapoc

Verification for ttl.sh/b5e53772774f48bf80dd6b528e5715a8/slsapoc --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The signatures were verified against the specified public key
  - Any certificates were verified against the Fulcio roots.

[{"critical":{"identity":{"docker-reference":"ttl.sh/b5e53772774f48bf80dd6b528e5715a8/slsapoc"},"image":{"docker-manifest-digest":"sha256:f82fe2b635e304c7d8445c0117a4dbe35dd3c840078a39e21c88073a885c5e0f"},"type":"Tekton container signature"},"optional":{}}]
```

However I wanted the output of `cosign verify-attestation`:

```bash
╭ cli on  chains-cli via 🐹 v1.17.3 ⌛ 2s
╰❯ cosign verify-attestation --key "${COSIGN_KEY}" ttl.sh/b5e53772774f48bf80dd6b528e5715a8/slsapoc

Verification for ttl.sh/b5e53772774f48bf80dd6b528e5715a8/slsapoc --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The signatures were verified against the specified public key
  - Any certificates were verified against the Fulcio roots.
{"payloadType":"application/vnd.in-toto+json","payload":"eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdGF0ZW1lbnQvdjAuMSIsInByZWRpY2F0ZVR5cGUiOiJodHRwczovL3Rla3Rvbi5kZXYvY2hhaW5zL3Byb3ZlbmFuY2UiLCJzdWJqZWN0IjpbeyJuYW1lIjoidHRsLnNoL2I1ZTUzNzcyNzc0ZjQ4YmY4MGRkNmI1MjhlNTcxNWE4L3Nsc2Fwb2MiLCJkaWdlc3QiOnsic2hhMjU2IjoiZjgyZmUyYjYzNWUzMDRjN2Q4NDQ1YzAxMTdhNGRiZTM1ZGQzYzg0MDA3OGEzOWUyMWM4ODA3M2E4ODVjNWUwZiJ9fV0sInByZWRpY2F0ZSI6eyJpbnZvY2F0aW9uIjp7InBhcmFtZXRlcnMiOlsiQlVJTERFUl9JTUFHRT17c3RyaW5nIGRvY2tlci5pby9jbmJzL3NhbXBsZS1idWlsZGVyOmJpb25pY0BzaGEyNTY6NmMwM2RkNjA0NTAzYjU5ODIwZmQxNWFkYmM2NWMwYTA3N2E0N2UzMWQ0MDRhM2RjYWQxOTBmMzE3OWU5MjBiNSBbXX0iLCJBUFBfSU1BR0U9e3N0cmluZyB0dGwuc2gvYjVlNTM3NzI3NzRmNDhiZjgwZGQ2YjUyOGU1NzE1YTgvc2xzYXBvYyBbXX0iLCJTT1VSQ0VfU1VCUEFUSD17c3RyaW5nIGFwcHMvcnVieS1idW5kbGVyIFtdfSIsIlBST0NFU1NfVFlQRT17c3RyaW5nIHdlYiBbXX0iLCJFTlZfVkFSUz17YXJyYXkgIFtdfSIsIlJVTl9JTUFHRT17c3RyaW5nICBbXX0iLCJDQUNIRV9JTUFHRT17c3RyaW5nIHR0bC5zaC9iNWU1Mzc3Mjc3NGY0OGJmODBkZDZiNTI4ZTU3MTVhOC9zbHNhcG9jLWNhY2hlIFtdfSIsIlVTRVJfSUQ9e3N0cmluZyAxMDAwIFtdfSIsIkdST1VQX0lEPXtzdHJpbmcgMTAwMCBbXX0iLCJTT1VSQ0VfU1VCUEFUSD1bXSIsIkVOVl9WQVJTPVtdIiwiUFJPQ0VTU19UWVBFPXdlYiIsIlJVTl9JTUFHRT1bXSIsIkNBQ0hFX0lNQUdFPVtdIiwiU0tJUF9SRVNUT1JFPWZhbHNlIiwiVVNFUl9JRD0xMDAwIiwiR1JPVVBfSUQ9MTAwMCIsIlBMQVRGT1JNX0RJUj1lbXB0eS1kaXIiXSwicmVjaXBlX3VyaSI6InRhc2s6Ly9idWlsZHBhY2tzIiwiZXZlbnRfaWQiOiI0OTYyMmEyYy04ZWY0LTQ2NTMtOTIyYi02ZTU4NDFiMjliN2YiLCJidWlsZGVyLmlkIjoidGVrdG9uLWNoYWlucyJ9LCJyZWNpcGUiOnsic3RlcHMiOlt7ImVudHJ5UG9pbnQiOiIjIS91c3IvYmluL2VudiBiYXNoXG5zZXQgLWVcblxuaWYgW1sgXCIkKHdvcmtzcGFjZXMuY2FjaGUuYm91bmQpXCIgPT0gXCJ0cnVlXCIgXV07IHRoZW5cbiAgZWNobyBcIlx1MDAzZSBTZXR0aW5nIHBlcm1pc3Npb25zIG9uICckKHdvcmtzcGFjZXMuY2FjaGUucGF0aCknLi4uXCJcbiAgY2hvd24gLVIgXCIkKHBhcmFtcy5VU0VSX0lEKTokKHBhcmFtcy5HUk9VUF9JRClcIiBcIiQod29ya3NwYWNlcy5jYWNoZS5wYXRoKVwiXG5maVxuXG5mb3IgcGF0aCBpbiBcIi90ZWt0b24vaG9tZVwiIFwiL2xheWVyc1wiIFwiJCh3b3Jrc3BhY2VzLnNvdXJjZS5wYXRoKVwiOyBkb1xuICBlY2hvIFwiXHUwMDNlIFNldHRpbmcgcGVybWlzc2lvbnMgb24gJyRwYXRoJy4uLlwiXG4gIGNob3duIC1SIFwiJChwYXJhbXMuVVNFUl9JRCk6JChwYXJhbXMuR1JPVVBfSUQpXCIgXCIkcGF0aFwiXG5kb25lXG5cbmVjaG8gXCJcdTAwM2UgUGFyc2luZyBhZGRpdGlvbmFsIGNvbmZpZ3VyYXRpb24uLi5cIlxucGFyc2luZ19mbGFnPVwiXCJcbmVudnM9KClcbmZvciBhcmcgaW4gXCIkQFwiOyBkb1xuICAgIGlmIFtbIFwiJGFyZ1wiID09IFwiLS1lbnYtdmFyc1wiIF1dOyB0aGVuXG4gICAgICAgIGVjaG8gXCItXHUwMDNlIFBhcnNpbmcgZW52IHZhcmlhYmxlcy4uLlwiXG4gICAgICAgIHBhcnNpbmdfZmxhZz1cImVudi12YXJzXCJcbiAgICBlbGlmIFtbIFwiJHBhcnNpbmdfZmxhZ1wiID09IFwiZW52LXZhcnNcIiBdXTsgdGhlblxuICAgICAgICBlbnZzKz0oXCIkYXJnXCIpXG4gICAgZmlcbmRvbmVcblxuZWNobyBcIlx1MDAzZSBQcm9jZXNzaW5nIGFueSBlbnZpcm9ubWVudCB2YXJpYWJsZXMuLi5cIlxuRU5WX0RJUj1cIi9wbGF0Zm9ybS9lbnZcIlxuXG5lY2hvIFwiLS1cdTAwM2UgQ3JlYXRpbmcgJ2VudicgZGlyZWN0b3J5OiAkRU5WX0RJUlwiXG5ta2RpciAtcCBcIiRFTlZfRElSXCJcblxuZm9yIGVudiBpbiBcIiR7ZW52c1tAXX1cIjsgZG9cbiAgICBJRlM9Jz0nIHJlYWQgLXIga2V5IHZhbHVlIFx1MDAzY1x1MDAzY1x1MDAzYyBcIiRlbnZcIlxuICAgIGlmIFtbIFwiJGtleVwiICE9IFwiXCIgXHUwMDI2XHUwMDI2IFwiJHZhbHVlXCIgIT0gXCJcIiBdXTsgdGhlblxuICAgICAgICBwYXRoPVwiJHtFTlZfRElSfS8ke2tleX1cIlxuICAgICAgICBlY2hvIFwiLS1cdTAwM2UgV3JpdGluZyAke3BhdGh9Li4uXCJcbiAgICAgICAgZWNobyAtbiBcIiR2YWx1ZVwiIFx1MDAzZSBcIiRwYXRoXCJcbiAgICBmaVxuZG9uZVxuIiwiYXJndW1lbnRzIjpbIi0tZW52LXZhcnMiLCIkKHBhcmFtcy5FTlZfVkFSU1sqXSkiXSwiZW52aXJvbm1lbnQiOnsiY29udGFpbmVyIjoicHJlcGFyZSIsImltYWdlIjoiZG9ja2VyLXB1bGxhYmxlOi8vYmFzaEBzaGEyNTY6YjIwODIxNWE0NjU1NTM4YmU2NTJiMjc2OWQ4MmU1NzZiYzRkMGEyYmIxMzIxNDRjMDYwZWZjNWJlOGMzZjVkNiJ9LCJhbm5vdGF0aW9ucyI6bnVsbH0seyJlbnRyeVBvaW50IjoiL2NuYi9saWZlY3ljbGUvY3JlYXRvciIsImFyZ3VtZW50cyI6WyItYXBwPSQod29ya3NwYWNlcy5zb3VyY2UucGF0aCkvJChwYXJhbXMuU09VUkNFX1NVQlBBVEgpIiwiLWNhY2hlLWRpcj0kKHdvcmtzcGFjZXMuY2FjaGUucGF0aCkiLCItY2FjaGUtaW1hZ2U9JChwYXJhbXMuQ0FDSEVfSU1BR0UpIiwiLXVpZD0kKHBhcmFtcy5VU0VSX0lEKSIsIi1naWQ9JChwYXJhbXMuR1JPVVBfSUQpIiwiLWxheWVycz0vbGF5ZXJzIiwiLXBsYXRmb3JtPS9wbGF0Zm9ybSIsIi1yZXBvcnQ9L2xheWVycy9yZXBvcnQudG9tbCIsIi1wcm9jZXNzLXR5cGU9JChwYXJhbXMuUFJPQ0VTU19UWVBFKSIsIi1za2lwLXJlc3RvcmU9JChwYXJhbXMuU0tJUF9SRVNUT1JFKSIsIi1wcmV2aW91cy1pbWFnZT0kKHBhcmFtcy5BUFBfSU1BR0UpIiwiLXJ1bi1pbWFnZT0kKHBhcmFtcy5SVU5fSU1BR0UpIiwiJChwYXJhbXMuQVBQX0lNQUdFKSJdLCJlbnZpcm9ubWVudCI6eyJjb250YWluZXIiOiJjcmVhdGUiLCJpbWFnZSI6ImRvY2tlci1wdWxsYWJsZTovL2NuYnMvc2FtcGxlLWJ1aWxkZXJAc2hhMjU2OjZjMDNkZDYwNDUwM2I1OTgyMGZkMTVhZGJjNjVjMGEwNzdhNDdlMzFkNDA0YTNkY2FkMTkwZjMxNzllOTIwYjUifSwiYW5ub3RhdGlvbnMiOm51bGx9LHsiZW50cnlQb2ludCI6IiMhL3Vzci9iaW4vZW52IGJhc2hcbnNldCAtZVxuZ3JlcCBcImRpZ2VzdFwiIC9sYXllcnMvcmVwb3J0LnRvbWwgfCBjdXQgLWQnXCInIC1mMiB8IGN1dCAtZCdcIicgLWYyIHwgdHIgLWQgJ1xcbicgfCB0ZWUgXCIkKHJlc3VsdHMuQVBQX0lNQUdFX0RJR0VTVC5wYXRoKVwiXG5cbiMgRGlzYWJsZSBzaGVsbGNoZWNrIGhlcmUgc2luY2UgJCgpIGlzIHVzZSBmb3IgdmFyaWFibGUgc3Vic3RpdHV0aW9uIGFuZCBub3QgY29tbWFuZCBleGVjdXRpb24uXG4jIHNoZWxsY2hlY2sgZGlzYWJsZT1TQzIwMDVcbmVjaG8gXCIkKHBhcmFtcy5BUFBfSU1BR0UpXCIgfCB0ZWUgXCIkKHJlc3VsdHMuQVBQX0lNQUdFX1VSTC5wYXRoKVwiXG4iLCJhcmd1bWVudHMiOm51bGwsImVudmlyb25tZW50Ijp7ImNvbnRhaW5lciI6InJlc3VsdHMiLCJpbWFnZSI6ImRvY2tlci1wdWxsYWJsZTovL2Jhc2hAc2hhMjU2OmIyMDgyMTVhNDY1NTUzOGJlNjUyYjI3NjlkODJlNTc2YmM0ZDBhMmJiMTMyMTQ0YzA2MGVmYzViZThjM2Y1ZDYifSwiYW5ub3RhdGlvbnMiOm51bGx9XX0sIm1ldGFkYXRhIjp7ImJ1aWxkU3RhcnRlZE9uIjoiMjAyMS0xMi0wMlQyMToyNzo1MloiLCJidWlsZEZpbmlzaGVkT24iOiIyMDIxLTEyLTAyVDIxOjI4OjU5WiJ9fX0=","signatures":[{"keyid":"SHA256:DojOZnqYRvBOs0OiTE4Zkivfx8nAqHuX0s2n/roaHQI","sig":"MEYCIQDR2nraonxpg0gD5P5IgQBBR7mSahNKzV5IgECIzZb5vAIhALNDJGTZCoCEIP1DmWmGpYI3c0ySuK/wM8Q/mGPIIqKc"}]}
```

pkg/chains/storage/oci/oci.go
```diff
@@ -194,9 +196,60 @@ func (b *Backend) Type() string {
...
+	// get signature for a specific image
+	ref, ok := images[0].(name.Digest)
+	if !ok {
+		return nil, errors.New("error parsing image")
+	}
+	img, err := ociremote.SignedImage(ref)
+	if err != nil {
+		return nil, err
+	}
+	sigImage, err := img.Signatures()
```
priyawadhwa, line 245: Hey @rgreinho, I believe if you want to get the payload for the attestation, you'll have to call `img.Attestations()` instead of `img.Signatures()` here, and then call `Payload()` on the output of that!

rgreinho: It worked 🤗

rgreinho:
@priyawadhwa I think I implemented everything now 😀 

However I have a few of questions:

1. Since the interfaces now return maps, what would you think of renaming the interfaces to `RetrieveSignatures` and `RetrievePayloads`?
2. Can all the backends support multiple images and multiple signatures? I was not sure how that would work with `tekton`, `gcs` and `docdb` so I assumed not. For instance for `tekton` that would mean that we would need to update the annotations to handle this use case.
3. The keys of the maps are not consistent between the different backends and I am not super happy with that. Ideally the key would always be the image ref (I think), but I was not sure how to achieve that for `docdb` and `gcs`, and for `tekton` I used the `IMAGE_URL` from the results, assuming there could only be one.

rgreinho:
The e2e tests failed due to a problem with the test Kubernetes cluster 🧪

priyawadhwa:
Hey @rgreinho! 

I think the integration tests are failing because the tests aren't compiling:

```
test/test_utils.go:265:63: cannot use signature (type map[string][]string) as type string in argument to strings.NewReader
test/test_utils.go:265:93: cannot use body (type map[string]string) as type string in argument to strings.NewReader
```
>Since the interfaces now return maps, what would you think of renaming the interfaces to RetrieveSignatures and RetrievePayloads?

sgtm! 


> Can all the backends support multiple images and multiple signatures? I was not sure how that would work with tekton, gcs and docdb so I assumed not. For instance for tekton that would mean that we would need to update the annotations to handle this use case.

So the only backend that supports images is the `oci` backend. The other backends can be used to store the `taskrun.yaml` so you'd only have one payload and one signature over that anyway!

> The keys of the maps are not consistent between the different backends and I am not super happy with that. Ideally the key would always be the image ref (I think), but I was not sure how to achieve that for docdb and gcs, and for tekton I used the IMAGE_URL from the results, assuming there could only be one.

I think maybe we don't need to worry about this because images should only be stored in `oci`? Unless I'm misunderstanding the question.

rgreinho:
Arg, you are right, I must have forgotten to update the e2e tests... 🤦 Sorry about that. 

The other answers make sense, so I'll update the code and my next commit should be good to review! 👍

rgreinho:
Alright, everything built and tested fine this time @priyawadhwa 🎉 So this work should be ready for review ✅

rgreinho:
Rebased 🙃

pkg/artifacts/signable.go
```diff
@@ -116,6 +116,7 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
 	urlSuffix := "IMAGE_URL"
 	digestSuffix := "IMAGE_DIGEST"
 	for _, res := range tr.Status.TaskRunResults {
+		// Note(rgreinho): using HasSuffix should be enough.
 		if strings.HasSuffix(res.Name, urlSuffix) || res.Name == urlSuffix {
```
priyawadhwa, line 120: i think you're correct that [HasSuffix is sufficient!](https://go.dev/play/p/IlhbHza049E) we could probably remove the second check in both places :)

pkg/chains/storage/docdb/docdb.go
```diff
@@ -93,33 +93,43 @@ func (b *Backend) Type() string {
...
+	m := make(map[string]string)
+	for _, d := range documents {
+		m[d.Name] = string(d.Signed)
+	}
+
+	return m, nil
 }
 
-func (b *Backend) retrieveDocument(opts config.StorageOpts) (SignedDocument, error) {
+func (b *Backend) retrieveDocument(opts config.StorageOpts) ([]SignedDocument, error) {
```
priyawadhwa, line 129: nit: change name of this func to `retrieveDocuments`

pkg/chains/storage/docdb/docdb_test.go
```diff
@@ -91,21 +91,21 @@ func TestBackend_StorePayload(t *testing.T) {
...
 
 			// Check the signature.
-			sig, err := b.RetrieveSignature(opts)
+			signatures, err := b.RetrieveSignatures(opts)
 			if err != nil {
 				t.Fatal(err)
 			}
-			if string(sig) != tt.args.signature {
-				t.Errorf("wrong signature, expected %s, got %s", tt.args.signature, string(sig))
+			if signatures[obj.Name][0] != tt.args.signature {
```
priyawadhwa, line 98: let's add a check to make sure there's exactly one signature here as well,

```go
if len(signatures[obj.Name]) != 1 {
  t.Fatalf("unexpected number of signatures: %d", len(signatures[obj.Name]))
}
```

pkg/chains/storage/tekton/tekton.go
```diff
@@ -110,18 +111,37 @@ func (b *Backend) retrieveAnnotationValue(annotationKey string, decode bool) (st
...
-	return b.retrieveAnnotationValue(fmt.Sprintf(SignatureAnnotationFormat, opts.Key), true)
+	signature, err := b.retrieveAnnotationValue(fmt.Sprintf(SignatureAnnotationFormat, opts.Key), true)
+	if err != nil {
+		return nil, err
+	}
+
+	m := make(map[string][]string)
+	for _, res := range b.tr.Status.TaskRunResults {
+		if strings.HasSuffix(res.Name, "IMAGE_URL") {
+			m[res.Value] = []string{signature}
```
priyawadhwa, line 124: should we just be mapping [SignatureAnnotation]:signature for the `tekton` case?

pkg/chains/storage/tekton/tekton.go
```diff
@@ -110,18 +111,37 @@ func (b *Backend) retrieveAnnotationValue(annotationKey string, decode bool) (st
...
+func (b *Backend) RetrievePayloads(opts config.StorageOpts) (map[string]string, error) {
 	b.logger.Infof("Retrieving payload on TaskRun %s/%s", b.tr.Namespace, b.tr.Name)
 	payload, err := b.retrieveAnnotationValue(fmt.Sprintf(PayloadAnnotationFormat, opts.Key), true)
 	if err != nil {
-		return "", err
+		return nil, err
+	}
+	m := make(map[string]string)
+	for _, res := range b.tr.Status.TaskRunResults {
+		if strings.HasSuffix(res.Name, "IMAGE_URL") {
```
priyawadhwa, line 140: Similarly, should we just be mapping `payloadAnnotation`:`payload` here?

priyawadhwa (commented):
This is looking really good! Left a few comments.

pkg/artifacts/signable.go
rgreinho: Addressed ✅

pkg/chains/storage/docdb/docdb.go
rgreinho: Addressed ✅

pkg/chains/storage/docdb/docdb_test.go
rgreinho: I can add that check to the tests for sure, but in this case shouldn't we rewrite the `RetrieveSignatures` function with the check instead?

Something like:

```go
func (b *Backend) RetrieveSignatures(opts config.StorageOpts) (map[string][]string, error) {
	// Retrieve the document.
	documents, err := b.retrieveDocuments(opts)
	if err != nil {
		return nil, err
	}

        if len(documents) != 1 {
            return nil, fmt.Errorf("unexpected number of signatures: %d", len(documents))
        }

	m := make(map[string][]string)
	// Extract and decode the signature.
	sig, err := base64.StdEncoding.DecodeString(d.Signature)
	if err != nil {
		return nil, err
	}
	m[d.Name] = []string{string(sig)}
	return m, nil
}
```

pkg/chains/storage/tekton/tekton.go
rgreinho: Sure thing! Addressed ✅ 

Should we also add the check to ensure there is only one image then?

pkg/chains/storage/tekton/tekton.go
rgreinho: Addressed ✅ 

Like above, should we check there is only one image?

pkg/chains/storage/docdb/docdb_test.go
priyawadhwa: i think it's ok to leave the check in the test for now!

pkg/chains/storage/tekton/tekton.go
priyawadhwa: i think this should be ok :)

pkg/chains/storage/docdb/docdb_test.go
rgreinho: Copy that 👍

pkg/chains/storage/oci/oci.go
```diff
@@ -194,10 +198,88 @@ func (b *Backend) Type() string {
...
+
+func (b *Backend) RetrievePayloads(opts config.StorageOpts) (map[string]string, error) {
+	images, err := b.RetrieveArtifact(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	m := make(map[string]string)
+	for ref, img := range images {
+		attImage, err := img.Attestations()
```
priyawadhwa, line 237: The `payload` here should depend on the value of `opts.PayloadFormat` correct? I believe if the `PayloadFormat` is `in-toto` then we want to call `img.Attestations()` here, but if the `PayloadFormat` is `simplesigning` then we'll want to call `img.Signatures()` here.

rgreinho: Make sense 👍 Addressed ✅

priyawadhwa (approved):
LGTM! Thanks for your patience and work in getting this merged @rgreinho , this is awesome :D

PR #289, opened 2021-11-23
Upgrade `sigstore/cosign` dependency to v1.3.1

priyawadhwa:
this should help unblock https://github.com/tektoncd/chains/pull/284 :)


rgreinho:
I am more learning from this PR than  anything else, but I the new functions provide a nice abstraction level compare to the previous and the package reorganization (`ociremote`, `mutate`) makes complete sense.

Issue #288, opened 2021-11-23
support for multiple output providers simultaneously


06kellyjac:
### Feature request

support for multiple output providers simultaneously for cases where provenance needs to be stored in multiple places

### Use case

1. Migration
  * An organization might already have provenance storage set up for a provider e.g. `oci` and they want to move to document storage `docdb` and to ease the transition they might want to output to both simultaneously during that transition/migration period
2. Duplication due to requirements, regulated or otherwise
  * An organization's security policy or regulatory requirements may dictate the data needs to be stored within multiple systems be it for redundancy or different methods of analysis/usage
    * e.g. `docdb` for querying, `oci` for keeping the provenance next to signatures, or `gcs` for longer term archive storage
    * If there is an issue with one of the storage methods either due to chains or an outage with that particular method of storage

---

*Also:* There is also the potential to support 2+ output locations for the same provider type e.g. for 2 separate `gcs` buckets, sending a copy to a 3rd party security operations center
This might be outside the scope of this Feature Request so it could go in a separate ticket/issue

priyawadhwa:
Hey @06kellyjac thank for opening this issue! I think this is a great idea.

Right now we have a config option where users choose the storage backend they want:
```
artifacts.taskrun.storage=[oci|gcs|docdb|tekton]
```

There are a couple ways we could add in multiple backends:
1. Turn the config value into a comma-separated list e.g. `artifacts.taskrun.storage=oci,gcs`

2. Separate each backend into its own config key:
```
artifacts.taskrun.storage.oci=enabled
artifacts.taskrun.storage.gcs=enabled
```
this could actually go hand in hand with https://github.com/tektoncd/chains/issues/279 (we could just assume that if no storage backend is enabled, then the entire artifact is disabled)

06kellyjac:
I like the 

```
artifacts.taskrun.storage.oci.enabled = true
artifacts.taskrun.storage.gcs.enabled = true
```

method since it keeps `enabled` alongside the other options like bucket name etc

For the `artifacts.taskrun.storage=oci,gcs` option it probably wouldn't be a comma separated list but a single string OR array

```
artifacts.taskrun.storage:
  - gcs
  - something
```

but again I think I prefer option 2 you proposed

Issue #279, opened 2021-11-04
"chains.tekton.dev/signed": "failed"


mtcolman:
Hi, I'm not sure if this is a bug or not, so raising in free-form, apologies if I've used the wrong thing.

I've followed the tutorial here: https://github.com/tektoncd/chains/blob/main/docs/tutorials/getting-started-tutorial.md

And at the end of it I get:
```
$ cosign verify-blob -key cosign.pub -signature ./signature ./payload 
Verified OK
```

However, when having a look around with:
```
$ oc get taskrun $TASKRUN -o=json | jq  -r ".metadata.annotations"
```

I can see that :
```
{
  "chains.tekton.dev/cert-taskrun-d72c4b52-569a-4e42-b030-0e4d5e6931fe": "",
  "chains.tekton.dev/chain-taskrun-d72c4b52-569a-4e42-b030-0e4d5e6931fe": "",
  "chains.tekton.dev/payload-taskrun-d72c4b52-569a-4e42-b030-0e4d5e6931fe": "<I've removed for brevity>",
  "chains.tekton.dev/retries": "3",
  "chains.tekton.dev/signature-taskrun-d72c4b52-569a-4e42-b030-0e4d5e6931fe": "<I've removed for brevity>",
  "chains.tekton.dev/signed": "failed",
  "pipeline.tekton.dev/release": "v0.22.0"
}
```

Shows:
```
"chains.tekton.dev/retries": "3",
chains.tekton.dev/signed": "failed"
```

Has something gone wrong here? And if so, why did the `cosign verify-blob` return `Verified OK`?

Thanks!

pxp928:
Based on the error messages it seems like there is an issue with the gcr registry that the example taskrun uses. If i change the registry to something else instead of `gcr.io/foo/bar` it changes to `chains.tekton.dev/signed: "true"`
```
    outputs:
    - name: builtImage
      resourceSpec:
        type: image
        params:
        - name: url
          value: gcr.io/foo/bar
```

priyawadhwa:
Hey @mtcolman thanks for pointing this out! My guess is the tutorial needs to be updated and Chains is trying to upload the signature for the image to an OCI registry (which it doesn't have auth for, so it fails). But, it first stores the signature and payload on the TaskRun correctly, which is why `verify-blob` passes.

Could you provide the output of:
```
kubectl describe cm -n tekton-chains chains-config
```

that should confirm my suspicion!

pxp928:
[...]

PR #297, opened 2021-12-02
WIP Update to allow for multiple backend selections

pxp928:
Changes made to allow for multiple backend to be selected. If none are selected, that artifact is disabled. Work in progress. Tests need to be altered or rewritten based on how the solution is implemented.

Issue #279 and #288


pkg/config/config.go
```diff
@@ -97,13 +100,20 @@ type TransparencyConfig struct {
...
+	taskrunFormatKey        = "artifacts.taskrun.format"
+	taskrunStorageTektonKey = "artifacts.taskrun.storage.tekton"
+	taskrunStorageOCIKey    = "artifacts.taskrun.storage.oci"
+	taskrunStorageGCSKey    = "artifacts.taskrun.storage.gcs"
+	taskrunStorageDocDBKey  = "artifacts.taskrun.storage.docdb"
+	taskrunSignerKey        = "artifacts.taskrun.signer"
+
+	ociEnabled          = "artifact.oci.enabled"
+	ociFormatKey        = "artifacts.oci.format"
+	ociStorageTektonKey = "artifacts.oci.storage.tekton"
```
priyawadhwa, line 112: I'm rethinking having a separate key for each backend now that I see it 🤔 

It might be difficult to keep the config organized. The advantage of a comma-separated list is that it would be backwards compatible as well.

WDYT of doing that instead?

```
artifacts.oci.storage=tekton,gcs
```

and then if the value is empty `artifacts.oci.storage=""` then we assume that we don't want to store anything at all (which would address the issue in https://github.com/tektoncd/chains/issues/279 as well)

pxp928: Yes I agree. The config does get very cluttered. Having a comma-separated would be the better way to go. Plus it would also require less unit test updates. Let me change to using that and I will re-push on this branch.

priyawadhwa: thank you! we can keep the defaults as they are as well that way :)

PR #302, opened 2021-12-07
added multi-backend support

pxp928:
Changes made to allow for multiple backend to be selected. If none are selected, that artifact is disabled. Work in progress. Tests need to be altered or rewritten based on how the solution is implemented. Continuation of PR #297

Issue #279 and #288

### DIFF
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -141,7 +141,9 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
 	var merr *multierror.Error
 	extraAnnotations := map[string]string{}
 	for _, signableType := range enabledSignableTypes {
-
+		if !signableType.Enabled(cfg) {
+			continue
+		}
 		payloadFormat := signableType.PayloadFormat(cfg)
 		// Find the right payload format and format the object
 		payloader, ok := allFormats[payloadFormat]
@@ -196,16 +198,18 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
 			}
 
 			// Now store those!
-			b := allBackends[signableType.StorageBackend(cfg)]
-			storageOpts := config.StorageOpts{
-				Key:           signableType.Key(obj),
-				Cert:          signer.Cert(),
-				Chain:         signer.Chain(),
-				PayloadFormat: payloadFormat,
-			}
-			if err := b.StorePayload(rawPayload, string(signature), storageOpts); err != nil {
-				logger.Error(err)
-				merr = multierror.Append(merr, err)
+			for _, backend := range signableType.StorageBackend(cfg).List() {
+				b := allBackends[backend]
+				storageOpts := config.StorageOpts{
+					Key:           signableType.Key(obj),
+					Cert:          signer.Cert(),
+					Chain:         signer.Chain(),
+					PayloadFormat: payloadFormat,
+				}
+				if err := b.StorePayload(rawPayload, string(signature), storageOpts); err != nil {
+					logger.Error(err)
+					merr = multierror.Append(merr, err)
+				}
 			}
 
 			if shouldUploadTlog(cfg, tr) {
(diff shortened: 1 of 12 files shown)


pkg/chains/signing.go
```diff
@@ -141,90 +141,93 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
 	var merr *multierror.Error
 	extraAnnotations := map[string]string{}
 	for _, signableType := range enabledSignableTypes {
+		if signableType.Enabled(cfg) {
```
priyawadhwa, line 144: let's check the opposite to make this more readable:

```go
if !signableType.Enabled(cfg) {
   continue
}
```

pkg/config/config.go
```diff
@@ -129,17 +130,26 @@ const (
 	ChainsConfig = "chains-config"
 )
 
+func (artifact *Artifact) Enabled() bool {
+
+	if artifact.StorageBackend.Len() == 1 && artifact.StorageBackend.Has("") {
```
priyawadhwa, line 135: can simplify to

```go
return !(artifact.StorageBackend.Len() == 1 && artifact.StorageBackend.Has(""))
```

pkg/config/config.go
```diff
@@ -257,3 +267,25 @@ func asString(key string, target *string, values ...string) cm.ParseFunc {
 		return nil
 	}
 }
+
+// asStringSet parses the value at key as a sets.String (split by ',') into the target, if it exists.
+func asStringSet(key string, target *sets.String, values ...string) cm.ParseFunc {
+	return func(data map[string]string) error {
+		if raw, ok := data[key]; ok {
+			splitted := strings.Split(raw, ",")
+			if raw != "" {
```
priyawadhwa, line 276: we can make this more readable by first checking

```
if raw == "" {
   // set empty target
   return nil
}
 ```

File: pkg/config/config.go [hunk 4]
```diff
@@ -257,3 +267,25 @@ func asString(key string, target *string, values ...string) cm.ParseFunc {
 		return nil
 	}
 }
+
+// asStringSet parses the value at key as a sets.String (split by ',') into the target, if it exists.
+func asStringSet(key string, target *sets.String, values ...string) cm.ParseFunc {
```
priyawadhwa, line 272: it might make more sense for this function to also take in a string set of possible values (i think that might make checking against the passed in set easier)

```go
func asStringSet(key string, target *sets.String, allowed sets.String) cm.ParseFunc
```

and then compare `allowed` with whatever the user has passed in.

priyawadhwa (commented):
Just a few nits! Would you mind adding an integration test to check for mulitple backends as well?

You could copy `TestOCIStorage` but also include `tekton` as a backend and verify both!

https://github.com/tektoncd/chains/blob/a7e0557bf10abbe1d4f3f90e947857ba87890194/test/e2e_test.go#L322

pkg/chains/signing.go
pxp928: changed in `signing.go` and `verifier.go`

pkg/config/config.go
pxp928: fixed.

pkg/config/config.go
pxp928: changed.

pkg/config/config.go
pxp928: changed.

pxp928:
> Just a few nits! Would you mind adding an integration test to check for mulitple backends as well?
> 
> You could copy `TestOCIStorage` but also include `tekton` as a backend and verify both!
> 
> https://github.com/tektoncd/chains/blob/a7e0557bf10abbe1d4f3f90e947857ba87890194/test/e2e_test.go#L322

Added new integration test with `OCI` and `tekton` backend storage.

pkg/chains/verifier.go
```diff
@@ -55,26 +55,29 @@ func (tv *TaskRunVerifier) VerifyTaskRun(ctx context.Context, tr *v1beta1.TaskRu
 	signers := allSigners(tv.SecretPath, cfg, logger)
 
 	for _, signableType := range enabledSignableTypes {
-
+		if !signableType.Enabled(cfg) {
```
priyawadhwa, line 58: cc @rgreinho just an fyi since you've been working on this code!

priyawadhwa (approved):
this looks great! thanks @pxp928 for adding this awesome new feature 🎉

PR #303, opened 2021-12-08
WIP: differentiate payload types on tekton storage

loosebazooka:
Adds two new annotations
```
+	chains.tekton.dev/taskrun-%s
+	chains.tekton.dev/attestation-%s
```
and removes the old typeless payload annotation
```
-       chains.tekton.dev/payload-%s
```
Storage on tekton now expects a payload type and will fail if not intoto or a taskrun

Signed-off-by: Appu Goundan <appu@google.com>

### DIFF
--- a/docs/tutorials/getting-started-tutorial.md
+++ b/docs/tutorials/getting-started-tutorial.md
@@ -56,7 +56,7 @@ Next, retrieve the signature and payload from the object (they are stored as bas
 
 ```shell
 $ export TASKRUN_UID=$(kubectl get taskrun $TASKRUN -o=json | jq -r '.metadata.uid')
-$ kubectl get taskrun $TASKRUN -o=json | jq  -r ".metadata.annotations[\"chains.tekton.dev/payload-taskrun-$TASKRUN_UID\"]" | base64 --decode > payload
+$ kubectl get taskrun $TASKRUN -o=json | jq  -r ".metadata.annotations[\"chains.tekton.dev/TODO_FIGURE_OUT_PAYLOAD_TYPE-taskrun-$TASKRUN_UID\"]" | base64 --decode > payload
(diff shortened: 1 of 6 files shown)


loosebazooka:
fyi @priyawadhwa

pkg/chains/storage/tekton/tekton.go
```diff
@@ -29,11 +30,12 @@ import (
 )
 
 const (
-	StorageBackendTekton      = "tekton"
-	PayloadAnnotationFormat   = "chains.tekton.dev/payload-%s"
```
loosebazooka, line 33: does removing this annotation result in an "api change"? or are these considered internal implementation details?

priyawadhwa: it could count as an "api change" if people have some client that depends on them (although i don't know of any tbh) -- i think we could play it safe and keep `payload` for now and deprecate it in the next release

we can add this into the [deprecation table](https://github.com/tektoncd/chains/blob/main/docs/deprecations.md#deprecation-table) as well!

docs/tutorials/getting-started-tutorial.md
```diff
@@ -56,7 +56,7 @@ Next, retrieve the signature and payload from the object (they are stored as bas
 
 ```shell
 $ export TASKRUN_UID=$(kubectl get taskrun $TASKRUN -o=json | jq -r '.metadata.uid')
-$ kubectl get taskrun $TASKRUN -o=json | jq  -r ".metadata.annotations[\"chains.tekton.dev/payload-taskrun-$TASKRUN_UID\"]" | base64 --decode > payload
+$ kubectl get taskrun $TASKRUN -o=json | jq  -r ".metadata.annotations[\"chains.tekton.dev/TODO_FIGURE_OUT_PAYLOAD_TYPE-taskrun-$TASKRUN_UID\"]" | base64 --decode > payload
```
loosebazooka, line 59: default is `tekton`

loosebazooka:
This probably needs a bit more design thought around it.

Issue #296, opened 2021-12-02
Distinguish attestations during storage


priyawadhwa:
Right now when we upload to OCI we distinguish between signatures and attestations.

We could do the same when uploading to gcs, tekton and docdb and append the ".att" suffix to help distinguish. 

for gcs this change would happen here:
https://github.com/tektoncd/chains/blob/e5f8d4424255798a633e200420fc00d8c19c044b/pkg/chains/storage/gcs/gcs.go#L66

and etc for the other backends

loosebazooka:
I'll take a look

loosebazooka:
Just to check my understanding.

Attestations will exist only for payloads that are of payload type "`in-toto`/`tekoton-provenance`", other types should just follow the existing flow and write raw payloads to the storage backend?

```
if payloadtype = "in-toto" or "tekton-provenance" {
  call it an attestation?
}
else {
  call it a payload?
}
```

priyawadhwa:
so basically payloads are always either a "signature" or an "attestation" (also signed). this check basically is used to determine which one is which.  We can probs copy cosign and append ".sig" for signatures and ".att" for attestations to help distinguish between the two!

loosebazooka:
impl for gcs: https://github.com/tektoncd/chains/pull/300

loosebazooka:
I'm a little confused about how this would look for the tekton backend? The tekton backend is storing stuff in the taskrun? Would it be cyclically referring to itself?

priyawadhwa:
yup, it would pretty much either be storing the taskrun or the intoto attestation back on the taskrun itself

we could just change the annotation name here to indicate which payload it is, but tbh i don't think it's a big deal either way

loosebazooka:
I think I'm missing something here, but what's the point of storing the taskrun in the taskrun? Is it storing a snapshot of the taskrun at a different time?

priyawadhwa:
sort of, it's mostly just a convenient. it's a snapshot of the taskrun before chains modifies it. afterwards, chains adds the TaskRun (payload), signature, and any certs that may have been issued back onto the TaskRun as an annotation

users would have to strip those new annotations to get the original payload for signature verification (totally possible, just kinda annoying).

loosebazooka:
gotcha, okay this makes sense now

loosebazooka:
So after discussion, it seems like we want a way to store multiple signed data in the backends from different signing events.

[...]

PR #304, opened 2021-12-08
go back to using payload with no extension

loosebazooka:
see: https://github.com/tektoncd/chains/issues/296#issuecomment-989177936


loosebazooka:
fyi: @priyawadhwa

priyawadhwa (approved):

Issue #306, opened 2021-12-11
SLSA v0.2 provenance implementation for invocation.parameters doesn't line up with spec


mlieberman85:
Below creates a list of key=value strings for invocation.parameters:
https://github.com/tektoncd/chains/blob/ddf7daffe805b7d94c59b742f050b8fc74d55446/pkg/chains/formats/intotoite6/intotoite6.go#L118-L139

However the docs on the spec state:

> invocation.parameters object, optional
>
> Collection of all external inputs that influenced the build on top of invocation.configSource. For example, if the invocation type were "make", then this might be the flags passed to make aside from the target, which is captured in invocation.configSource.entryPoint.
> Consumers SHOULD accept only "safe" invocation.parameters. The simplest and safest way to achieve this is to disallow any parameters altogether.
>This is an arbitrary JSON object with a schema defined by buildType.
>This is considered to be incomplete unless metadata.completeness.parameters is true. Unset or null is equivalent to empty.

Key being `arbitrary JSON object`

See: https://github.com/slsa-framework/slsa/blob/main/docs/provenance/v0.2.md

PR #307, opened 2021-12-11
Make SLSA provenance v0.2 invocation.parameters compliant with spec

mlieberman85:
The SLSA provenance v0.2 spec states that invocation.parameters
must be an arbitrary JSON object. This modifies the format
implementation to comply with that requirement.

For more info: https://github.com/tektoncd/chains/issues/306

---

This should fix: #306 

@priyawadhwa Want to get your input on whether the output makes sense and if you think it breaks anything because people expect strings of key=value.

There's a lot of options we can take to make the parameters output nicer, but I kept it simple for now until I get input from others.


dlorenc:
Looks like you missed the e2e test definition.

mlieberman85:
> Looks like you missed the e2e test definition.

Oops, will fix shortly.

mlieberman85:
I can't get some of these tests to run accurately local so might take me a little.

dlorenc:
Feel free to test with `git push -f` !

priyawadhwa (approved):
Thanks for fixing this!

Issue #309, opened 2021-12-12
SLSA v0.2 provenance implementation for materials.digest (if using git) doesn't line up with spec


mlieberman85:
materials.digest as implemented has a key for materials.revision but not for materials.digest which is in the spec.

> materials[*].digest object (DigestSet), optional
> Collection of cryptographic digests for the contents of this artifact.

One thing to not is there is currently some ambiguity in SLSA as to whether or not "commit id" counts as digest. https://github.com/slsa-framework/slsa/issues/214 -- However looking at a https git reference can be confusing without adding `git+` to the URI. Do other folks have any thoughts?

Similar to #306 and #307

See: https://github.com/slsa-framework/slsa/blob/main/docs/provenance/v0.2.md

priyawadhwa:
Looks like there are two bugs here:
1. Instead of using the "revision" key we should be using "sha1" to indicate a commit, which is what the provenance in the the [github actions example](https://github.com/slsa-framework/slsa/blob/main/docs/provenance/v0.2.md#github-actions) does

```
  "configSource": {
    "entryPoint": "build.yaml:build",
    // The git repo that contains the build.yaml referenced in the entrypoint.
    "uri": "git+https://github.com/foo/bar.git",
    // The resolved git commit hash reflecting the version of the repo used
    // for this build.
    "digest": {"sha1": "abc..."}
  },
```

2. we need to add the `git+` prefix and `.git` suffix to the URI

mlieberman85:
There is also: https://github.com/in-toto/attestation/issues/28 here. There is some discussion if some of this at the attestation spec level as well.

PR #310, opened 2021-12-13
Fix bugs in provenance implementation for materials.digest

priyawadhwa:
Makes the following fixes:
1. Replaces the `revision` key with `sha1` for git commits
2. uses the SPDX download locaton format for specifying the git URI 

fixes https://github.com/tektoncd/chains/issues/309

### DIFF
--- a/pkg/chains/formats/intotoite6/intotoite6.go
+++ b/pkg/chains/formats/intotoite6/intotoite6.go
@@ -202,7 +202,7 @@ func materials(tr *v1beta1.TaskRun) []slsa.ProvenanceMaterial {
 	if gitCommit != "" && gitURL != "" {
 		mats = append(mats, slsa.ProvenanceMaterial{
 			URI:    gitURL,
-			Digest: map[string]string{"revision": gitCommit},
+			Digest: map[string]string{"sha1": gitCommit},
 		})
 		return mats
 	}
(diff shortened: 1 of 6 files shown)


pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -286,5 +289,18 @@ func gitInfo(tr *v1beta1.TaskRun) (commit string, url string) {
 			url = r.Value
 		}
 	}
+
+	url = spdxGit(url, "")
 	return
 }
+
+// supports the SPDX format which is recommended by in-toto
+// ref: https://spdx.dev/spdx-specification-21-web-version/
```
loosebazooka, line 298: This link is more targeted in the doc (but maybe more likely to break?) https://spdx.dev/spdx-specification-21-web-version/#h.49x2ik5

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -286,5 +289,18 @@ func gitInfo(tr *v1beta1.TaskRun) (commit string, url string) {
...
+// ref: https://github.com/in-toto/attestation/blob/849867bee97e33678f61cc6bd5da293097f84c25/spec/field_types.md
+func spdxGit(url, revision string) string {
```
loosebazooka, line 300: I think revision can be a sha1, is that something we should test?
image

loosebazooka: although that does make the "digest" a bit redundant.

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -286,5 +289,18 @@ func gitInfo(tr *v1beta1.TaskRun) (commit string, url string) {
 			url = r.Value
 		}
 	}
+
+	url = spdxGit(url, "")
```
loosebazooka, line 293: we don't have to validate url do we? It's pretty much expected to be a fully formed url?

priyawadhwa: yah, i think it's expected!

pkg/chains/formats/intotoite6/intotoite6.go
priyawadhwa: ya so if the revision is a commit ID then I just defaulted to putting it in the "digest" section instead of the URI -- does that make sense?

PR #319, opened 2022-01-06
Add support for all of cosign's ambient credentials.

mattmoor:
This change expands the supported signing credentials to include all of
cosign's ambient credential options, including Kubernetes (works with
GKE, EKS and soon AKS), and SPIFFE.

/assign @priyawadhwa


priyawadhwa (approved):
thanks for updating this!

PR #327, opened 2022-01-12
Use cosign flags with 2 hyphens

concaf:
cosign will soon be moving away from flags with a single hyphen to
double hyphens, e.g. `-key` will be deprecated in favor of `--key`.
This commit fixes that.


priyawadhwa (approved):
thanks for updating this!

PR #330, opened 2022-01-12
Update tutorial

rgreinho:
Updates the "getting started" tutorial to take advantage of the `tkn`
cli. Also clarifies the backend configuration options and fixes the
commands in the signature verification section.


priyawadhwa (approved):
thanks for fixing this up!

Issue #308, opened 2021-12-11
SLSA v0.2 provenance implementation for builder.id doesn't line up with spec


mlieberman85:
Similar to: #306 

The builder.id is supposed to be a TypeURI. "tekton-chains" does not match up. This should be a trivial change but would require making a decision around a URI for tekton-chains as a builder type.

I think this might require a little bit of discussion as we might want to encode additional metadata about the builder outside of just the fact that the build was recorded by "tekton-chains." For example do we want to allow for the configuration of additional identity metadata around the infrastructure that built it or what type of pipeline/task ran?

I could see something like `https://tekton.cd/chains-build/v1/buildpacks/v1` or similar being useful. I think TypeURI's are not supposed to have query parameters, but if they did it might make some of this a little cleaner.

From docs:
>builder.id string (TypeURI), required
>URI indicating the builder's identity.

See: https://github.com/slsa-framework/slsa/blob/main/docs/provenance/v0.2.md

priyawadhwa:
> For example do we want to allow for the configuration of additional identity metadata around the infrastructure that built it or what type of pipeline/task ran? 

I think the spot for this would be `metadata.buildInvocationId`, which the docs describe as:

> Identifies this particular build invocation, which can be useful for finding associated logs or other ad-hoc analysis. The exact meaning and format is defined by builder.id; by default it is treated as opaque and case-sensitive. The value SHOULD be globally unique.


Right now we have the `buildType` set to `https://tekton.dev/attestations/chains@v2`, so perhaps `builder.id` could be something like `https://tekton.dev/chains/v2`? This field can actually also be [configured by the user](https://github.com/tektoncd/chains/blob/main/docs/config.md#in-toto-configuration) via the config to be whatever they want, this would just be the default.

mlieberman85:
This makes sense! I can open up a PR later today since it sounds like just a one (or a couple of) line change.

priyawadhwa:
Awesome! I think it would just have to be changed here and in the associated unit tests:

https://github.com/tektoncd/chains/blob/16daafa426772f465a1055094e3a4ae74beb390a/pkg/config/config.go#L161

priyawadhwa:
this was fixed by https://github.com/tektoncd/chains/pull/331

PR #331, opened 2022-01-12
Make builder id for SLSA provenance valid URI

mlieberman85:
The SLSA provenance v0.2 spec requires that the builder.id be a
valid URI. It was just "tekton-chains" this changes it to
"https://tekton.dev/chains/v2"

### DIFF
--- a/pkg/chains/formats/provenance/provenance_test.go
+++ b/pkg/chains/formats/provenance/provenance_test.go
@@ -274,10 +274,10 @@ spec:
 			"my-param={string string-param []}",
 			"my-array-param={array  [my array]}",
 		},
-		ID: "tekton-chains",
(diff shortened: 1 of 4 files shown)


mlieberman85:
This should fix #308

loosebazooka:
will wait for @priyawadhwa to approve.

priyawadhwa (approved):
thanks @mlieberman85 !

priyawadhwa:
Looks like you'll need to update [TestParse](https://github.com/tektoncd/chains/blob/17c1789f63095bb0212a084dc7cd220426915317/pkg/config/store_test.go#L93) as well

mlieberman85:
Thanks, will update tomorrow and push out.

Issue #332, opened 2022-01-13
`artifacts.oci.format: tekton` causes signature not being stored to the OCI or annotations.


pxp928:
# Expected Behavior

Configmap setting for chains:
```
  artifacts.oci.format: tekton
  artifacts.oci.storage: tekton,oci
  artifacts.taskrun.format: in-toto
  artifacts.taskrun.storage: tekton,oci
```

Does the OCI format `tekton` need to be removed? It causes signature not to be store in OCI. The documentation and code options need to be removed if this is not intended to work.

```
Running cosign verify --key k8s://tekton-chains/signing-secrets "${DOCKER_IMG}":
latest
sha256-46fbe3e8658a6eb85b99693aee3137ccc522af95213fb10d19785404c98e4b92.att
Error: no matching signatures:

main.go:48: error during command execution: no matching signatures:
```
# Steps to Reproduce the Problem

1. Build and deployed chains via the main branch using `ko`
2. Change the configmap to the above
2. Running through the tutorial on chains to store in oci

# Additional Info

- Kubernetes version:

 ```
 kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.4", GitCommit:"b695d79d4f967c403a96986f1750a35eb75e75f1", GitTreeState:"clean", BuildDate:"2021-11-17T15:48:33Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.4", GitCommit:"b695d79d4f967c403a96986f1750a35eb75e75f1", GitTreeState:"clean", BuildDate:"2021-11-17T15:42:41Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

```
kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'
v0.29.0
```

**Comment by priyawadhwa** [comment]
+1 to removing the `tekton` option for `artifacts.oci.format`

**Comment by pxp928** [comment]
@priyawadhwa Should the storage backend for `artifacts.oci.storage` also not have `tekton` as an option?

```
asString(ociFormatKey, &cfg.Artifacts.OCI.Format, "tekton", "simplesigning"),
[...]

PR #334, opened 2022-01-14
Remove un-used options from `artifacts.oci.format` and `artifacts.oci.storage`

pxp928:
Address issue #332

Removing the `tekton` option for `artifacts.oci.format` as it causes the errors with signature validation via cosign.
Change to have `oci` as an the only option for `artifacts.oci.storage`


pxp928:
Reworking the integration tests

priyawadhwa (approved):
thanks for fixing this!

PR #335, opened 2022-01-16
Add installation instructions for OpenShift

concaf:
This commit documents the steps required to install Chains on OpenShift.
These steps are pretty similar to what's required for other components like Pipeline - https://github.com/tektoncd/pipeline/blob/main/docs/install.md#installing-tekton-pipelines-on-openshift

CC: @vdemeester @sbose78


cooktheryan:
thanks @concaf !!!

README.md
```diff
@@ -56,6 +56,44 @@ NAME                                       READY   STATUS      RESTARTS   AGE
 tekton-chains-controller-c4f7c57c4-nrjb2   1/1     Running     0          160m
 ```
 
+#### Installation on OpenShift
+Prerequisite: you'll need [Tekton Pipelines](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#installing-tekton-pipelines-on-openshift) installed on your OpenShift cluster before you install Chains.
+
+To install Tekton Chains on OpenShift, you must first apply the `anyuid` security
```
vdemeester, line 62: Is it required for the chains controller to run as root ? or to run as a specific user ?

sbose78: The recent version needs to run as root. I had to specify runAsUser: 0

vdemeester: Ah ok 👍🏼 We might want to see how to enhance this in the future (cc @priyawadhwa) but for now, let's document it this way 👼🏼

vdemeester (approved):

README.md
concaf: @vdemeester isn't this the case with pipeline and triggers as well!? 🤔

priyawadhwa (approved):

PR #338, opened 2022-01-19
Add `chains-info` ConfigMap with version info

### LINKED ISSUE #336: Add a `chains-info` ConfigMap to expose Chains' version
Other Tekton components like Pipeline and Triggers expose a `<component>-info` ConfigMap which can be queried by external tools such as CLI, primarily to query for the component's version.

See https://github.com/tektoncd/pipeline/blob/e0e994e3def8efe51d921b1cd579c8b9cde32ddd/config/config-info.yaml#L15-L29 and https://github.com/tektoncd/triggers/blob/7fbff3b122fcb77d44e1b39bb45c8a935e61f5ed/config/config-info.yaml#L15-L29

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: pipelines-info
  namespace: tekton-pipelines
  labels:
    app.kubernetes.io/instance: default
    app.kubernetes.io/part-of: tekton-pipelines
data:
  # Contains pipelines version which can be queried by external
  # tools such as CLI. Elevated permissions are already given to
  # this ConfigMap such that even if we don't have access to
  # other resources in the namespace we still can have access to
  # this ConfigMap.
  version: "devel"
```

concaf:
This commit adds a `chains-info` ConfigMap which contains the version of
Chains installed on the cluster. This is meant to be primarily read by
other tools like `tkn` to display component version.

This ConfigMap can be read by someone who does not even have permissions
to view other resources in the `tekton-chains` namespace, hence making
it easier to advertise the version instead of having tools wrangle with
labels on deployments to which permissions might be limited.

This is in line with similar `<component>-info` ConfigMaps used in other
Tekton components like Pipeline and Triggers.

Fix #336

FYI: @priyawadhwa the `version` field will be updated during release here - https://github.com/tektoncd/chains/blob/61c344977ae3edbd7827e649c6d6126f30649969/release/publish.yaml#L133-L134

CC: @vinamra28 since you implemented this for other components, LMK if I missed something.

### DIFF
--- a/config/100-rolebinding.yaml
+++ b/config/100-rolebinding.yaml
@@ -138,3 +138,42 @@ roleRef:
   kind: Role
   name: tekton-chains-leader-election
   apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: tekton-chains-info
+  namespace: tekton-chains
+  labels:
+    app.kubernetes.io/component: chains
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-chains
+rules:
+  # All system:authenticated users need to have access
+  # to the chains-info ConfigMap even if they don't
+  # have access to other resources present in the
+  # installed namespace
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["chains-info"]
+    verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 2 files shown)


concaf:
Pipeline PR for reference - https://github.com/tektoncd/pipeline/pull/3971

vinamra28 (commented):
@concaf looks good, just one minor fix

vinamra28 (approved):

priyawadhwa (approved):
thanks for adding this!

Issue #320, opened 2022-01-10
Error uploading signature and attestation


sbose78:
I upgraded from v0.5.0 today and I'm getting a `getting signed image: open containers/auth.json: permission denied` in the logs. This seems to be happening while the "Starting attestation upload to OCI " is attempted. Seems to me that the signature isn't getting pushed to OCI for the same reason. 

* Major changes post v0.5.x include https://github.com/tektoncd/chains/pull/284
* The error message appeared in https://github.com/tektoncd/chains/pull/291/commits/07c1b5ffb72c5474c86dbbdccbcb49aa7a7f98b5

# Expected Behavior

Upload of signature & attestation with tekton-provenance should succeed in chains v0.6.x, since it had worked in 0.5.x ( I'm aware tekton-provenance has been deprecated, that shouldn't impact this 🤔  )

# Actual Behavior

Signing fails.

```
{"level":"info","ts":"2022-01-10T00:23:58.771Z","logger":"watcher","caller":"chains/signing.go:185","msg":"Signing object with x509","commit":"48a85c4","knative.dev/controller":"github.com.tektoncd.chains.pkg.reconciler.taskrun.Reconciler","knative.dev/kind":"tekton.dev.TaskRun","knative.dev/traceid":"fb99b1f9-acdf-434a-87eb-1787388908c5","knative.dev/key":"tekton-chains/new-pipeline-sye5yl-buildah-2gms7"}
{"level":"info","ts":"2022-01-10T00:23:58.771Z","logger":"watcher","caller":"oci/oci.go:71","msg":"Storing payload on TaskRun tekton-chains/new-pipeline-sye5yl-buildah-2gms7","commit":"48a85c4","knative.dev/controller":"github.com.tektoncd.chains.pkg.reconciler.taskrun.Reconciler","knative.dev/kind":"tekton.dev.TaskRun","knative.dev/traceid":"fb99b1f9-acdf-434a-87eb-1787388908c5","knative.dev/key":"tekton-chains/new-pipeline-sye5yl-buildah-2gms7"}
[...]

PR #346, opened 2022-01-21
Add securityContext to controller pod

concaf:
This commit adds securityContext to the controller pod which has been
missing till now. This is in line with securityContext in other Tekton
components like pipeline and triggers.

There are a few reasons why securityContext is important to add to the
controller:
- the securityContext specifies the user and group as 65532 which is
  what is used by the distroless nonroot image that ko uses in the
  controller image.
- securityContext is also useful in distributions like OpenShift which
  assign a random user while running a pod. If securityContext is not
  set, the controller fails with permission denied errors like in #320

### DIFF
--- a/config/100-deployment.yaml
+++ b/config/100-deployment.yaml
@@ -98,6 +98,11 @@ spec:
               fieldPath: metadata.namespace
         - name: METRICS_DOMAIN
           value: tekton.dev/chains
+        securityContext:
+          allowPrivilegeEscalation: false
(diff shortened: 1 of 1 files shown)


vdemeester (commented):
+:100: for consistency with the rest of the component controller's deployment

mattmoor (approved):

priyawadhwa (approved):
thanks for fixing this!

PR #360, opened 2022-02-03
Change default Fulcio URL to match cosign's

tcnghia:
Currently *cosign* is using https://v1.fulcio.sigstore.dev as the default. Changing *chains* to use the same.

Signed-off-by: Nghia Tran <tcnghia@gmail.com>

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -152,7 +152,7 @@ func defaultConfig() *Config {
 		},
 		Signers: SignerConfigs{
 			X509: X509Signer{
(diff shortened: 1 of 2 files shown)


priyawadhwa (commented):
TY for fixing this!!

So we do have a test for this but unfortunately we can't run it in CI because our CI cluster doesn't have workload identity enabled.

Would you mind running it locally to make sure this passes?

```
go test -tags=e2e -run TestFulcio ./test/... 
```

LMK if it doesn't work!

priyawadhwa (approved):

PR #365, opened 2022-02-04
Fix default fulcio addr in docs

priyawadhwa:
cc @tcnghia


priyawadhwa:
hey @lukehinds thanks for approving! think this also need an /lgtm to merge

lukehinds:
> hey @lukehinds thanks for approving! think this also need an /lgtm to merge

woops forgot to tag it

PR #370, opened 2022-02-09
Bring in a newer version of cosign

### LINKED ISSUE #367: Wire through the ability to configure how to obtain the Spiffe token from Fulcio provider.
### Feature request
Once this goes in:
https://github.com/sigstore/cosign/issues/1420

We should make sure we document how to utilize that for acquiring auth for Fulcio using it.

/assign @vaikas

vaikas:
This allows specifying the Spiffe socket path via env variable, add docs for how to use
Spiffe for authenticating to Fulcio and clarify couple of things.
As a result of that, update the signature of a call to getting
a signer.

Fix #367 


go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/tektoncd/chains
 
 go 1.16
 
-replace k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20211109043538-20434351676c
+replace (
+	github.com/sigstore/cosign => github.com/sigstore/cosign v1.5.2-0.20220208171233-ddca5deac2ba
```
mattmoor, line 6: Do we actually need a `replace` or are you using this to fast-forward to HEAD?

vaikas:
Just a newer version

go.mod
mattmoor: Ok, we shouldn't need a replace to pull in the newer version.  This should do it:
```
go get -d github.com/sigstore/cosign@main
./hack/update-deps.sh
```

vaikas: That's right! Spaced, thanks! Done

priyawadhwa (approved):

PR #387, opened 2022-03-01
Fix the inconsistent labels on chains resource

piyush-garg:
This will fix the inconsistent labels used on all
resources which is also violating the policy of
labels mentioned here https://github.com/tektoncd/pipeline/blob/main/docs/developers/resources-labelling.md


priyawadhwa (approved):
thanks for fixing this!

PR #388, opened 2022-03-01
Revert default fulcio URL to fulcio.sigstore.dev

priyawadhwa:
now, fulcio.sigstore.dev and v1.fulcio.sigstore.dev point to the same service, so we can switch back.

cc @tcnghia who originally found this bug


wlynch (approved):
/lgtm

PR #389, opened 2022-03-03
Add support for Grafeas storage backend

chuangw6:
# Changes

This PR implements the support for Grafeas storage backend (i.e. [Google Container Analysis](https://cloud.google.com/container-analysis/docs)) via [Grafeas](https://grafeas.io/) open source API. 

If Google Container Analysis server is chosen, this new feature will enable Tekton Chains to write provenance to Container Analysis that then can be consumed by GCP services such as [Binary Authorization](https://cloud.google.com/binary-authorization).

# Submitter Checklist

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
  functionality, content, code)
- [ ] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

#### 1. configure container analysis as storage backend for taskrun and/or oci artifacts.
- `kubectl patch configmap chains-config -n tekton-chains -p='{"data":{"artifacts.taskrun.storage": "grafeas"}}'`
- `kubectl patch configmap chains-config -n tekton-chains -p='{"data":{"artifacts.oci.storage": "grafeas"}}'`

#### 2. inform tekton chains project ID and note name that will be used to create occurrences.
- `kubectl patch configmap chains-config -n tekton-chains -p='{"data":{"storage.grafeas.projectid": "<GCP_PROJECT_ID>"}}'`
- `kubectl patch configmap chains-config -n tekton-chains -p='{"data":{"storage.grafeas.noteid": "<NOTE_NAME>"}}'` (optional, if `noteid` is not configured, we will generate a name in the format of `tekton-<namespace>`)

### 3. Authenticate to GCP (if Container Analysis is chosen for the storage backend)
- [Use Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity)
- Enable WI for the KSA `tekton-chains-controller` under `tekton-chains` namespace


pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+// It is stored as base64 encoded JSON.
+type Backend struct {
+	logger         *zap.SugaredLogger
+	tr             *v1beta1.TaskRun
+	client         pb.GrafeasV1Beta1Client
+	cfg            config.Config
+	occurrenceRefs []string // store occurrence IDs that are automatically generated during the time of creation
+}
+
+func NewStorageBackend(logger *zap.SugaredLogger, tr *v1beta1.TaskRun, cfg config.Config) (*Backend, error) {
```
Yongxuanzhang, line 55: should have comment for exported functions

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	conn, err := grpc.Dial("dns:///containeranalysis.googleapis.com",
+		grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{})),
+		grpc.WithDefaultCallOptions(grpc.PerRPCCredentials(creds)),
+	)
+	if err != nil {
+		return nil, err
+	}
+	// defer conn.Close()
```
Yongxuanzhang, line 75: is this needed?

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+		cfg:            cfg,
+		occurrenceRefs: []string{},
+	}
+
+	// create note
+	err = b.createNote()
+
+	if err != nil {
+		// ignore note already exisits error
+		logger.Info(err)
```
Yongxuanzhang, line 94: I think this should be `logger.Error(err) `?

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+			Attestation: &attestationpb.Attestation{
+				Signature: &attestationpb.Attestation_GenericSignedAttestation{
+					GenericSignedAttestation: &attestationpb.GenericSignedAttestation{
+						// Uspecified ContentType because we have simplesigning for OCI, In-toto for TaskRun
+						ContentType:       attestationpb.GenericSignedAttestation_CONTENT_TYPE_UNSPECIFIED,
+						SerializedPayload: payload,
+						Signatures: []*commonpb.Signature{
+							{
+								Signature:   []byte(signature),
+								PublicKeyId: b.cfg.Signers.KMS.KMSRef,
```
wlynch, line 226: What happens if a KMS key isn't used (e.g. k8s secret, fulcio)?

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+// Retrieve payloads from container analysis and store it in a map
+func (b *Backend) RetrievePayloads(opts config.StorageOpts) (map[string]string, error) {
+	// initialize an empty map for result
+	result := make(map[string]string)
+
+	// get all occurrences created using this backend
+	occurrences, err := b.getOccurrences()
+	if err != nil {
+		return nil, err
+	}
```
wlynch, line 133: IIUC, this is relying on the `Backend.occurrences` field, which means that this will only work if the instances reading the occurrences is the same as the one that wrote it. I don't think we should make this assumption - it limits usage of this library (e.g. what if I want to retrieve the payloads in a CLI or something else that is only reading the results).

pkg/chains/storage/ca/ca_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	type args struct {
+		tr        *v1beta1.TaskRun
+		signed    []byte
+		signature string
+		opts      config.StorageOpts
+	}
+	tests := []struct {
+		name    string
+		args    args
```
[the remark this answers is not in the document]
Yongxuanzhang: maybe this is to keep consistent with other storage tests like,
https://github.com/tektoncd/chains/blob/8b29855fca5574de429fb20234cf66aefe003119/pkg/chains/storage/gcs/gcs_test.go#L32-L41

pkg/chains/storage/ca/ca.go
chuangw6: sure. Thanks!

pkg/chains/storage/ca/ca.go
chuangw6: Make sense to me! Thanks for pointing out that!

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+	logger         *zap.SugaredLogger
+	tr             *v1beta1.TaskRun
+	client         pb.GrafeasV1Beta1Client
+	cfg            config.Config
+	occurrenceRefs []string // store occurrence IDs that are automatically generated during the time of creation
+}
+
+func NewStorageBackend(logger *zap.SugaredLogger, tr *v1beta1.TaskRun, cfg config.Config) (*Backend, error) {
+	// add TypeMeta information to taskrun objects
+	addTypeInformationToObject(tr)
```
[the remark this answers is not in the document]
chuangw6: This is needed when we compose the URI for taskrun artifact to solve the issue that GroupVersionKind returns empty string. I thought it would be something general to add for a taskrun, and otherwise if we move it closer to URI generation, it would be called each time we store attestation of an artifact from the taskrun.

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+	return result, nil
+}
+
+func (b *Backend) Type() string {
+	return StorageBackendCA
+}
+
+// ----------------------------- Helper Functions ----------------------------
+func (b *Backend) createNote() error {
+	noteID := b.cfg.Storage.ContainerAnalysis.NoteID
```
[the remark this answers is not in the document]
chuangw6: Great point! I can add a default value in the configmap! In case clients don't set projectID, would it be okay if we just log the error?

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, occ := range occurrences {
+		// get the Signature identifier
+		name := occ.GetResource().GetUri()
+		// get "Signatures" field from the occurrence DSSE envelop
+		// signatures := occurrence.GetEnvelope().GetSignatures()
+		signatures := occ.GetAttestation().GetAttestation().GetGenericSignedAttestation().Signatures
```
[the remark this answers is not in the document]
chuangw6: Ah a mistake. thanks for pointing out this!

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+			continue
+		}
+		obj.GetObjectKind().SetGroupVersionKind(gvk)
+		break
+	}
+
+	return nil
+}
+
+func (b *Backend) RetrieveResourceURI(opts config.StorageOpts) string {
```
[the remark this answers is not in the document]
chuangw6: Nah, it's just used in the test file which is within the same package. I'll make the first letter lowercase.

pkg/chains/storage/ca/ca.go
chuangw6: That's true. I was only looking at the place where the retrieve functions are used [here](https://github.com/tektoncd/chains/blob/8b29855fca5574de429fb20234cf66aefe003119/pkg/chains/verifier.go#L57-L75) in verifier.go. In verifier, as a receiver function of the backend, `RetrievePayloads` retrieves payload for each backend instance of each `signableType`. So for taskrun artifact, it only reads the occurrence of taskrun artifact attestation. And for OCI artifact, it reads the occurrences of OCI images outputted from a taskrun.

> it limits usage of this library (e.g. what if I want to retrieve the payloads in a CLI or something else that is only reading the results).

Can you elaborate a bit more here?

pkg/chains/storage/ca/ca_test.go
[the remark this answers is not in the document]
chuangw6: Yes, I was trying to follow the tests of other storage backend support. But I'll improve the readability there.

pkg/chains/storage/ca/ca.go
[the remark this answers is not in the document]
wlynch: My concern about doing this is if it's modifying the TaskRun, this could modify the checksum being produced depending on the ordering of when/how the object is signed.

The safest thing to do would be to not modify the incoming TaskRun, and only extract the values where needed.

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+func (b *Backend) createOccurrenceRequest(payload []byte, signature string, opts config.StorageOpts) *pb.CreateOccurrenceRequest {
+	occurrenceDetails := &pb.Occurrence_Attestation{
+		Attestation: &attestationpb.Details{
+			Attestation: &attestationpb.Attestation{
+				Signature: &attestationpb.Attestation_GenericSignedAttestation{
+					GenericSignedAttestation: &attestationpb.GenericSignedAttestation{
+						// Uspecified ContentType because we have simplesigning for OCI, In-toto for TaskRun
+						ContentType:       attestationpb.GenericSignedAttestation_CONTENT_TYPE_UNSPECIFIED,
```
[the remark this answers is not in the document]
chuangw6: That's right. I thought we have in-toto format for taskrun, so I just make it `GenericSignedAttestation_CONTENT_TYPE_UNSPECIFIED`.

Do you recommend that we differentiate taskrun and OCI first, and then use `GenericSignedAttestation_SIMPLE_SIGNING_JSON` for OCI and `GenericSignedAttestation_CONTENT_TYPE_UNSPECIFIED` for taskrun?

pkg/chains/storage/ca/ca.go
chuangw6: Should I delete this `conn.Close()` for now, and add that later if there is a big change in storage that allows us to close all connections?

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+	// build connection through grpc
+	// implicit uses Application Default Credentials to authenticate.
+	// Requires `gcloud auth application-default login` to work locally
+	ctx := context.Background()
+	creds, err := oauth.NewApplicationDefault(ctx, "https://www.googleapis.com/auth/cloud-platform")
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := grpc.Dial("dns:///containeranalysis.googleapis.com",
```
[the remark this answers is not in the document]
chuangw6: Sure, I've made endpoint configuration in a new commit under this pr, but for client, maybe we can hold off and implement that later.

pkg/chains/storage/ca/ca.go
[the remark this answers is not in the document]
chuangw6: Great point! Yes, storage backends are created before the objects get signed in the `SignTaskRun` function ([source](https://github.com/tektoncd/chains/blob/ae2c66b7814e565c7a100f3849b9ec15337f0759/pkg/chains/signing.go#L127-L172)).
I'll move this to where it gets used (inside `StorePayload` where objects have been signed already).

pkg/chains/storage/ca/ca.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	return resourceURI
+}
+
+// Given the TaskRun, retrieve the OCI image's URL.
+func (b *Backend) RetrieveOCIURI(opts config.StorageOpts) string {
+	images := artifacts.ExtractOCIImagesFromResults(b.tr, b.logger)
+
+	for _, image := range images {
+		ref := image.(name.Digest)
```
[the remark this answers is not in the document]
chuangw6: Yes, `ExtractOCIImagesFromResults` uses `name.NewDigest(fmt.Sprintf("%s@%s", img.url, img.digest))` to create image digest struct ([source](https://github.com/tektoncd/chains/blob/ae2c66b7814e565c7a100f3849b9ec15337f0759/pkg/artifacts/signable.go#L145)).

pkg/chains/storage/ca/ca.go
chuangw6: I'll add todo comment here stating that we will add other key refs later.

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,354 @@
...
+)
+
+// Backend is a storage backend that stores signed payloads in the storage that
+// is built on the top of grafeas i.e. container analysis.
+type Backend struct {
+	logger         *zap.SugaredLogger
+	tr             *v1beta1.TaskRun
+	client         pb.GrafeasV1Beta1Client
+	cfg            config.Config
+	occurrenceRefs []string // store occurrence IDs that are automatically generated during the time of creation
```
wlynch, line 54: I still think we need to change this behavior - we should be able to use the lookup functions without needing to statefully store the written occurrences at write time.

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,354 @@
...
+	// implicit uses Application Default Credentials to authenticate.
+	// Requires `gcloud auth application-default login` to work locally
+	creds, err := oauth.NewApplicationDefault(ctx, "https://www.googleapis.com/auth/cloud-platform")
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := grpc.Dial(cfg.Storage.Grafeas.Server,
+		grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{})),
+		grpc.WithDefaultCallOptions(grpc.PerRPCCredentials(creds)),
```
wlynch, line 69: We'll want to restrict the usage of Google credentials to known Google hosts - as a next step, we'll want to add support for specifying access tokens via secrets for general Grafeas servers.

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	for _, gvk := range gvks {
+		if len(gvk.Kind) == 0 {
+			continue
+		}
+		if len(gvk.Version) == 0 || gvk.Version == runtime.APIVersionInternal {
+			continue
+		}
+		obj.GetObjectKind().SetGroupVersionKind(gvk)
```
wlynch, line 311: Can we return the GVK instead of modifying the incoming Object? I'm a bit nervous about modifying the original content that we're using to sign.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,350 @@
...
+		pb.ListOccurrencesResponse{},
+		pb.Occurrence{},
+		pb.Resource{},
+		attestationpb.Details{},
+		attestationpb.Attestation{},
+		attestationpb.GenericSignedAttestation{},
+		commonpb.Signature{},
+		commonpb.Envelope{},
+		commonpb.EnvelopeSignature{},
+	)
```
wlynch, line 276: Typically it's not common to need to ignore all of these individually - usually I see a single [protocmp.Transform()](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp#Transform) is used instead (unless individual fields need to be ignored). Can this be used instead?

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,354 @@
...
+		result = append(result, occ)
+	}
+
+	return result, nil
+}
+
+// addTypeInformationToObject adds TypeMeta information to a runtime.Object based upon the loaded scheme.Scheme
+// inspired by: https://github.com/kubernetes/cli-runtime/blob/v0.19.2/pkg/printers/typesetter.go#L41
+// Souce: https://github.com/kubernetes/client-go/issues/308#issuecomment-700099260
+func addTypeInformationToObject(obj runtime.Object) error {
```
wlynch, line 298: We should add tests for non-trivial helper funcs like this as well.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,350 @@
...
+	wanted := &pb.ListOccurrencesResponse{
+		Occurrences: []*pb.Occurrence{
+			// occurrence for taskrun
+			{
+				// Occurrence Name will be automatically generated by grafeas server.
+				// In this fake grafeas server, we mock this behaviour by just using resource uri.
+				// In the real world, the occurrence name will be in the format of `projects/<PROJECT_NAME>/occurrences/<AUTO-GENERATED-ID>`.
+				// projects/my_project/occurrences/06d6e0d6-ee2b-4629-b44a-2188ac92eee4
+				Name: "/apis//namespaces/foo1//bar1@uid1",
+				// It's expected that the faked taskrun test does not have group version kind information.
```
wlynch, line 185: Shouldn't addTypeInformationToObject be filling this in?

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,350 @@
...
+	trOpts       config.StorageOpts
+	ociPayload   []byte
+	ociSignature string
+	ociOpts      config.StorageOpts
+}
+
+type testConfig struct {
+	name    string
+	args    args
+	wantErr bool
```
wlynch, line 52: I don't think this is used right now, but it would be good to add test cases for common error cases.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	// get uri
+	objectIdentifier := backend.retrieveResourceURI(opts)
+
+	// check signature
+	expect_signature := map[string][]string{objectIdentifier: []string{signature}}
+	got_signature, err := backend.RetrieveSignatures(ctx, opts)
+	if err != nil {
+		t.Fatal(err)
```
wlynch, line 154: nit: add some description to the error so it's clearer in logs what's failing here (e.g. similar to what you're doing with the StorePayload error above)

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	// close connection
+	conn.Close()
+}
+
+// test attestation storage and retrieval
+func testInterface(ctx context.Context, t *testing.T, test testConfig, backend Backend, payload []byte, signature string, opts config.StorageOpts) {
+	if err := backend.StorePayload(ctx, payload, signature, opts); (err != nil) != test.wantErr {
+		t.Errorf("Backend.StorePayload() error = %v, wantErr %v", err, test.wantErr)
```
wlynch, line 144:
```suggestion
		t.Fatalf("Backend.StorePayload() error = %v, wantErr %v", err, test.wantErr)
```

Fatal will fail the test and stop execution.
Error will fail the test and continue execution.

pkg/config/config.go
```diff
@@ -150,6 +163,12 @@ func defaultConfig() *Config {
 		Transparency: TransparencyConfig{
 			URL: "https://rekor.sigstore.dev",
 		},
+		// Storage: StorageConfigs{
+		// 	Grafeas: GrafeasConfig{
+		// 		NoteID: "tekton-chains",
+		// 		Server: "dns:///containeranalysis.googleapis.com",
+		// 	},
+		// },
```
wlynch, line 171: Remove?

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,350 @@
...
+				ociPayload:   []byte("oci payload"),
+				ociSignature: "oci signature",
+				ociOpts:      config.StorageOpts{Key: "cfe4f0bf41c8", PayloadFormat: formats.PayloadTypeSimpleSigning},
+			},
+			wantErr: false,
+		},
+	}
+
+	conn, client, err := setupConnection()
+
```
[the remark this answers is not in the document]
chuangw6: Good catch!

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,354 @@
...
+	"google.golang.org/grpc/credentials/oauth"
+	"google.golang.org/grpc/status"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	StorageBackendGrafeas = "grafeas"
+	ProjectNameFormat     = "projects/%s"
+	NoteNameFormat        = "projects/%s/notes/%s"
```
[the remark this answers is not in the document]
chuangw6: Yes!

pkg/chains/storage/grafeas/grafeas.go
chuangw6: That's a good point. IIUC I am thinking to check the suffix of the configured server ends with `googleapis.com` to make sure it's Google API endpoint.

The second point makes sense to me! Do we want to have this implemented in next pr or within this one?

wlynch: Another PR is fine - important thing is to at least restrict the creds down in this PR.

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: wow this is a great improvement. It works!

pkg/chains/storage/grafeas/grafeas_test.go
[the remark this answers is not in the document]
chuangw6: 76b5a8e6b93446a3690c3113f76182b5e61b1868

pkg/chains/storage/grafeas/grafeas.go
chuangw6: Great. this is the commit for the change 3849acdd9d641ea53d69b73ffe39ec0663447126 & cf900903f93e0fde060730961f6472b5548b18a4

pkg/chains/storage/grafeas/grafeas.go
chuangw6: Yes, I just found the correct method to get gvk. here is the commit a0c2680ac800e9c02c61ee5e1c287321e3e93363

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: solved! a0c2680ac800e9c02c61ee5e1c287321e3e93363

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: add test for tekton format (which should return error because we don't want grafeas backend to support tekton format payload.) 820eac28cf421f647b4d9b6b702409c928735bff

pkg/chains/storage/grafeas/grafeas.go
chuangw6: Thanks for proposing this. Yes this indeed needs to be changed for the change of creating backends `per process` in which we'll not be able to store data statefully for a specific taskrun.

Here is the commit 14d2f37c8240d8f0bdadb5d813a46db727650f2e

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,384 @@
...
+		ref := image.(name.Digest)
+		result = append(result, ref.Name())
+	}
+
+	return result
+}
+
+func checkTrustedHost(server string) error {
+	for _, t := range trustedHosts {
+		if strings.HasSuffix(server, t) {
```
wlynch, line 379: This won't work, since you could construct a URL like `example.com/googleapis.com`. Use https://pkg.go.dev/net/url#URL.Hostname instead (and for now, don't worry about supporting wildcards).

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,384 @@
...
+
+const (
+	StorageBackendGrafeas = "grafeas"
+	projectNameFormat     = "projects/%s"
+	noteNameFormat        = "projects/%s/notes/%s"
+)
+
+var trustedHosts = []string{
+	"googleapis.com",
+	// add more trusted hosts here
```
wlynch, line 47: We should add documentation about why this restriction is in place, and what needs to be done for other providers to be added.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,382 @@
...
+					},
+				},
+				payload:   []byte("taskrun payload"),
+				signature: "taskrun signature",
+				opts:      config.StorageOpts{Key: "taskrun.uuid", PayloadFormat: formats.PayloadTypeInTotoIte6},
+			},
+			wantErr: false,
+		},
+		{
+			name: "TEST 2: simplesining for oci, no error",
```
wlynch, line 100:
```suggestion
			name: "TEST 2: simplesigning for oci, no error",
```

wlynch: (also super nit - you can probably get rid of the `TEST #` prefixes)

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,384 @@
...
+	uriFilters = append(uriFilters, b.getTaskRunURI())
+
+	// step 2: find all occurrences by using ListOccurrences filters, and put them in result
+	for _, uriFilter := range uriFilters {
+		occs, err := b.findOccurrencesForCriteria(ctx, b.getProjectPath(), uriFilter)
+		if err != nil {
+			return result, err
+		}
+		result = append(result, occs...)
+	}
```
wlynch, line 313: This makes a ListRequest for every URI - could we construct a filter query that would fetch all of them in 1 request?

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,382 @@
...
+	var objectIdentifier string
+	switch opts.PayloadFormat {
+	case formats.PayloadTypeSimpleSigning:
+		objectIdentifier = backend.getOCIURI(opts)
+	case formats.PayloadTypeInTotoIte6:
+		objectIdentifier = backend.getTaskRunURI()
+	default:
+		// for other signing formats, grafeas backend will not support
+		// we set a fake identifier for testing only
+		objectIdentifier = "placeholder_uri"
```
wlynch, line 190: Can we t.Fatal instead?

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,382 @@
...
+				// i.e. projects/my_project/occurrences/06d6e0d6-ee2b-4629-b44a-2188ac92eee4
+				Name:     "/apis/tekton.dev/v1beta1/namespaces/foo1/TaskRun/bar1@uid1",
+				Resource: &pb.Resource{Uri: "/apis/tekton.dev/v1beta1/namespaces/foo1/TaskRun/bar1@uid1"},
+				NoteName: "projects/test-project/notes/test-note",
+				Details: &pb.Occurrence_Attestation{
+					Attestation: &attestationpb.Details{
+						Attestation: &attestationpb.Attestation{
+							Signature: &attestationpb.Attestation_GenericSignedAttestation{
+								GenericSignedAttestation: &attestationpb.GenericSignedAttestation{
+									// ContentType: for taskrun, this will be GenericSignedAttestation_CONTENT_TYPE_UNSPECIFIED, so this will not show up
```
wlynch, line 234: You should be able to still set this - GenericSignedAttestation_CONTENT_TYPE_UNSPECIFIED just happens to map to the default value.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,382 @@
...
+					},
+				},
+			},
+		},
+	}
+	got, err := client.ListOccurrences(ctx,
+		&pb.ListOccurrencesRequest{
+			// This is just a placeholder.
+			// ProjectID doesn't matter here because we assume there is only one project in the mocked server.
+			Parent: fmt.Sprintf("project/PLACEHOLDER"),
```
wlynch, line 298: Should we just omit this then?

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,382 @@
...
+		// mock how filter works in ListOccurrencesRequest
+		uriFilter := strings.Split(req.GetFilter(), "=")[1] // url filter that has quotes
+		uriFilter = uriFilter[1 : len(uriFilter)-1]         // remove quotes
+
+		for id, occ := range s.occurences {
+			if id == uriFilter {
+				occurrences = append(occurrences, occ)
+			}
+		}
+	}
```
wlynch, line 380: You should document how exactly this can be used - if you're implementing non-obvious logic like filtering, you may want to consider adding additional tests. Alternatively, you can break the read tests into 2 separate components: 1) to make sure we are properly handling returned content 2) make sure we're constructing the right request.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,382 @@
...
+							TaskRunResults: []v1beta1.TaskRunResult{
+								{Name: "IMAGE_DIGEST", Value: "sha256:cfe4f0bf41c80609214f9b8ec0408b1afb28b3ced343b944aaa05d47caba3e00"},
+								{Name: "IMAGE_URL", Value: "gcr.io/test/kaniko-chains1"},
+							},
+						},
+					},
+				},
+				payload:   []byte("oci payload"),
+				signature: "oci signature",
+				// the Key field must be the same as the first 12 chars of the image digest
```
wlynch, line 119: What happens if it's not? Is this something we should add additional validation for?

wlynch (changes requested):

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: the tests still pass if we omit this. just though it might be useful to put here to show how a `ListOccurrencesRequest` is created. what do you think about it?

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: thanks!

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: Thanks for the comment! In the implementation code of `chains.SignTaskRun` function, the key is set, and for oci image, it will be set to the first 12 chars of the image digest. So in this test, we just want to add this information here in case others might need to know in order to add more tests later.

I added more comments here in the latest version.

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: Great catch. Circular dependencies were introduced here. 
In the latest version, I added extra tests to make sure the mocked `ListOccurrences` function and its helper function of the filter logic can be trusted.

pkg/chains/storage/grafeas/grafeas.go
chuangw6: Thanks for pointing out this! I refactored this by constructing a chained filter query first and then sending the request once!

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: sgtm!

chuangw6:
hi @wlynch ,
Can you please review again if you get a chance? I'll try to see how to solve the issue of the re-request review button. 😅
Thanks a lot!

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,415 @@
...
+		t.Errorf("Wrong list of occurrences received for multiple filters, got=%s", cmp.Diff(got, wantedResponse, protocmp.Transform()))
+	}
+
+	// 3. test a single filter - return one occurrence
+	got, err = client.ListOccurrences(ctx, &pb.ListOccurrencesRequest{
+		Filter: `resourceUrl="gcr.io/test/kaniko-chains1@sha256:cfe4f0bf41c80609214f9b8ec0408b1afb28b3ced343b944aaa05d47caba3e00" OR resourceUrl="/apis/tekton.dev/v1beta1/namespaces/foo2/TaskRun/bar2@uid2"`,
+	})
+	if err != nil {
+		t.Fatal("Failed to call ListOccurrences. error ", err)
+	}
```
wlynch, line 233: We're overloading the StorePayload tests a bit too much here - it probably makes more sense to break out the "is the filter working properly" tests into their own TestFunc.

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,378 @@
...
+func NewStorageBackend(ctx context.Context, logger *zap.SugaredLogger, tr *v1beta1.TaskRun, cfg config.Config) (*Backend, error) {
+	// build connection through grpc
+	// implicit uses Application Default Credentials to authenticate.
+	// Requires `gcloud auth application-default login` to work locally
+	creds, err := oauth.NewApplicationDefault(ctx, "https://www.googleapis.com/auth/cloud-platform")
+	if err != nil {
+		return nil, err
+	}
+
+	server := "dns:///containeranalysis.googleapis.com"
```
wlynch, line 64: We should document this behavior in the README.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,415 @@
...
+
+type mockGrafeasV1Beta1Server struct {
+	// Embed for forward compatibility.
+	// Tests will keep working if more methods are added in the future.
+	pb.UnimplementedGrafeasV1Beta1Server
+
+	// Assume there is only one project for storing notes and occurences
+	occurences map[string]*pb.Occurrence
+	notes      map[string]*pb.Note
+}
```
wlynch, line 61: Move this definition below next to where the receiver funcs are defined. It helps readability to keep like things together.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -0,0 +1,415 @@
...
+
+var grafeasServer mockGrafeasV1Beta1Server
```
wlynch, line 63: This works for now since we only have 1 test, but if we ever have multiple tests run in parallel this single package level var will cause contention. I'd remove this and just create new instances in `setupConnection` instead.

wlynch (changes requested):
Looking good! Few more changes, mostly structural / documentation though!

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: Make sense to me!
How about we just leave the test for empty filter here?  And then we make another test func in which we test different filters but still use the two sample occurrences for testing (otherwise feel like there are too much lines for manually creating occurrence objects :D

pkg/chains/storage/grafeas/grafeas.go
chuangw6: For sure, it seems like readme file doesn't have a dedicated section about storage backends or configuration. I am thinking about adding config docs in [this markdown file](https://github.com/tektoncd/chains/blob/main/docs/config.md#storage-configuration). Sounds good?

wlynch (approved):
/lgtm

priyawadhwa (commented):
Nice! I'm not super familiar with grafeas, but would it be possible to write an integration test for it?

I'm wondering if we could deploy an instance of grafeas to the integration test cluster & test against that?

priyawadhwa:
> maybe squash the commits before merged?

 yah i think that's required!

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+
+	// Check if projectID is configured. If not, stop and return an error
+	if b.cfg.Storage.Grafeas.ProjectID == "" {
+		return errors.New("Project ID needs to be configured!")
+	}
+
+	// check if noteID is configured. If not, we give it a name as `tekton-<namespace>`
+	if b.cfg.Storage.Grafeas.NoteID == "" {
+		generatedNoteID := fmt.Sprintf("tekton-%s", b.tr.GetNamespace())
```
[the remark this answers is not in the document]
chuangw6: Hi @priyawadhwa ,
Thanks for proposing this. The current behaviour is that we want to use a single note (either configured by user or this auto-created one) to link to all occurrences created for all taskruns. 
If there are multiple TaskRuns in the same namespace, the note will not be overrode. Instead, it will return AlreadyExistsError when creating an existing note.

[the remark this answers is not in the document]
priyawadhwa: thanks for explaining!

chuangw6:
> > maybe squash the commits before merged?
> 
> yah i think that's required!

Done! Thanks for the heads up!

pkg/chains/storage/grafeas/grafeas.go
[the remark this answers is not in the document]
chuangw6: For the integration test, how about we add that in another pr considering the size of the current one?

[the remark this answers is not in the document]
priyawadhwa: sgtm

priyawadhwa (approved):

PR #390, opened 2022-03-03
add mongodb service support

06kellyjac:
# Changes

Added the docstore mongodb service support to tekton chains so it can be used with mongodb

Sometimes people are looking to integrate chains but need to keep (some of) their build systems on-prem.
MongoDB is a common database that organizations can setup and integration into chains is very simple via the docstore library used.

I've added a bit of documentation around the subject.

---

One shortcoming of this setup is the URI is just for the collection and you need to provide tekton-chains-controller with a mongodb connection URI via MONGO_SERVER_URL
people might also need to mount root certificate files

https://github.com/google/go-cloud/blob/master/docstore/mongodocstore/urls.go#L44

I'm open to other suggestions. Maybe we can pull in config and then setEnv from within the controller? But that feels a bit brittle
With the prospect of chains being added to the operator I'd imagine it could take config and patch the deployment as appropriate so shouldn't be too many problems here

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [X] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](CONTRIBUTING.md) for more details._

# Release Notes

```release-note
Added mongodb storage driver (see docs for integration guidance)
```


wlynch:
Generally LGTM!

If we want to make this more configurable within chains itself (i.e. without needing to modify the deployment / env vars), we could probably replicate what the [mongodb driver is doing](https://github.com/google/go-cloud/blob/docstore/mongodocstore/v0.24.0/docstore/mongodocstore/urls.go#L44-L54) from `docdb.NewStorageBackend` and pull the server URL from a Chains config value.

That said, I'm fine with this PR as is (we can always add this later).

wlynch:
@priyawadhwa Do we have e2e tests for docdb integrations, or are we just relying on gocloud for integration coverage here?

priyawadhwa:
yup we just rely on gocloud, no testing support at the moment

06kellyjac:
Rebased

wlynch:
/lgtm

Vendor diff makes me sad, but I don't think it's worth blocking this feature for.

@chuangw6 @ykass This may be another reason to split up the writer - pluggable attestation writer implementations 🤔

hectorj2f:
@wlynch Is this an alternative to use Grafeas ?

wlynch:
> @wlynch Is this an alternative to use Grafeas ?

Kinda? It's adding another driver to the existing `storage.docdb` support. Grafeas is just another storage location for attestations (along with GCS, OCI, DocDB, etc.)

priyawadhwa (approved):

PR #391, opened 2022-03-03
Add missing unit tests for oci StorePayload

Yongxuanzhang:
This PR mocks the docker registry and add some missing tests cases for StorePayload.
Prior to this commit, StorePayload in oci are not fully covered by unit
tests.


priyawadhwa:
Thank you for opening a PR, but I think this flow is covered by our integration test:

https://github.com/tektoncd/chains/blob/8b29855fca5574de429fb20234cf66aefe003119/test/e2e_test.go#L322

I'm just hesitant to add in unit tests which require servers. If this covers a different flow than the integration test, maybe we should move it there?

Yongxuanzhang:
> 
Hi, thank you for your comment! 😄 
I'm not sure if fake registry would be an overkill for the unit test, if it is then I'm also ok with it and happy to close this pr.
I feel like the unit test can provide a straightforward case to illustrate the input and output, and probably useful for debugging.

Yongxuanzhang:
> Thank you for opening a PR, but I think this flow is covered by our integration test:
> 
> https://github.com/tektoncd/chains/blob/8b29855fca5574de429fb20234cf66aefe003119/test/e2e_test.go#L322
> 
> I'm just hesitant to add in unit tests which require servers (it feels like overkill for a unit test). If this covers a different flow than the integration test, maybe we should move it there?

Hi, thank you for your comment! 😄 
I'm not sure if fake registry would be an overkill for the unit test, if it is then I'm also ok with it and happy to close this pr.
I feel like the unit test can provide a straightforward case to illustrate the input and output, and probably useful for debugging.

pkg/chains/storage/oci/oci_test.go
```diff
@@ -14,23 +14,118 @@ limitations under the License.
...
+			Identity: payload.Identity{
+				DockerReference: u.Host + "/task/" + tr.Name,
+			},
+			Image: payload.Image{
+				DockerManifestDigest: digest,
+			},
+			Type: payload.CosignSignatureType,
+		},
+	}
+	simplebytes, _ := json.Marshal(simple)
```
wlynch, line 100:
```suggestion
	simplebytes, err := json.Marshal(simple)
```

pkg/chains/storage/oci/oci_test.go
```diff
@@ -14,23 +14,118 @@ limitations under the License.
...
+			},
+		},
+		Predicate: slsa.ProvenancePredicate{},
+	}
+	intotoStatementBytes, err := json.Marshal(intotoStatement)
+	if err != nil {
+		t.Fatalf("failed to marshal: %v", err)
+	}
+
+	sampleIntotoStatementBytes, err := json.Marshal(in_toto.Statement{})
```
wlynch, line 125: nit: Since you're doing the json marshal for each message type, you could change `rawPayload []byte` -> `payload interface{}` and marshal it below!

wlynch (commented):
LGTM! Some minor comments.

/lgtm

pkg/chains/storage/oci/oci_test.go
Yongxuanzhang: yes! thanks!

wlynch:
> Thank you for opening a PR, but I think this flow is covered by our integration test:
> 
> https://github.com/tektoncd/chains/blob/8b29855fca5574de429fb20234cf66aefe003119/test/e2e_test.go#L322
> 
> I'm just hesitant to add in unit tests which require servers (it feels like overkill for a unit test). If this covers a different flow than the integration test, maybe we should move it there?

Missed this comment before! Sorry about that.

FWIW, I think the trade-off is worth it - right now AFAICT we don't have any unit tests that actually test a successful OCI upload, and it would be generally useful to add some for a fast `go test` feedback loop w/o requiring running the integration tests. Even though you're right in that it is a server that we're making network requests to, it's running on localhost and doesn't require external network, auth, etc. While it is slower, it's still pretty fast for a unit test (100 runs took ~3.5s). This is also a pattern that we've been using for [go-containerregistry](https://github.com/google/go-containerregistry/blob/f1fa40b162a1601a863364e8a2f63bbb9e4ff36e/pkg/v1/remote/image_test.go#L662-L669) and [pipelines](https://github.com/tektoncd/pipeline/blob/863ae571eb1d437463288933d85eeb1de83f6867/test/remote_test.go#L39-L41) unit tests.

wdyt? Happy to brainstorm other ideas! 😃

priyawadhwa:
SGTM! Let's go for it :)

priyawadhwa (approved):

PR #394, opened 2022-03-07
Remove usage of context.TODO

wlynch:
This PR removes usage of context.TODO/context.Background by plumbing
through a context through the respective functions.
This is a breaking library change, though hopefully one that is minimally
intrusive.

Also unexports a few methods in the Tekton storage backend that do not need to be exposed.


wlynch:
/hold 

There's actually not that many more instances of context.TODO outside of storage - I can tackle them in this PR.

priyawadhwa (approved):
🎉

PR #398, opened 2022-03-21
Create formatters and backends in configmap observer

chuangw6:
# Changes

This PR refactors the creation of all the formatters and backends that we need to create and store taskrun's payload. Instead of creating them per request in the `SignTaskRun` function, we create them per process. Specifically, they are created at the controller level. Only when the configmap is updated, they will be updated.

# Submitter Checklist

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
  functionality, content, code)
- [ ] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

### DIFF
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -47,19 +47,15 @@ type TaskRunSigner struct {
 	// The values are actual `Payloader` interfaces that can generate payload in different format from taskrun.
 	Formatters map[formats.PayloadType]formats.Payloader
 
-	// Signers: sign payload
-	// The keys are the names of the signers. {x509, kms}
-	// The values are key signers used to sign the payload.
-	Signers map[string]signing.Signer
-
 	// Backends: store payload and signature
 	// The keys are different storage option's name. {docdb, gcs, grafeas, oci, tekton}
 	// The values are the actual storage backends that will be used to store and retrieve provenance.
 	Backends          map[string]storage.Backend
+	SecretPath        string
 	Pipelineclientset versioned.Interface
 }
 
-func AllSigners(ctx context.Context, sp string, cfg config.Config, l *zap.SugaredLogger) map[string]signing.Signer {
+func allSigners(ctx context.Context, sp string, cfg config.Config, l *zap.SugaredLogger) map[string]signing.Signer {
 	all := map[string]signing.Signer{}
 	for _, s := range signing.AllSigners {
 		switch s {
@@ -131,6 +127,8 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
 		&artifacts.OCIArtifact{Logger: logger},
 	}
 
+	signers := allSigners(ctx, ts.SecretPath, cfg, logger)
+
 	rekorClient, err := getRekor(cfg.Transparency.URL, logger)
 	if err != nil {
 		return err
@@ -167,7 +165,7 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
 
 			// Sign it!
 			signerType := signableType.Signer(cfg)
-			signer, ok := ts.Signers[signerType]
+			signer, ok := signers[signerType]
 			if !ok {
 				logger.Warnf("No signer %s configured for %s", signerType, signableType.Type())
 				continue
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -34,7 +34,6 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	versioned "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	"go.uber.org/zap"
-	"k8s.io/client-go/kubernetes"
 	"knative.dev/pkg/logging"
 )
 
@@ -43,15 +42,24 @@ type Signer interface {
 }
 
 type TaskRunSigner struct {
-	KubeClient        kubernetes.Interface
+	// Formatters: format payload
+	// The keys are the names of different formatters {tekton, in-toto, simplesigning}. The first two are for TaskRun artifact, and simplesigning is for OCI artifact.
+	// The values are actual `Payloader` interfaces that can generate payload in different format from taskrun.
+	Formatters map[formats.PayloadType]formats.Payloader
+
+	// Signers: sign payload
+	// The keys are the names of the signers. {x509, kms}
+	// The values are key signers used to sign the payload.
+	Signers map[string]signing.Signer
+
+	// Backends: store payload and signature
+	// The keys are different storage option's name. {docdb, gcs, grafeas, oci, tekton}
+	// The values are the actual storage backends that will be used to store and retrieve provenance.
+	Backends          map[string]storage.Backend
 	Pipelineclientset versioned.Interface
-	SecretPath        string
 }
 
-// Set this as a var for mocking.
-var getBackends = storage.InitializeBackends
-
-func allSigners(ctx context.Context, sp string, cfg config.Config, l *zap.SugaredLogger) map[string]signing.Signer {
+func AllSigners(ctx context.Context, sp string, cfg config.Config, l *zap.SugaredLogger) map[string]signing.Signer {
 	all := map[string]signing.Signer{}
 	for _, s := range signing.AllSigners {
 		switch s {
@@ -77,7 +85,7 @@ func allSigners(ctx context.Context, sp string, cfg config.Config, l *zap.Sugare
 	return all
 }
 
-func allFormatters(cfg config.Config, l *zap.SugaredLogger) map[formats.PayloadType]formats.Payloader {
+func AllFormatters(cfg config.Config, l *zap.SugaredLogger) map[formats.PayloadType]formats.Payloader {
 	all := map[formats.PayloadType]formats.Payloader{}
 
 	for _, f := range formats.AllFormatters {
@@ -114,7 +122,6 @@ func allFormatters(cfg config.Config, l *zap.SugaredLogger) map[formats.PayloadT
 
 // SignTaskRun signs a TaskRun, and marks it as signed.
 func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) error {
-	// Get all the things we might need (storage backends, signers and formatters)
 	cfg := *config.FromContext(ctx)
 	logger := logging.FromContext(ctx)
 
@@ -123,15 +130,6 @@ func (ts *TaskRunSigner) SignTaskRun(ctx context.Context, tr *v1beta1.TaskRun) e
 		&artifacts.TaskRunArtifact{Logger: logger},
 		&artifacts.OCIArtifact{Logger: logger},
 	}
-
-	// Storage
-	allBackends, err := getBackends(ctx, ts.Pipelineclientset, ts.KubeClient, logger, tr, cfg)
-	if err != nil {
-		return err
-	}
-
-	signers := allSigners(ctx, ts.SecretPath, cfg, logger)
-	allFormats := allFormatters(cfg, logger)
 
 	rekorClient, err := getRekor(cfg.Transparency.URL, logger)
 	if err != nil {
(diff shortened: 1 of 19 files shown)


pkg/chains/factory.go
```diff
@@ -0,0 +1,131 @@
+package chains
```
Yongxuanzhang, line 1: should add license above

chuangw6: Thanks for pointing out this!

pkg/chains/storage/oci/oci_test.go
```diff
@@ -70,14 +68,24 @@ func TestBackend_StorePayload(t *testing.T) {
...
+			if err := b.StorePayload(ctx, tt.fields.tr, tt.args.rawPayload, tt.args.signature, tt.args.storageOpts); (err != nil) != tt.wantErr {
 				t.Errorf("Backend.StorePayload() error = %v, wantErr %v", err, tt.wantErr)
 			}
 		})
 	}
 }
+
+func setupMocks() func() {
+	oldAuth := getAuthenticator
+	getAuthenticator = func(_ context.Context, _ *v1beta1.TaskRun, _ kubernetes.Interface) (remote.Option, error) {
```
wlynch, line 85: Can we setup this per-backend instead of at the package level?

pkg/reconciler/taskrun/controller.go
```diff
@@ -31,15 +31,36 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
-			SecretPath:        SecretPath,
-		},
+		TaskRunSigner: tsSigner,
 	}
+
 	impl := taskrunreconciler.NewImpl(ctx, c, func(impl *controller.Impl) controller.Options {
-		cfgStore := config.NewConfigStore(logger)
+		cfgStore := config.NewConfigStore(logger, func(name string, value interface{}) {
+			// get updated config
+			tsSigner.Config = *value.(*config.Config)
```
wlynch, line 49: If the config is invalid, don't continue - log and return (perhaps even shutdown?). We don't want the absence of a config to start causing bad behavior.

pkg/reconciler/taskrun/controller.go
```diff
@@ -31,15 +31,36 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
+			formatters, signers, backends, err := chains.AbstractFactory(ctx, pipelineClient, SecretPath, kubeClient, tsSigner.Config, logger)
+			if err != nil {
+				logger.Error(err)
+			}
+
+			// update the corresponding fields in taskrunSigner
+			tsSigner.AllFormatters = formatters
+			tsSigner.AllSigners = signers
+			tsSigner.AllBackends = backends
+		})
```
wlynch, line 61: We should add a test for this.

test/test_utils.go
```diff
@@ -239,7 +240,7 @@ func verifySignature(ctx context.Context, t *testing.T, c *clients, tr *v1beta1.
 	logger := logging.FromContext(ctx)
 
 	// Initialize the backend.
-	backends, err := chainsstrorage.InitializeBackends(ctx, c.PipelineClient, c.KubeClient, logger, tr, *cfg)
+	_, _, backends, err := chains.AbstractFactory(ctx, c.PipelineClient, taskrun.SecretPath, c.KubeClient, *cfg, logger)
```
wlynch, line 243: Perhaps we should keep the InitializeBackends func? No need to create all of the clients / signers / formatters if we only care about the backends.

wlynch (changes requested):

test/test_utils.go
chuangw6: `InitializeBackends` function still exists, but the name was changed to unexported `allBackends(...)` to keep consistent and was moved to `factory.go`.
How about making `AllBackends` exported and just call `AllBackends` here instead of using `AbstractFactory` to initialize all?

pkg/chains/storage/oci/oci_test.go
chuangw6: When we initialize a k8s Keychain to override default one, we need namespace and service account specified in taskrun spec. And different taskrun might have different spec for namespace and service account to use. 
So IIUC we might want to setup this for per taskrun? Please correct me if I'm wrong.

wlynch: Yup, but here we're overriding a package level func, which will affect all OCI backends (and worst case if tests run in parallel this may cause issues).

Similar to how you're defining clients per backend now, you can do the same for the authenticator func e.g.

```go
type Backend struct {
  ...
  getAuthenticator func(context.Context, *v1beta1.TaskRun, kubernetes.Interface) (remote.Option, error)
}
```

and then within tests you can just set the func to the no-op implementation.

test/test_utils.go
wlynch: SGTM

pkg/chains/factory.go
```diff
@@ -0,0 +1,144 @@
...
+		case docdb.StorageTypeDocDB:
+			docdbBackend, err := docdb.NewStorageBackend(ctx, logger, cfg)
+			if err != nil {
+				return nil, err
+			}
+			backends[backendType] = docdbBackend
+		}
+	}
+	return backends, nil
+}
```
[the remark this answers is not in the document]
chuangw6: Yeah that's a totally reasonable concern though I didn't change the implementation of `allSigners`, `allFormatters ` (from signing.go), or `allBackends` (from storage.go though the name is changed from `InitializeBackends` to `allBackends` for consistency).
The reason why I pull them out is that I though they will be created per process level so it might make more sense to put them in factory.go. But it makes sense to me to have them in  `signing.go` too.

Do we want to put all three functions back to where they were, and call them separately in controller.go instead of using a single factory function?

pkg/chains/factory.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/tektoncd/chains/pkg/chains/storage/gcs"
+	"github.com/tektoncd/chains/pkg/chains/storage/oci"
+	tekton_storage "github.com/tektoncd/chains/pkg/chains/storage/tekton"
+	"github.com/tektoncd/chains/pkg/config"
+	versioned "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	"go.uber.org/zap"
+	"k8s.io/client-go/kubernetes"
+)
+
+func AbstractFactory(ctx context.Context, pipelineClientset versioned.Interface, secretPath string, kc kubernetes.Interface, cfg config.Config, logger *zap.SugaredLogger) (map[formats.PayloadType]formats.Payloader, map[string]signing.Signer, map[string]storage.Backend, error) {
```
Yongxuanzhang, line 38: what are the benefits of putting all of these in the AbstractFactory? why not call them separately?

chuangw6: Just thinking it looks cleaner to have a single factory function to return all 3 objects we need to sign a taskrun. not sure if it's a common use case.
The other option is to keep `allFormatters` and `allSigners` as is in the `signing.go` as @wlynch suggested [here](https://github.com/tektoncd/chains/pull/398#discussion_r831569414). But those functions will be called in `controller.go` not in the `signing.go` file so we might need to export them. 

@wlynch what do you think about this?

Yongxuanzhang: the only issue I can see is to couple these together, so if you only want to get backends somewhere else you would need to call the function and process all of them

pkg/chains/signing_test.go
```diff
@@ -160,25 +159,31 @@ func TestTaskRunSigner_SignTaskRun(t *testing.T) {
...
 						StorageBackend: sets.NewString(tt.configuredBackend),
 						Signer:         "x509",
 					},
 				},
-			})
+			}
+			ctx = config.ToContext(ctx, cfg.DeepCopy())
 
+			logger := logging.FromContext(ctx)
+			SecretPath := "./signing/x509/testdata/"
```
wlynch, line 180: Typically local variables are lower case to avoid confusion with package level exported variables.
```suggestion
			secretPath := "./signing/x509/testdata/"
```

pkg/chains/signing_test.go
```diff
@@ -160,25 +159,31 @@ func TestTaskRunSigner_SignTaskRun(t *testing.T) {
...
+			formatters, signers, fakeBackends := AllFormatters(*cfg, logger), AllSigners(ctx, SecretPath, *cfg, logger), fakeAllBackends(tt.backends)
 			ts := &TaskRunSigner{
+				AllFormatters:     formatters,
+				AllSigners:        signers,
+				AllBackends:       fakeBackends,
```
wlynch, line 185:
```suggestion
			ts := &TaskRunSigner{
				AllFormatters:     AllFormatters(*cfg, logger),
				AllSigners:        AllSigners(ctx, SecretPath, *cfg, logger),
				AllBackends:       fakeAllBackends(tt.backends),
```

pkg/chains/storage/docdb/docdb.go
```diff
@@ -48,26 +47,21 @@ type SignedDocument struct {
...
 		logger: logger,
-		tr:     tr,
 		coll:   coll,
-	}
+	}, nil
 }
 
 // StorePayload implements the Payloader interface.
-func (b *Backend) StorePayload(ctx context.Context, signed []byte, signature string, opts config.StorageOpts) error {
+func (b *Backend) StorePayload(ctx context.Context, tr *v1beta1.TaskRun, signed []byte, signature string, opts config.StorageOpts) error {
```
wlynch, line 64: nit: here and elsewhere, if the variable isn't being used use `_` as the variable name to make it clear that this isn't important for this func. 

```suggestion
func (b *Backend) StorePayload(ctx context.Context, _ *v1beta1.TaskRun, signed []byte, signature string, opts config.StorageOpts) error {
```

pkg/chains/storage/gcs/gcs.go
```diff
@@ -181,18 +179,18 @@ func (b *Backend) retrieveObject(ctx context.Context, object string) (string, er
 	return string(payload), nil
 }
 
-func (b *Backend) sigName(opts config.StorageOpts) string {
-	return fmt.Sprintf(SignatureNameFormat, b.tr.Namespace, b.tr.Name, opts.Key)
+func (b *Backend) sigName(tr *v1beta1.TaskRun, opts config.StorageOpts) string {
+	return fmt.Sprintf(SignatureNameFormat, tr.Namespace, tr.Name, opts.Key)
```
wlynch, line 183: here and elsewhere: you no longer need `Backend` anymore to implement these.

```suggestion
func sigName(tr *v1beta1.TaskRun, opts config.StorageOpts) string {
	return fmt.Sprintf(SignatureNameFormat, tr.Namespace, tr.Name, opts.Key)
```

pkg/chains/storage/storage_test.go
```diff
@@ -1,68 +0,0 @@
...
-			gotTypes := []string{}
-			for _, g := range got {
-				gotTypes = append(gotTypes, g.Type())
-			}
-			if !reflect.DeepEqual(gotTypes, tt.want) {
-				t.Errorf("InitializeBackends() = %v, want %v", gotTypes, tt.want)
-			}
-		})
-	}
-}
```
wlynch, line 68: Are we certain that this is all captured by the existing tests? Do we need to do anything extra for the "multi" test case?

wlynch (changes requested):

pkg/chains/storage/storage_test.go
chuangw6: Thanks for pointing out this. In previous implementation, I had a `factory.go` file and removed this test storage file. 
Just recovered this file, and added more individual tests and more items to `multi` test :D

pkg/chains/storage/gcs/gcs.go
chuangw6: Good point! 👍

pkg/chains/storage/docdb/docdb.go
chuangw6: Yes, that makes things clear! Thanks! 👍

pkg/chains/signing.go
```diff
@@ -43,15 +42,13 @@ type Signer interface {
 }
 
 type TaskRunSigner struct {
-	KubeClient        kubernetes.Interface
+	AllFormatters     map[formats.PayloadType]formats.Payloader
+	AllSigners        map[string]signing.Signer
+	AllBackends       map[string]storage.Backend
```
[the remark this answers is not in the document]
chuangw6: Make sense to me! Changed and godoc documentation is added. Thanks!

priyawadhwa (approved):
Nice!!

PR #409, opened 2022-03-31
Use sigstore/scaffolding github actions now that it's out there.

vaikas:
Now that sigstore/scaffolding has a reusable action use that rather than manully install scaffolding.
Also use chainguard-dev reusable action for gathering failure information.


priyawadhwa (approved):
nice!

Issue #426, opened 2022-04-21
Grafeas unit tests are flaky


priyawadhwa:
The grafeas `[pkg/chains/storage/grafeas TestBackend_ListOccurrences]` unit test is flaky:

```
    grafeas_test.go:85: Wrong list of occurrences received for empty filter, got=  (*grafeas_go_proto.ListOccurrencesResponse)(Inverse(protocmp.Transform, protocmp.Message{
          }))
```

an [example failure log](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_chains/420/pull-tekton-chains-unit-tests/1517164418503282689/) provided.

cc @chuangw6 who added this in, do you have any ideas why it might be flaky?

chuangw6:
Hi @priyawadhwa ,
Thanks for reporting this. Just digged into this and found out it's because when retrieving all occurrences stored as a map in the fake server, the order retrieved sometimes is not same as the insertion order (as Go maps do not maintain the insertion order). So I added the code to sort all occurrences by resource uri when calling `ListOccurrences`. This will make sure the test is not flaky anymore.

savitaashture:
Hi @chuangw6 and @priyawadhwa 
even we see chains nightly failing here https://dashboard.dogfooding.tekton.dev/#/namespaces/tekton-nightly/pipelineruns/chains-release-nightly-5wt2l?pipelineTask=unit-tests&step=unit-test
because of flakiness

priyawadhwa:
Thank you @chuangw6 for looking into it! @savitaashture hopefully tonight's build passes now that the fix is about to be merged 🤞🏽

chuangw6:
Yeah, now it should fix the issue. From the log, I see it failed because of the same issue I just mentioned. @savitaashture

dprotaso:
This consistently fails here: https://github.com/knative/pkg/runs/6265134542?check_suite_focus=true

All that check is doing is testing the latest `knative.dev/pkg` with the latest chains version

```
--- FAIL: TestInitializeBackends (0.01s)
    logger.go:130: 2022-05-02T22:24:40.280Z	DEBUG	storage/storage_test.go:81	Backend: map[]
    logger.go:130: 2022-05-02T22:24:40.280Z	DEBUG	storage/storage_test.go:81	Backend: map[tekton:0xc000710348]
    --- FAIL: TestInitializeBackends/gcs (0.01s)
[...]

PR #427, opened 2022-04-22
Fix flaky grafeas unit test

chuangw6:
Fix the issue reported in #426


priyawadhwa (approved):
Thanks for fixing this so quickly!

PR #432, opened 2022-04-26
Change SCC from anyuid to nonroot

### LINKED ISSUE #366: Question: Use of anyuid for tekton-chains-controller
Hi, from looking at the [Installation on OpenShift](https://github.com/tektoncd/chains#installation-on-openshift) section of the README it states "To install Tekton Chains on OpenShift, you must first apply the `anyuid` security context constraint to the `tekton-chains-controller service account`" can you please confirm why?

From looking at the [deployment manifest](https://github.com/tektoncd/chains/blob/main/config/100-deployment.yaml) I can see the security context being as follows:
```
    spec:
      serviceAccountName: tekton-chains-controller
      containers:
      - name: tekton-chains-controller
      ...
      ...
        securityContext:
          allowPrivilegeEscalation: false
          # User 65532 is the distroless nonroot user ID
          runAsUser: 65532
          runAsGroup: 65532
```

If the reason is due to `runAsUser: 65532` and `runAsGroup: 65532` then this can be achieved with the SCC `nonroot`, which doesn't permit the use of `root` user (UID=0) which the `anyuid` SCC does.

concaf:
This commit changes the SCC required to run Chains on OpenShift to
nonroot, since anyuid is no longer required.

Fix #366

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ for more information.
 
    ```bash
    oc new-project tekton-chains
-   oc adm policy add-scc-to-user anyuid -z tekton-chains-controller
+   oc adm policy add-scc-to-user nonroot -z tekton-chains-controller
    ```
 
 1. Install Tekton Chains:


concaf:
@priyawadhwa @vdemeester PTAL 🙃

vdemeester (approved):
/meow

priyawadhwa (approved):

PR #438, opened 2022-05-03
Disable unit test that requires GCP credentials

imjasonh:
This causes tests to fail in environments where credentials aren't set up.

https://github.com/tektoncd/chains/issues/426#issuecomment-1115516150

Unit tests shouldn't rely on being able to reach out to external services.

cc @dprotaso


dprotaso:
related: https://github.com/knative/pkg/issues/2499

/lgtm

PR #444, opened 2022-05-11
Allow TaskRun payloads other than IMAGE_URL

concaf:
Prior to this commit, in the `tekton` backend, a payload or a signature
was considered valid only if the TaskRun status had `IMAGE_URL` in the
results. This is incorrect since Chains considers both, `IMAGE_URL` and
`IMAGE_DIGEST` as results of interest. Moreover, Chains considers
PipelineResources to indicate an output of interest as well.

This commit removes checking for `IMAGE_URL` in the TaskRun status to
validate payloads, instead relies solely on the presence of a payload
annotation on the TaskRun given that it's only added when Chains has
signed something (after validation anyway).

### DIFF
--- a/pkg/chains/storage/tekton/tekton.go
+++ b/pkg/chains/storage/tekton/tekton.go
@@ -17,8 +17,6 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
-	"strings"
-
 	"github.com/tektoncd/chains/pkg/config"
 
 	"github.com/tektoncd/chains/pkg/patch"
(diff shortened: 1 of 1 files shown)


concaf:
This PR is more of a question, TBH!

I am facing this in the CLI since Chains is being used as a library there.
Particularly this snippet:

https://github.com/tektoncd/cli/blob/947a78dadd7ea49787597a04ea94ce57f24238ec/pkg/cmd/chain/payload.go#L96-L104

I am testing `tkn chain payload <taskrun>` on the [TaskRun](https://raw.githubusercontent.com/tektoncd/chains/main/examples/taskruns/task-output-image.yaml) that is being used in Chains' [getting started tutorial](https://github.com/tektoncd/chains/blob/main/docs/tutorials/getting-started-tutorial.md). `tkn` is unable to find any payload on the TaskRun since it does not contain `IMAGE_URL` even though it does contain the correct payload annotation i.e. `chains.tekton.dev/payload-taskrun-adc185bc-38cf-45f7-83f4-c751c39c4d8f`.

priyawadhwa:
I think @rgreinho might know a bit more about the CLI codebase and can weigh in here!

concaf:
@priyawadhwa sure, thanks. However, do you think that this PR makes sense from a Chains point of view? As in, do you think that `RetrievePayloads()` and `RetrieveSignatures()` methods should look for _only_ `IMAGE_URL` like it is the case today?

lcarva:
IMO this change makes sense because not every TaskRun produces an image.

priyawadhwa (approved):

PR #449, opened 2022-05-16
Fix default params in attestation invocation

lcarva:
Default parameters should always be overwritten by values from the
taskrun spec. This is how Tekton Pipelines behaves. The attestation
should also follow the same rule.

This also fixes zero-value default value representation, i.e. an empty
string or an empty array.

The representation of parameter values is now also consistent. Prior to
this commit, %v on the Param was used for default values while
.StringVal or .ArrayVal was used for non-default values.


pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -114,23 +114,38 @@ func metadata(tr *v1beta1.TaskRun) *slsa.ProvenanceMetadata {
...
-	}
-	// add params
+
+	// get implicit parameters from defaults
 	if ts := tr.Status.TaskSpec; ts != nil {
 		for _, p := range ts.Params {
 			if p.Default != nil {
-				v := p.Default.StringVal
-				if v == "" {
+				// TODO: Consider using p.Default.MarshalJSON()
```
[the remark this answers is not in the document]
lcarva: Ok, I added a commit to do this. Let me know what you think. I can easily drop it if necessary.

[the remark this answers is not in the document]
lcarva: @priyawadhwa, please take a look when you have a chance :pray:

lcarva:
@priyawadhwa, @wlynch ptal :pray:

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -84,7 +84,7 @@ func (i *InTotoIte6) generateAttestationFromTaskRun(tr *v1beta1.TaskRun) (interf
 				ID: i.builderID,
 			},
 			BuildType:   tektonID,
-			Invocation:  invocation(tr),
+			Invocation:  invocation(tr, i.logger),
```
wlynch, line 87: Instead of passing through a logger, let's just return back the error.

pkg/chains/formats/intotoite6/intotoite6_test.go
```diff
@@ -66,10 +66,10 @@ func TestCreatePayload1(t *testing.T) {
 			},
 			Invocation: slsa.ProvenanceInvocation{
 				Parameters: map[string]string{
```
wlynch, line 68: I wonder if we should use `map[string]ArrayOrString` (https://pkg.go.dev/github.com/tektoncd/pipeline@v0.36.0/pkg/apis/pipeline/v1beta1#ArrayOrString) here instead and not bother with the string encoding. 🤔  wdyt?

wlynch (commented):
Overall looks good! Few small things.

pkg/chains/formats/intotoite6/intotoite6_test.go
lcarva: Yeah, that's a better approach. I'll give that a shot.

pkg/chains/formats/intotoite6/intotoite6.go
lcarva: Do we want the whole attestation process to fail if that's the case? That makes sense to me, just want to confirm intended behavior.

pkg/chains/formats/intotoite6/intotoite6_test.go
lcarva: Done. Here's an example from a dummy task I used:
```
🐚 cosign download attestation quay.io/lucarval/minimal-container:latest | jq '.payload | @base64d | fromjson | .predicate.invocation.parameters'
{
  "BUILDER_IMAGE": "quay.io/buildah/stable:v1.18.0",
  "BUILD_EXTRA_ARGS": "",
  "CONTEXT": ".",
  "DOCKERFILE": "./Dockerfile",
  "FORMAT": "oci",
  "IMAGE": "quay.io/lucarval/minimal-container:latest",
  "PUSH_EXTRA_ARGS": "",
  "SKIP_PUSH": "false",
  "STORAGE_DRIVER": "vfs",
  "SUPER_ARRAY": [
    "hello",
    "array"
  ],
  "TLSVERIFY": "true"
}
```

lcarva:
@priyawadhwa, need an approval now :pray:

priyawadhwa (approved):
LGTM

PR #453, opened 2022-05-23
Switch to Grafeas v1 API in grafeas storage

chuangw6:
## Background
- Tekton Chains captures two types of artifacts: OCI artifacts, TaskRun itself.
- For OCI artifacts: Tekton Chains generates `simplesigning` attestation for it.
- For TaskRun artifacts: Tekton Chains generates `intoto` attestation with SLSA 0.2 predicate for it. 
_(Chains also supports `tekton` format for taskrun artifact, but we want grafeas backend storage to only accept `intoto` format)_

## Changes Summary:
- Switch from Grafeas v1beta1 api to [Grafeas v1 api](https://pkg.go.dev/github.com/grafeas/grafeas@v0.2.2-0.20220620154915-0ec8596e37d3/proto/v1/grafeas_go_proto).
- Store `simplesigning` attestation in grafeas ATTESTATION occurrence. (see the illustration diagram below)
- Store `intoto` attestation in BUILD occurrence. Prior to this, intoto attestation was also stored in the ATTESTATION occurrence. (see the illustration diagram below)
- If a taskrun generates X images
  - X number of ATTESTATION occurrences will be created to store the X `simplesigning` payload, and the [`resourceUri`](https://github.com/grafeas/grafeas/blob/0ec8596e37d3d853ef0ab8318f69b112cdc161ee/proto/v1/grafeas.proto#L198-L201) of the ATTESTATION occurrences will be set to the image identifier `IMAGE_URL@IMAGE_DIGEST`.
  - X number of BUILD occurrence will be created in order to associate the taskrun `intoto` payload with every single image generated from the taskrun, and the [`resourceUri`](https://github.com/grafeas/grafeas/blob/0ec8596e37d3d853ef0ab8318f69b112cdc161ee/proto/v1/grafeas.proto#L198-L201) will be also set to `IMAGE_URL@IMAGE_DIGEST`.
- Grafeas Note:
  > since a grafeas occurrence has to be linked to a grafeas note, we need to create two notes (attestation note and build note)
  - create ATTESTATION note to link ATTESTATION occurrence.
  - create BUILD note to link BUILD occurrence.
  - user-configured note name will be added a suffix ('-simplesigning'
    or '-intoto') to differentiate the two note types.

Screen Shot 2022-06-27 at 18 15 36

Screen Shot 2022-06-28 at 17 07 33


chuangw6:
cc @wlynch

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -186,117 +184,191 @@ func (b *Backend) Type() string {
...
+	if err := json.Unmarshal(payload, &in); err != nil {
 		return nil, err
 	}
 
-	occurrence := &pb.Occurrence{
-		Resource: &pb.Resource{
-			Uri: uri,
+	intotostatement, err := grafeasutil.ToProto(&in)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to convert to Grafeas proto: %v", err)
```
wlynch, line 309:
```suggestion
		return nil, fmt.Errorf("Unable to convert to Grafeas proto: %w", err)
```

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -186,117 +184,191 @@ func (b *Backend) Type() string {
...
+func (b *Backend) createBuildOccurrence(ctx context.Context, tr *v1beta1.TaskRun, payload []byte, signature string, uri string) (*pb.Occurrence, error) {
+	in := intoto.ProvenanceStatement{}
+	if err := json.Unmarshal(payload, &in); err != nil {
 		return nil, err
 	}
 
-	occurrence := &pb.Occurrence{
-		Resource: &pb.Resource{
-			Uri: uri,
+	intotostatement, err := grafeasutil.ToProto(&in)
```
wlynch, line 307: nit: 
```suggestion
	pb, err := grafeasutil.ToProto(&in)
```

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -186,117 +184,191 @@ func (b *Backend) Type() string {
...
+				Details:     occurrenceDetails,
+				Envelope:    envelope,
+			},
+		},
+	)
+}
+
+func (b *Backend) createBuildOccurrence(ctx context.Context, tr *v1beta1.TaskRun, payload []byte, signature string, uri string) (*pb.Occurrence, error) {
+	in := intoto.ProvenanceStatement{}
+	if err := json.Unmarshal(payload, &in); err != nil {
```
wlynch, line 303: It'd be interesting to just pass the intoto struct reference around rather than generating the intoto statement -> serializing it -> de-serializing it -> converting it to grafeas.

Might be worth looking into to simplify things (we can also do this in another PR if needed)

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -186,117 +184,191 @@ func (b *Backend) Type() string {
...
+			return nil, err
+		}
+		occs = append(occs, occ)
+		return occs, nil
 	}
 
-	return nil
+	// create Occurrence_Build for TaskRun
+	for _, uri := range allURIs {
+		occ, err := b.createBuildOccurrence(ctx, tr, payload, signature, fmt.Sprint("https://", uri))
```
wlynch, line 254: Not sure if we should do this 🤔 

What kind of URIs are we expecting here, and why do we need to append the https (e.g. if it's needed, why isn't it already present)?

wlynch (changes requested):

pkg/chains/storage/grafeas/grafeas.go
chuangw6: > What kind of URIs are we expecting here?

URIs are the identifiers of the container images built from the taskrun (`image_url@image_digest`).

> Not sure if we should do this 🤔

The reason why we create build occurrence per image built with identical intoto payload and signature but differing in the `resourceUri` field is that we want to enable users to retrieve provenance for a particular artifact built. 

For example, if a taskrun produces two images identified by `foo` and `bar`, we want to have the provenance associated with each of them because it's the same taskrun that produces them. This would allow users/provenance consumers to retrieve the provenance for a specific artifact by filtering the `resourceUri` field that is set to be the image identifier.

> why do we need to append the https (e.g. if it's needed, why isn't it already present)?

For attestation occurrence that stores simplesigning attestation for OCI artifact, the resourceUri is just set to be `image_url@image_digest`. 
For build occurrence that contains intoto attestation for the taskrun, we want to add a prefix `https://` to differentiate the build occurrence resourceUri and attestation occurrence resourceUri.

Example: taskrun `X` generates a single image artifact `URL@Digest`. 
- If users want to see the image simplesigning attestation, they can use `URL@Digest` as the filter of [ListOccurrences](https://cloud.google.com/container-analysis/docs/reference/rest/v1/projects.occurrences/list)
- If users want to see the provenance (how the image was built), they can use `https://URL@Digest` as the filter to retrieve the taskrun `X`'s intoto provenance.

> why isn't it already present?
Because prior to this, we use [K8s resource uri](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-urist) as the resourceUri for taskrun payload. And that doesn't work for consumers i.e. BinAuthz to look up occurrences because they only know the identifier of the deployed artifact at deployment time.


These are great questions. Please let me know if that makes sense. Feel free to follow up if you have more questions.

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -328,57 +398,27 @@ func (b *Backend) findOccurrencesForCriteria(ctx context.Context, projectPath st
...
-		tr.GetGroupVersionKind().GroupVersion().String(),
-		tr.GetNamespace(),
-		tr.GetGroupVersionKind().Kind,
-		tr.GetName(),
-		string(tr.UID),
-	)
-}
+// retrieveOCIURIs returns the uri of a specific OCI artifact and the uris of all the images generated by the taskrun
+// If it's taskrun payload & sig (judged from config.StorageOpts), the first returned value will be empty string.
+func (b *Backend) retrieveOCIURIs(tr *v1beta1.TaskRun, opts config.StorageOpts) (string, []string) {
```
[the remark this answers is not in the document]
chuangw6: I was trying to simplify the way we get the specific image uri when OCI artifact payload (simplesigning) and signatures are passed in, and the way to get the URIs of all images built from a taskrun when Taskrun artifact payload (intoto) are passed in. Basically, unify how to retrieve them into a single function.

If OCI artifact payload is passed, the first returned value will be the image uri of that image, and the second returned value contains the URIs of all images built by the taskrun.
If taskrun artifact payload is passed, the first returned value will be empty string, and the second returned value also contains the URIs of all images built by the taskrun.


> why not just return []string?

we can do this. But when we need to get uri for a specific image artifact, we still need a filtering process somewhere else (either another helper function or just inline).

Let me know what you think. I am happy with both.

pkg/chains/storage/grafeas/grafeas.go
chuangw6: Yeah, agree that for grafeas backend, there are extra steps as you just described.

If we want to simplify that, we might need to change the function `StorePayload`'s signature. But the sacrifice is that other backend options have to serialize when `StorePayload` is called, right?

pkg/chains/storage/grafeas/grafeas.go
chuangw6: `pb` name conflicts with the short name for the imported grafeas proto pkg :D
Changed from `intotostatement` to `pbf`. Please let me know if you have recommendations for the name.

pkg/chains/storage/grafeas/grafeas.go
chuangw6: Discussed offline with @sherzberg-1 about this. Thanks @sherzberg-1 for the help and sharing the knowledge!

Just realized that here we can also use `IMAGE_URL@IMAGE_DIGEST` for the build occurrence without having `https://` as the prefix. In other words, if a taskrun generates an image called `foo@bar`, we create one attestation occurrence with `foo@bar` as the resourceUri field, and one build occurrence with `foo@bar` as the resourceUri field too. For consumers, they can filter on both the `resourceUri` and the `kind` field to look up build OR attestation occurrence based on their needs. 
nit: two occurrences can have same resourceUri as long as they are linked to different grafeas notes.

Let me know what you think. @wlynch Thanks!

pkg/chains/storage/grafeas/grafeas.go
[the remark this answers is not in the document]
wlynch: Having trouble following why this behavior would change based on the attestation format - wouldn't the URIs be the same in either case?

pkg/chains/storage/grafeas/grafeas.go
wlynch: Yeah... I think that's okay though? Seems reasonable to keep it as it's full spec and only serialize when we actually make the request. Fine for this now, but let's look into cleaning this up in another PR.

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -186,117 +184,191 @@ func (b *Backend) Type() string {
...
-								PublicKeyId: b.cfg.Signers.KMS.KMSRef, // TODO: currently we only support storing kms keyID, will add other keys' ids later i.e. k8s secret, fulcio
-							},
-						},
-					},
+		Attestation: &pb.AttestationOccurrence{
+			SerializedPayload: payload,
+			Signatures: []*pb.Signature{
+				{
+					Signature: []byte(signature),
+					// TODO: currently we only support storing KMS keyID, may add other keys' ids later i.e. k8s secret, fulcio
```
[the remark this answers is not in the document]
wlynch:
```suggestion
					// TODO(#471): currently we only support storing KMS keyID, may add other keys' ids later i.e. k8s secret, fulcio
```

pkg/chains/storage/grafeas/grafeas.go
wlynch: Cool, let's use the original identifier then then so we're not mucking with any inputs the user gives us.

chuangw6: Done!

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -186,117 +184,187 @@ func (b *Backend) Type() string {
...
+
+// createOccurrence creates grafeas occurrences in the grafeas server that stores the original payload and its signature
+// for a single oci artifact
+//   - its simplesigning payload and signature will be stored in ATTESTATION occurrence
+//   - the identifier/ResourceUri is IMAGE_URL@IMAGE_DIGEST
+//
+// for a taskrun object
+//   - its intoto payload and signature will be stored in a BUILD occurrences for each image artifact generated from the taskrun
+//   - each BUILD occurrence will have the same data but differ in the ResourceUri field
+//   - the identifier/ResourceUri is https://IMAGE_URL@IMAGE_DIGEST
```
wlynch, line 241: This is no longer true, right?

pkg/config/config.go
```diff
@@ -153,6 +155,7 @@ const (
 	docDBUrlKey              = "storage.docdb.url"
 	grafeasProjectIDKey      = "storage.grafeas.projectid"
 	grafeasNoteIDKey         = "storage.grafeas.noteid"
+	grafeasNoteHintKey       = "storage.grafeas.notehint"
```
wlynch, line 158: This added feature seems unrelated to the broader goal of the PR. Can we split this out into another change? (this PR looks close, so feel free to base it off this branch).

config/100-rolebinding.yaml
```diff
@@ -102,6 +102,8 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: tekton-chains-controller
+  annotations:
+    iam.gke.io/gcp-service-account: chuangw@chuangw-test.iam.gserviceaccount.com
```
wlynch, line 106: remove.

wlynch (commented):
Looking good! Just a few last cleanups.

config/100-rolebinding.yaml
chuangw6: Sorry. Forgot deleting it. I changed it for testing locally 😂

examples/kaniko/taskrun.yaml
```diff
@@ -15,16 +15,17 @@
...
-  name: kaniko-run
+  generateName: kaniko-run-
 spec:
+  serviceAccountName: myksa
   taskRef:
     name: kaniko-chains
   params:
   - name: IMAGE
-    value: ${REGISTRY}/kaniko-chains
+    value: us-central1-docker.pkg.dev/chuangw-test/kaniko-example/kaniko-chains
```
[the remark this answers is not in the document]
chuangw6: done!

Issue #460, opened 2022-06-14
Add maven package support using simple type hinting in intoto provenance


ywluogg:
### Feature request

Add maven package support using simple type hinting in intoto.
Currently we support git source and OCI images. Supporting maven packages in image in simple type hinting gives extra support of such type of artifact and let users to use Chains before TEP 0075, 0076 and 0109 can be released.

### Use case
We will have:
``` yaml
      results:
        - name: mvn1_pkg_INTOTO_TARGET_NAME
           value: maven-test-0.0.1.jar
        - name: mvn1_pkg_INTOTO_TARGET_DIGEST
           value: sha256@abc
        - name: mvn1_pom_INTOTO_TARGET_NAME
           value: maven-test-0.0.1.pom
        - name: mvn1_pom_INTOTO_TARGET_DIGEST
           value: sha256@def
        - name: mvn1_src_INTOTO_TARGET_NAME
           value: maven-test-0.0.1-sources.jar
        - name: mvn1_src_INTOTO_TARGET_DIGEST
           value: sha256@xyz
```

From this, we will generate a subject in intoto provenence:
```
{"name": "maven-test-0.0.1.jar", "digest": {"sha256": "abc"}}
{"name": "maven-test-0.0.1.pom", "digest": {"sha256": "def"}}
{"name": "maven-test-0.0.1-sources.jar", "digest": {"sha256": "xyz"}}
```

Issue #276, opened 2021-11-02
Consider renaming `IMAGE_URL`/`IMAGE_DIGEST` type hints to `ARTIFACT_NAME`/`ARTIFACT_DIGEST`


priyawadhwa:
When generating provenance from a `TaskRun` [Chains looks for](https://github.com/tektoncd/chains/blob/main/docs/config.md#chains-type-hinting) `*IMAGE_URL` and `*IMAGE_DIGEST` results types to figure out what artifact was actually built. This is what goes into the [subject](https://github.com/tektoncd/chains/blob/main/PROVENANCE_SPEC.md#schema) section of the provenance. Right now, we assume OCI images were built and only have support for them.

I was thinking that we could rename these results to something more generic, like `*ARTIFACT_NAME` and `*ARTIFACT_DIGEST` so that other types of `subjects` could be included in provenance. We can still infer if something is an OCI image, but now we can also support generic files that may have been built in the TaskRun along with digests for them. This could be useful if something other than an image is being built and published from a TaskRun.

The only thing I'm not sure about is if we should also be signing these `ARTIFACT`s if they **aren't** OCI images. If a TaskRun builds and publishes a binary somewhere, should Chains be responsible for signing it somehow if it's being included in provenance?

sbose78:
This makes sense.

Have you considered where the signature would be pushed, in case of binaries?

priyawadhwa:
So if binaries are stored in one of our supported storage backends, (`gcs` or `docdb`) then we would need some way of referencing their location in the Task (maybe `gcs://mybucket/myfile` or `firestore://something` for docdb. Then, Chains could find the file, sign it and store the signature in the same spot. We could also add this to the config as a new artifact type to configure where signature are stored, so we'd add:

- `artifacts.file.storage`
- `artifacts.file.signer`

I could see this being useful for something like signing entire release yamls for Tekton/Chains releases, which are stored in GCS.

sbose78:
Makes sense.

Another comment on this:

Do you think we could make the names IMAGE_URL/IMAGE_DIGEST , ARTIFACT_NAME/ARTIFACT_DIGEST configurable for an instance of Tekton Chains installation ?

priyawadhwa:
> Do you think we could make the names IMAGE_URL/IMAGE_DIGEST , ARTIFACT_NAME/ARTIFACT_DIGEST configurable for an instance of Tekton Chains installation ?

[...]

PR #465, opened 2022-06-20
Add Intoto Signable Target support in Chains using strings in results.

ywluogg:
Currently only OCI images are supported as outputs of TaskRun in intoto provenance.
This PR adds Intoto provenance metadata as outputs with the following suffixes:
(ARTIFACT_URI, ARTIFACT_DIGEST)

Example usage for Maven packages:
``` yaml
      results:
        - name: mvn1_pkg_ARTIFACT_URI
           value: maven-test-0.0.1.jar
        - name: mvn1_pkg_ARTIFACT_DIGEST
           value: sha256@abc
        - name: mvn1_pom_ARTIFACT_URI
           value: maven-test-0.0.1.pom
        - name: mvn1_pom_ARTIFACT_DIGEST
           value: sha256@def
        - name: mvn1_src_ARTIFACT_URI
           value: maven-test-0.0.1-sources.jar
        - name: mvn1_src_ARTIFACT_DIGEST
           value: sha256@xyz
```
From this, we will generate a subject in intoto provenence:
```
{"name": "maven-test-0.0.1.jar", "digest": {"sha256": "abc"}}
{"name": "maven-test-0.0.1.pom", "digest": {"sha256": "def"}}
{"name": "maven-test-0.0.1-sources.jar", "digest": {"sha256": "xyz"}}
```

This fulfills #460


wlynch (changes requested):

pkg/artifacts/signable.go
```diff
@@ -175,6 +195,86 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
 	return objs
 }
 
+// ExtractMavenPackagesFromResults extracts Maven packages within TaskRun results and store them as StructuredSignable.
+func ExtractMavenPackagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []interface{} {
+	mavenPackages := map[string]*StructuredSignable{}
+	var objs []interface{}
+	// suffises stores the package, POM and source suffises of a Maven package.
+	// Structure is {name_of_target_suffix, digest_suffix, prefix_marker}.
```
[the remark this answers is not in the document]
ywluogg: I changed this to be a simpler format

pkg/artifacts/signable.go
```diff
@@ -175,6 +195,86 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
...
+	suffises := [][]string{{"MAVEN_PKG", "MAVEN_PKG_DIGEST", "PKG_"}, {"MAVEN_POM", "MAVEN_POM_DIGEST", "POM_"}, {"MAVEN_SRC", "MAVEN_SRC_DIGEST", "SRC_"}}
```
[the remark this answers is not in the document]
ywluogg: I changed to a more generic one

pkg/artifacts/signable.go
```diff
@@ -175,6 +195,86 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
 	return objs
 }
 
+// ExtractMavenPackagesFromResults extracts Maven packages within TaskRun results and store them as StructuredSignable.
+func ExtractMavenPackagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []interface{} {
```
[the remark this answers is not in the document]
ywluogg: I'm still keeping them in the same directory since they are more generalized to be targeting all artifacts

pkg/artifacts/signable.go
```diff
@@ -175,6 +195,86 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
...
+	for _, res := range tr.Status.TaskRunResults {
+		for _, s := range suffises {
+			nameSuffix := s[0]
+			digestSuffix := s[1]
+			markerPrefix := s[2]
+			if strings.HasSuffix(res.Name, nameSuffix) {
+				p := strings.TrimSuffix(res.Name, nameSuffix)
+				marker := markerPrefix + p
+				if v, ok := mavenPackages[marker]; ok {
+					v.Name = strings.Trim(res.Value, "\n")
```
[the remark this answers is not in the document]
ywluogg: I was copying the checks in ExtractOCIImagesFromResults(), but I don't see any Task that's providing extra \n though, but maybe it worths still putting the trim here?

pkg/artifacts/signable.go
```diff
@@ -175,6 +195,86 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
...
+				} else {
+					mavenPackages[marker] = &StructuredSignable{Name: strings.Trim(res.Value, "\n")}
+				}
+				// TODO: add logic for adding Maven package as input.
+				mavenPackages[marker].ProvType = "output"
```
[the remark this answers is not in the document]
ywluogg: Removed this field

pkg/artifacts/signable.go
```diff
@@ -175,6 +195,86 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
...
+				objs = append(objs, mvn)
+			}
+		}
+	}
+
+	return objs
+}
+
+func checkDigest(name string) error {
+	return checkElement("digest", name, digestChars, 7+64, 7+64)
```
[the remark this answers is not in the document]
ywluogg: Good call! Made it as a regex match

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -152,6 +152,22 @@ func GetSubjectDigests(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []intoto.
 		}
 	}
 
+	mvns := artifacts.ExtractMavenPackagesFromResults(tr, logger)
+	for i, obj := range mvns {
+		if m, ok := obj.(*artifacts.StructuredSignable); ok {
+			splits := strings.Split(m.Digest, ":")
+			logger.Warnf("Split %s %s", splits[0], splits[1])
+			subjects = append(subjects, intoto.Subject{
```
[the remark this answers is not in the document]
ywluogg: I'm thinking later this can also be materials

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -152,6 +152,22 @@ func GetSubjectDigests(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []intoto.
 		}
 	}
 
+	mvns := artifacts.ExtractMavenPackagesFromResults(tr, logger)
```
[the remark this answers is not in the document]
ywluogg: I'm changing this to a more general intoto provenance struct, so seems like to me this can be resolved?

pkg/artifacts/signable.go
```diff
@@ -25,6 +27,14 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+const (
+	digestFormat = `sha[0-9]+\:[a-zA-Z0-9]+`
```
wlynch, line 31: I think this will match some unintended SHAs - e.g. `sha25600000000000000...:a`. You probably want something like this instead - https://regex101.com/r/NI7n8j/2 (though it doesn't check whether the checksum algo is valid). We should add tests for this regex.

I realized after the last round that you were likely basing the old implementation off of go-containerregistry. You might be interested in https://github.com/google/go-containerregistry/pull/1395 which updates that implementation to use github.com/opencontainers/go-digest as an alternative.

pkg/artifacts/signable.go
```diff
@@ -76,6 +86,18 @@ type image struct {
 	digest string
 }
 
+// StructuredSignable contains info for signable targets to become either subjects or materials in Intoto statements.
+// Name is the identifier used within TaskRun results to identify the target.
+// Uri is the resource uri for the target needed iff the target is a material.
+// Digest is the target's SHA digest.
+// Description describes some basic info of this signable target.
+type StructuredSignable struct {
```
wlynch, line 94: This is just used internally, right? If so, we should unexport this.

pkg/artifacts/signable.go
```diff
@@ -76,6 +86,18 @@ type image struct {
...
+	Name        string
+	Description string
+	Uri         string
```
wlynch, line 97:
```suggestion
	URI         string
```

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/artifacts/signable.go
```diff
@@ -175,6 +197,61 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
 	return objs
 }
 
+// ExtractIntotoSignableTargetFromResults extracts signable targets that aim to generate Intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+func ExtractIntotoSignableTargetFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []interface{} {
+	ss := map[string]*StructuredSignable{}
+	var objs []interface{}
+	suffises := []string{"INTOTO_TARGET_NAME", "INTOTO_TARGET_DIGEST"}
```
wlynch, line 204: `INTOTO` doesn't really have anything to do with the input type, right? It's what we want the output type to be? How does this interact if users specify a different output format that isn't intoto?

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -151,6 +152,22 @@ func GetSubjectDigests(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []intoto.
 		}
 	}
 
+	mvns := artifacts.ExtractIntotoSignableTargetFromResults(tr, logger)
+	for i, obj := range mvns {
+		if m, ok := obj.(*artifacts.StructuredSignable); ok {
+			splits := strings.Split(m.Digest, ":")
+			logger.Warnf("Split %s %s", splits[0], splits[1])
```
wlynch, line 159: remove?

pkg/chains/formats/intotoite6/intotoite6.go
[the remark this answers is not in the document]
wlynch: I'd rather have separate funcs for parsing inputs from parsing outputs (even if they share a good amount of common logic), but not a blocker for this PR.

wlynch (changes requested):

pkg/artifacts/signable.go
```diff
@@ -175,6 +197,61 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
 	return objs
 }
 
+// ExtractIntotoSignableTargetFromResults extracts signable targets that aim to generate Intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+func ExtractIntotoSignableTargetFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []interface{} {
```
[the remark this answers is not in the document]
ywluogg: Done!

pkg/artifacts/signable.go
ywluogg: Done!

pkg/chains/formats/intotoite6/intotoite6.go
ywluogg: Which part are you suggesting for removal?

pkg/artifacts/signable.go
ywluogg: done!

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -39,6 +39,7 @@ const (
 	commitParam                  = "CHAINS-GIT_COMMIT"
 	urlParam                     = "CHAINS-GIT_URL"
 	ChainsReproducibleAnnotation = "chains.tekton.dev/reproducible"
+	StructuredTargetVCS          = "SIGNABLE_VCS"
```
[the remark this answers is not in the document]
ywluogg: Removed

pkg/artifacts/signable.go
ywluogg: I think if users are trying to specify different output format, maybe additional fields need to be added, and maybe this struct can be reused for that purpose. This function is only used for intoto provenance generation though

pkg/artifacts/signable.go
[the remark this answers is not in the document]
lcarva: I don't see this change in the PR.

pkg/artifacts/signable.go
```diff
@@ -175,6 +154,78 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
...
+				ss[marker] = &StructuredSignable{Digest: strings.TrimSpace(res.Value.StringVal)}
+			}
+		}
+
+	}
+
+	// Only add it if we got both the artifact name and digest.
+	// TODO: add logic for adding signable target as input.
+	for _, s := range ss {
+		if s != nil && s.Digest != "" {
```
lcarva, line 191: I'm finding the multiple nested if-else-blocks difficult to read. Consider reversing some of the checks and using `continue` to reduce nesting. Maybe something like this (untested):
```go
if s == nil || s.Digest == "" {
  continue
}
if err := checkDigest(s.Digest); err != nil {
  logger.Errorf("error getting digest %s: %v", s.Digest, err)
  continue
}

var newObj interface{}
newObj = s
if s.Name != "" && isOCIImage {
  dgst, err := name.NewDigest(fmt.Sprintf("%s@%s", s.Name, s.Digest))
  if err != nil {
    logger.Errorf("error getting digest: %v", err)
    continue
  }
  newObj = dgst
}
objs = append(objs, newObj)
```

pkg/artifacts/signable_test.go
```diff
@@ -317,11 +319,52 @@ func TestExtractOCIImagesFromResults(t *testing.T) {
 		return a.String() < b.String()
 	})
 	if !cmp.Equal(got, want, ignore...) {
-		t.Fatalf("not the same %s", cmp.Diff(got, want, ignore...))
+		t.Fatalf("not the same %s", cmp.Diff(want, got, ignore...))
 	}
 }
 
-func digest(t *testing.T, dgst string) name.Digest {
+func TestExtractIntotoSignableTargetFromResults(t *testing.T) {
```
lcarva, line 326: It would be super to add a couple of test cases to verify things are being ignored properly given the complex logic in `extractTargetFromResults`

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -151,6 +151,18 @@ func GetSubjectDigests(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []intoto.
 		}
 	}
 
+	sts := artifacts.ExtractIntotoSignableTargetFromResults(tr, logger)
+	for _, obj := range sts {
+		splits := strings.Split(obj.(*artifacts.StructuredSignable).Digest, ":")
+		logger.Warnf("Split %s %s", splits[0], splits[1])
```
lcarva, line 157: Why is this a warning?

Also, could you check the length of `splits` to ensure it is always 2? A warning if it's not 2 followed by a `continue` will handle some corner case we haven't thought of.

pkg/artifacts/signable.go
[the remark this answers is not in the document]
ywluogg: Ah I should've updated this comment - I shared the code into extractTargetsInResults, which require ExtractIntotoSignableTargetFromResults() to be exposed and needing the returned type to be []interface{}. I don't think the struct Digest is applicable for other types of artifacts, and it's more specific for image digest.

pkg/artifacts/signable_test.go
ywluogg: Ah yeah they are covered in TestExtractIntotoSignableTargetFromResults cases (miss_target_name_INTOTO_TARGET_DIGEST, wrong_digest_format_INTOTO_TARGET_NAME, etc.). I've added a few more in TestExtractOCIImagesFromResults as well

pkg/artifacts/signable.go
ywluogg: Sure!

pkg/artifacts/signable.go
```diff
@@ -175,6 +154,78 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
...
+		if s.Name != "" {
+			if isOCIImage {
+				dgst, err := name.NewDigest(fmt.Sprintf("%s@%s", s.Name, s.Digest))
+				if err != nil {
+					logger.Errorf("error getting digest: %v", err)
+					continue
+				}
+				objToAppend = dgst
+			}
+			objs = append(objs, objToAppend)
```
lcarva, line 210: Oh, if we're only appending if `s.Name != ""`, should the beginning of the for-loop also `continue` if `s.Name == ""` ?

It would also skip the `checkDigest` call, not sure if that would be intended.

pkg/artifacts/signable.go
```diff
@@ -175,6 +154,78 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
...
+
+func extractTargetFromResults(tr *v1beta1.TaskRun, identifierSuffix string, digestSuffix string, logger *zap.SugaredLogger, isOCIImage bool) []interface{} {
+	ss := map[string]*StructuredSignable{}
+	var objs []interface{}
+
+	for _, res := range tr.Status.TaskRunResults {
+		if strings.HasSuffix(res.Name, identifierSuffix) {
+			marker := strings.TrimSuffix(res.Name, identifierSuffix)
+			if v, ok := ss[marker]; ok {
+				v.Name = strings.TrimSpace(res.Value.StringVal)
```
wlynch, line 170: I think this should be URI?

pkg/artifacts/signable.go
```diff
@@ -175,6 +154,78 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
 	return objs
 }
 
+// ExtractIntotoSignableTargetFromResults extracts signable targets that aim to generate intoto provenance as materials within TaskRun results and store them as structuredSignable.
+func ExtractIntotoSignableTargetFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []interface{} {
+	return extractTargetFromResults(tr, "INTOTO_TARGET_NAME", "INTOTO_TARGET_DIGEST", logger, false)
+}
+
+func extractTargetFromResults(tr *v1beta1.TaskRun, identifierSuffix string, digestSuffix string, logger *zap.SugaredLogger, isOCIImage bool) []interface{} {
```
wlynch, line 162: Having to pass in a `isOCIImage` bool is probably a sign that we should split the 2nd half of this up. What we don't want to happen is having to add a bunch of bools to control behavior for different types.

If OCI image handling wants to do additional transformations later, that's fine.

pkg/artifacts/signable.go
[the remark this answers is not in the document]
wlynch: I'm not a big fan of having a func that can return different types - clients need to type switch in order to know what happened. I'd return the same type here. If we need to do additional type specific transformations we can do that in their respective calling funcs.

pkg/artifacts/signable.go
```diff
@@ -175,6 +154,78 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
 	return objs
 }
 
+// ExtractIntotoSignableTargetFromResults extracts signable targets that aim to generate intoto provenance as materials within TaskRun results and store them as structuredSignable.
+func ExtractIntotoSignableTargetFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []interface{} {
```
wlynch, line 158: unexport

pkg/artifacts/signable.go
wlynch: Don't think this change made it to the branch

pkg/chains/formats/intotoite6/intotoite6.go
wlynch: the log - this was taken care of!

pkg/artifacts/signable.go
wlynch: This is just a URI + digest though, right? I'm a bit hesitant to tie the naming of this directly to intoto, since it's not really intoto specific - intoto is just a way to format artifact data.

This also makes it easier to support other formats in the future (i.e. if slsa and intoto diverge).

pkg/artifacts/signable.go
ywluogg: I think I wanted to use this as artifacts. ExtractIntotoSignableTargetFromResults in intotoite6.go

pkg/artifacts/signable.go
ywluogg: I think for other signables, they are not essentially URI though

pkg/artifacts/signable.go
ywluogg: I think this is intended, based on the old behavior: https://github.com/tektoncd/chains/blob/b1c8b5aba749f98a82f91b8dd903967d4d42ccc1/pkg/artifacts/signable.go#L144

wlynch (commented):
Looking good! Just some final bikeshedding around naming.

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -151,6 +151,21 @@ func GetSubjectDigests(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []intoto.
...
 	}
 
+	sts := artifacts.ExtractSignableTargetFromResults(tr, logger)
+	for _, obj := range sts {
+		splits := strings.Split(obj.Digest, ":")
+		if len(splits) != 2 {
+			logger.Errorf("Digest %s should be in the format of: algorthm:abc", obj.Digest)
+			continue
+		}
+		subjects = append(subjects, intoto.Subject{
```
priyawadhwa, line 161: Do we confirm that these artifacts exist before adding them as subjects anywhere? I'm asking because I'm wondering if you could configure a TaskRun's results to say something was created when it wasn't, and then add it as a Subject to the attestation.

Right now we only support OCI images and if the image doesn't exist then we can't store the attestation alongside it, so it kind of gets around that.

pkg/artifacts/signable.go
```diff
@@ -175,6 +167,65 @@ func ExtractOCIImagesFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger)
 	return objs
 }
 
+// ExtractSignableTargetFromResults extracts signable targets that aim to generate intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+func ExtractSignableTargetFromResults(tr *v1beta1.TaskRun, logger *zap.SugaredLogger) []*StructuredSignable {
+	objs := []*StructuredSignable{}
+	ss := extractTargetFromResults(tr, "CHAINS_SIGNABLE_NAME", "CHAINS_SIGNABLE_DIGEST", logger)
```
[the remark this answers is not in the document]
priyawadhwa: I think generalizing to `ARTIFACT_URL` / `ARTIFACT_DIGEST` makes sense! I actually had a similar idea for replacing `IMAGE_URL` AND `IMAGE_DIGEST` with those exact definitions a while ago (https://github.com/tektoncd/chains/issues/276) 😄

pkg/chains/formats/intotoite6/intotoite6.go
wlynch: Even if we did verify existence, a malicious task could lie and point to an existing version. 🤔 

We'd probably need to couple this with Spire + Trusted Resources in order to trust the value.

priyawadhwa: SPIRE would guarantee the result hasn't been modified, but not that the artifact actually exists. I don't know much about Trusted Resources, would adding that in provide that guarantee? If not, I think Chains is the only spot we can make sure the artifact exists (agree it's not perfect, but it does come with some nice guarantees). 

At the moment, the Subjects themselves are also signed before the signed attestation is generated and attached because they are an artifact Chains recognizes (`oci`). The new Subjects added here wouldn't be signed, and we wouldn't know if they exist or not. I wonder if something like what I described in https://github.com/tektoncd/chains/issues/276#issuecomment-959295196 might be the move. Then for an arbitrary file we could sign it, create the attestation and store that somewhere.

pkg/artifacts/signable.go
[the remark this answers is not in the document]
ywluogg: Changed to `ARTIFACT_URL ` / `ARTIFACT_DIGEST `!

pkg/artifacts/signable.go
```diff
@@ -76,6 +78,16 @@ type image struct {
 	digest string
 }
 
+// StructuredSignable contains info for signable targets to become either subjects or materials in intoto Statements.
+// Name is the identifier that uniquely identifies the signable. It can be the same as URI.
+// URI is the resource uri for the target needed iff the target is a material.
+// Digest is the target's SHA digest.
+type StructuredSignable struct {
+	Name   string
```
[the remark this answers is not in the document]
ywluogg: Removed

wlynch:
/lgtm

holding off on approval for any additional feedback from @priyawadhwa

wlynch (approved):

PR #491, opened 2022-07-15
[TEP 109] Add feature to extract structured signable targets and store them in subjects and materials in intoto provenance

ywluogg:
Implementation of TEP 109.
Example usage for Maven packages:
``` yaml
results:
        - name: img_1-ARTIFACT_INPUTS
           value:
               uri: gcr.io/foo/bar
               digest: sha123@qwe
        - name: mvn1_pkg-ARTIFACT_OUTPUTS
           value:
               uri: maven-test-0.0.1.jar
               digest: sha256@abc
        - name: mvn1_pom-ARTIFACT_OUTPUTS
           value:
               uri: maven-test-0.0.1.pom
               digest: sha256@def
        - name: mvn1_src-ARTIFACT_OUTPUTS
           value:
               uri: maven-test-0.0.1-sources.jar
               digest: sha256@xyz
```

From this, we will generate `subjects` in intoto provenence:
```
{"name": "maven-test-0.0.1.jar", "digest": {"sha256": "abc"}}
{"name": "maven-test-0.0.1.pom", "digest": {"sha256": "def"}}
{"name": "maven-test-0.0.1-sources.jar", "digest": {"sha256": "xyz"}}
```
In `materials`:
```
{"uri": "gcr.io/foo/bar", "digest": {"sha123": "qwe"}}
```

The feature requires Pipeline v0.38 or later.

# Submitter Checklist
* [x]  Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
* [x]  Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
* [x]  Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
* [x]  Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
* [x]  Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
* [ ]  Release notes contains the string "action required" if the change requires additional action from users switching to the new release
# Release Notes
```
Added structured results provenance retrieval for both TaskRun and PipelineRun. (TEP 109 implementation)
```

### DIFF
--- a/pkg/artifacts/signable.go
+++ b/pkg/artifacts/signable.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
+	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
 	"github.com/opencontainers/go-digest"
 	"github.com/tektoncd/chains/pkg/chains/formats"
 	"github.com/tektoncd/chains/pkg/chains/objects"
@@ -26,6 +27,11 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	ArtifactsInputsResultName  = "ARTIFACT_INPUTS"
+	ArtifactsOutputsResultName = "ARTIFACT_OUTPUTS"
 )
 
 type Signable interface {
@@ -255,6 +261,88 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
 
 	}
 	return ss
+}
+
+// RetrieveMaterialsFromStructuredResults retrieves structured results from Tekton Object, and convert them into materials.
+func RetrieveMaterialsFromStructuredResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []slsa.ProvenanceMaterial {
+	// Retrieve structured provenance for inputs.
+	mats := []slsa.ProvenanceMaterial{}
+	ssts := ExtractStructuredTargetFromResults(obj, ArtifactsInputsResultName, logger)
+	for _, s := range ssts {
+		if err := checkDigest(s.Digest); err != nil {
+			logger.Debugf("Digest for %s not in the right format: %s, %v", s.URI, s.Digest, err)
+			continue
+		}
+		splits := strings.Split(s.Digest, ":")
+		alg := splits[0]
+		digest := splits[1]
+		mats = append(mats, slsa.ProvenanceMaterial{
+			URI:    s.URI,
+			Digest: map[string]string{alg: digest},
+		})
+	}
+	return mats
+}
+
+// ExtractStructuredTargetFromResults extracts structured signable targets aim to generate intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+// categoryMarker categorizes signable targets into inputs and outputs.
+func ExtractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []*StructuredSignable {
+	objs := []*StructuredSignable{}
+	if categoryMarker != ArtifactsInputsResultName && categoryMarker != ArtifactsOutputsResultName {
+		return objs
+	}
+
+	// TODO(#592): support structured results using Run
+	results := []objects.Result{}
+	tr, isTaskRun := obj.GetObject().(*v1beta1.TaskRun)
+	if isTaskRun {
+		for _, res := range tr.Status.TaskRunResults {
+			results = append(results, objects.Result{
+				Name:  res.Name,
+				Value: res.Value,
+			})
+		}
+
+	} else {
+		pr := obj.GetObject().(*v1beta1.PipelineRun)
+		for _, res := range pr.Status.PipelineResults {
+			results = append(results, objects.Result{
+				Name:  res.Name,
+				Value: res.Value,
+			})
+		}
+	}
+	for _, res := range results {
+		if strings.HasSuffix(res.Name, categoryMarker) {
+			valid, err := isStructuredResult(res, categoryMarker)
+			if err != nil {
+				logger.Debugf("ExtractStructuredTargetFromResults: %v", err)
+			}
+			if valid {
+				objs = append(objs, &StructuredSignable{URI: res.Value.ObjectVal["uri"], Digest: res.Value.ObjectVal["digest"]})
+			}
+		}
+	}
+	return objs
+}
+
+func isStructuredResult(res objects.Result, categoryMarker string) (bool, error) {
+	if !strings.HasSuffix(res.Name, categoryMarker) {
+		return false, nil
+	}
+	if res.Value.ObjectVal == nil {
+		return false, fmt.Errorf("%s should be an object: %v", res.Name, res.Value.ObjectVal)
+	}
+	if res.Value.ObjectVal["uri"] == "" {
+		return false, fmt.Errorf("%s should have uri field: %v", res.Name, res.Value.ObjectVal)
+	}
+	if res.Value.ObjectVal["digest"] == "" {
+		return false, fmt.Errorf("%s should have digest field: %v", res.Name, res.Value.ObjectVal)
+	}
+	if err := checkDigest(res.Value.ObjectVal["digest"]); err != nil {
+		return false, fmt.Errorf("error getting digest %s: %v", res.Value.ObjectVal["digest"], err)
+	}
+	return true, nil
 }
 
 func checkDigest(dig string) error {
--- a/pkg/artifacts/signable_test.go
+++ b/pkg/artifacts/signable_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/go-containerregistry/pkg/name"
+	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
 	"github.com/tektoncd/chains/pkg/chains/objects"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -378,6 +379,208 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
 	})
 	if !cmp.Equal(got, want, ignore...) {
 		t.Fatalf("not the same %s", cmp.Diff(want, got, ignore...))
+	}
+}
+
+func TestExtractStructuredTargetFromResults(t *testing.T) {
+	tr := &v1beta1.TaskRun{
+		Status: v1beta1.TaskRunStatus{
+			TaskRunStatusFields: v1beta1.TaskRunStatusFields{
+				TaskRunResults: []v1beta1.TaskRunResult{
+					{
+						Name: "mvn1_pkg" + "_" + ArtifactsOutputsResultName,
+						Value: *v1beta1.NewObject(map[string]string{
+							"uri":           "projects/test-project/locations/us-west4/repositories/test-repo/mavenArtifacts/com.google.guava:guava:31.0-jre",
+							"digest":        digest1,
+							"signable_type": "",
+						}),
+					},
+					{
+						Name: "mvn1_pom_sha512" + "_" + ArtifactsOutputsResultName,
+						Value: *v1beta1.NewObject(map[string]string{
+							"uri":           "com.google.guava:guava:31.0-jre.pom",
+							"digest":        digest2,
+							"signable_type": "",
+						}),
+					},
+					{
+						Name: "img1_input" + "_" + ArtifactsInputsResultName,
+						Value: *v1beta1.NewObject(map[string]string{
+							"uri":    "gcr.io/foo/bar",
+							"digest": digest3,
+						}),
+					},
+					{
+						Name: "img2_input_no_digest" + "_" + ArtifactsInputsResultName,
+						Value: *v1beta1.NewObject(map[string]string{
+							"uri":    "gcr.io/foo/foo",
+							"digest": "",
+						}),
+					},
+				},
+			},
+		},
+	}
+
+	wantInputs := []*StructuredSignable{
+		{URI: "gcr.io/foo/bar", Digest: digest3},
+	}
+	gotInputs := ExtractStructuredTargetFromResults(objects.NewTaskRunObject(tr), ArtifactsInputsResultName, logtesting.TestLogger(t))
+	if diff := cmp.Diff(gotInputs, wantInputs, cmpopts.SortSlices(func(x, y *StructuredSignable) bool { return x.Digest < y.Digest })); diff != "" {
+		t.Errorf("Inputs are not as expected: %v", diff)
+	}
+
+	wantOutputs := []*StructuredSignable{
+		{URI: "projects/test-project/locations/us-west4/repositories/test-repo/mavenArtifacts/com.google.guava:guava:31.0-jre", Digest: digest1},
+		{URI: "com.google.guava:guava:31.0-jre.pom", Digest: digest2},
+	}
+	gotOutputs := ExtractStructuredTargetFromResults(objects.NewTaskRunObject(tr), ArtifactsOutputsResultName, logtesting.TestLogger(t))
+	opts := append(ignore, cmpopts.SortSlices(func(x, y *StructuredSignable) bool { return x.Digest < y.Digest }))
+	if diff := cmp.Diff(gotOutputs, wantOutputs, opts...); diff != "" {
+		t.Error(diff)
+	}
+}
+
+func TestRetrieveMaterialsFromStructuredResults(t *testing.T) {
+	tr := &v1beta1.TaskRun{
+		Status: v1beta1.TaskRunStatus{
+			TaskRunStatusFields: v1beta1.TaskRunStatusFields{
+				TaskRunResults: []v1beta1.TaskRunResult{
+					{
+						Name: "img1_input" + "_" + ArtifactsInputsResultName,
+						Value: *v1beta1.NewObject(map[string]string{
+							"uri":    "gcr.io/foo/bar",
+							"digest": "sha256:05f95b26ed10668b7183c1e2da98610e91372fa9f510046d4ce5812addad86b7",
+						}),
+					},
+					{
+						Name: "img2_input_no_digest" + "_" + ArtifactsInputsResultName,
+						Value: *v1beta1.NewObject(map[string]string{
+							"uri":    "gcr.io/foo/foo",
+							"digest": "",
+						}),
+					},
+					{
+						Name: "img2_input_invalid_digest" + "_" + ArtifactsInputsResultName,
+						Value: *v1beta1.NewObject(map[string]string{
+							"uri":    "gcr.io/foo/foo",
+							"digest": "sha:123",
+						}),
+					},
+				},
+			},
+		},
+	}
+	wantMaterials := []slsa.ProvenanceMaterial{
+		{
+			URI:    "gcr.io/foo/bar",
+			Digest: map[string]string{"sha256": "05f95b26ed10668b7183c1e2da98610e91372fa9f510046d4ce5812addad86b7"},
+		},
+	}
+
+	gotMaterials := RetrieveMaterialsFromStructuredResults(objects.NewTaskRunObject(tr), ArtifactsInputsResultName, logtesting.TestLogger(t))
+
+	if diff := cmp.Diff(gotMaterials, wantMaterials, ignore...); diff != "" {
+		t.Fatalf("Materials not the same %s", diff)
+	}
+}
+
+func TestValidateResults(t *testing.T) {
+	tests := []struct {
+		name           string
+		obj            objects.Result
+		categoryMarker string
+		wantResult     bool
+		wantErr        error
+	}{
+		{
+			name:           "valid result",
+			categoryMarker: ArtifactsOutputsResultName,
+			obj: objects.Result{
+				Name: "valid_result-ARTIFACT_OUTPUTS",
+				Value: v1beta1.ParamValue{
+					ObjectVal: map[string]string{
+						"uri":    "gcr.io/foo/bar",
+						"digest": digest3,
+					},
+				},
+			},
+			wantResult: true,
+			wantErr:    nil,
+		},
+		{
+			name:           "invalid result without digest field",
+			categoryMarker: ArtifactsOutputsResultName,
+			obj: objects.Result{
+				Name: "missing_digest-ARTIFACT_OUTPUTS",
+				Value: v1beta1.ParamValue{
+					ObjectVal: map[string]string{
+						"uri": "gcr.io/foo/bar",
+					},
+				},
+			},
+			wantResult: false,
+			wantErr:    fmt.Errorf("missing_digest-ARTIFACT_OUTPUTS should have digest field: map[uri:gcr.io/foo/bar]"),
+		},
+		{
+			name:           "invalid result without uri field",
+			categoryMarker: ArtifactsOutputsResultName,
+			obj: objects.Result{
+				Name: "missing_digest-ARTIFACT_OUTPUTS",
+				Value: v1beta1.ParamValue{
+					ObjectVal: map[string]string{
+						"digest": digest3,
+					},
+				},
+			},
+			wantResult: false,
+			wantErr:    fmt.Errorf("missing_digest-ARTIFACT_OUTPUTS should have uri field: map[digest:sha256:05f95b26ed10668b7183c1e2da98610e91372fa9f510046d4ce5812addad86b7]"),
+		},
+		{
+			name:           "invalid result wrong digest format",
+			categoryMarker: ArtifactsOutputsResultName,
+			obj: objects.Result{
+				Name: "missing_digest-ARTIFACT_OUTPUTS",
+				Value: v1beta1.ParamValue{
+					ObjectVal: map[string]string{
+						"uri":    "gcr.io/foo/bar",
+						"digest": "",
+					},
+				},
+			},
+			wantResult: false,
+			wantErr:    fmt.Errorf("missing_digest-ARTIFACT_OUTPUTS should have digest field: map[digest: uri:gcr.io/foo/bar]"),
+		},
+		{
+			name:           "invalid result wrong type hinting",
+			categoryMarker: ArtifactsOutputsResultName,
+			obj: objects.Result{
+				Name: "missing_digest-ARTIFACTs_OUTPUTS",
+				Value: v1beta1.ParamValue{
+					ObjectVal: map[string]string{
+						"uri":    "gcr.io/foo/bar",
+						"digest": digest3,
+					},
+				},
+			},
+			wantResult: false,
+			wantErr:    nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := isStructuredResult(tt.obj, tt.categoryMarker)
+			if got != tt.wantResult {
+				t.Errorf("Validation result is not as the expected: got %v and wanted %v", got, tt.wantResult)
+			}
+			if !tt.wantResult && tt.wantErr != nil {
+				if diff := cmp.Diff(err.Error(), tt.wantErr.Error()); diff != "" {
+					t.Errorf("Validation error is not as the expected: %s", diff)
+				}
+			}
+
+		})
 	}
 }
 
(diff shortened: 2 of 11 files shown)


ywluogg:
@wlynch Hi Billy, do you mind reviewing this PR? Thanks!

pkg/artifacts/signable.go
```diff
@@ -257,6 +276,59 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+				continue
+			}
+			if err := checkDigest(res.Value.ObjectVal["digest"]); err != nil {
+				logger.Errorf("error getting digest %s: %v", res.Value.ObjectVal["digest"], err)
+				continue
+			}
+			if res.Value.ObjectVal["uri"] == "" {
+				logger.Errorf("URI cannot be empty for %s", res.Name)
+				continue
+			}
```
chuangw6, line 295: Similar to checking if `URI` is empty, should we check if `digest` field exists / is empty instead of relying on `checkDigest` to check the format only?

pkg/artifacts/signable.go
```diff
@@ -38,6 +45,18 @@ type Signable interface {
 	Enabled(cfg config.Config) bool
 }
 
+type ProvenanceSchema struct {
+	// Properties is the JSON Schema properties to support key-value pairs parameter.
+	Properties map[string]v1beta1.PropertySpec
+}
```
chuangw6, line 51: This struct sounds more like a general schema for object type instead of something specific to provenance. Maybe rename and/or add some comments?

pkg/chains/formats/intotoite6/taskrun/taskrun.go
```diff
@@ -92,6 +94,18 @@ func materials(tro *objects.TaskRunObject) []slsa.ProvenanceMaterial {
...
+	for _, s := range ssts {
+		splits := strings.Split(s.Digest, ":")
+		alg := splits[0]
+		digest := splits[1]
+		mats = append(mats, slsa.ProvenanceMaterial{
+			URI:    s.URI,
+			Digest: map[string]string{alg: digest},
+		})
+	}
+
```
chuangw6, line 108: We also need to do this for [pipelinerun level material extraction](https://github.com/tektoncd/chains/blob/e0f5af8998ce80a477af7fd1fc25fed9147f5fc7/pkg/chains/formats/intotoite6/pipelinerun/pipelinerun.go#L171-L217).

chuangw6 (commented):
Thanks @ywluogg!

pkg/artifacts/signable.go
ywluogg: Thanks for the feedback! Are you trying to say that this is a more generic struct and probably should be named as something more generic as well? I think this struct is not used anywhere else, and for now it's only used for provenance, and I don't see a trend that this struct would be used under other circumstances. So it seems reasonable to keep the naming still under the scope of provenance.

pkg/artifacts/signable.go
ywluogg: I think that check is done in O(1) time at `!strings.HasPrefix(dig, prefix)` in `checkDigest`, so probably no?

chuangw6: Synched offline with @ywluogg. `checkDigest` helps check the format and emptiness for the key `digest` at the same time. Sounds good.

pkg/artifacts/signable.go
chuangw6: Synced offline with @ywluogg. My concern was that the name `ProvenanceSchema` might sound like we define a new provenance schema alternative to intoto/slsa. We can come back to this later if others have better suggestions on the name.

ywluogg: +1 totally. I'm happy to grab more feedback along the way for better naming here. One good thing is that this is only used in limited scope and doesn't affect external integrations. I will make the struct not exposed

pkg/artifacts/signable.go
```diff
@@ -257,6 +276,59 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
 	return ss
 }
 
+// ExtractStructuredTargetFromResults extracts structured signable targets aim to generate intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+func ExtractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []*StructuredSignable {
+	objs := []*StructuredSignable{}
+	for _, res := range obj.GetResults() {
+		if strings.HasSuffix(res.Name, categoryMarker) {
+			if validateObjectKeys(structuredSignableSchema.Properties, &res.Value) != nil {
```
wlynch, line 284: nit: break the individual result parsing out into its own func so that it's easier to test.

wlynch, line 284: Do we need the `validateObjectKeys` func if we're effectively re-validating the object fields when we try to extract the object values below?

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +381,72 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+	}
+	wantInputs := []*StructuredSignable{
+		{URI: "gcr.io/foo/bar", Digest: digest3},
+	}
+
+	wantOutputs := []*StructuredSignable{
+		{URI: "projects/test-project/locations/us-west4/repositories/test-repo/mavenArtifacts/com.google.guava:guava:31.0-jre", Digest: digest1},
+		{URI: "com.google.guava:guava:31.0-jre.pom", Digest: digest2},
+	}
+	gotInputs := ExtractStructuredTargetFromResults(objects.NewTaskRunObject(tr), ArtifactsInputsResultName, logtesting.TestLogger(t))
```
wlynch, line 431: nit: break this out so that inputs are grouped with inputs, outputs are grouped with outputs. You can even embed this with a t.Run to make this a little cleaner:

```go
t.Run("inputs", func(t *testing.T) {
  got := ...
  if diff := ... {
    t.Fatal()
  }
})

t.Run("outputs", func(t *testing.T) {
  got := ...
  if diff := ... {
    t.Fatal()
  }
})
```

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +381,72 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+	gotOutputs := ExtractStructuredTargetFromResults(objects.NewTaskRunObject(tr), ArtifactsOutputsResultName, logtesting.TestLogger(t))
+	gotInputs = sortArtifacts(gotInputs)
```
wlynch, line 433: Pass a https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortSlices option to the comparison below - this will ensure that both inputs and outputs are sorted.

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +381,72 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+	gotOutputs = sortArtifacts(gotOutputs)
+	if !cmp.Equal(gotInputs, wantInputs, ignore...) {
+		t.Fatalf("Inputs not the same %s", cmp.Diff(wantInputs, gotInputs, ignore...))
+	}
```
wlynch, line 437:
```suggestion
	if diff := cmp.Diff(got, want); diff != "" {
		t.Fatalf(diff)
	}
```

pkg/artifacts/signable.go
```diff
@@ -257,6 +276,59 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
 	return ss
 }
 
+// ExtractStructuredTargetFromResults extracts structured signable targets aim to generate intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+func ExtractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []*StructuredSignable {
```
wlynch, line 280: it wasn't obvious to me what categoryMarker meant just by reading the header + godoc. Perhaps rename (nameSuffix?) and/or add more godoc to explain?

pkg/artifacts/signable.go
```diff
@@ -257,6 +276,59 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+	objs := []*StructuredSignable{}
+	for _, res := range obj.GetResults() {
+		if strings.HasSuffix(res.Name, categoryMarker) {
+			if validateObjectKeys(structuredSignableSchema.Properties, &res.Value) != nil {
+				logger.Errorf("%s should follow this schema: %v", res.Name, structuredSignableSchema)
```
wlynch, line 285: Drop this and other log message here to warn - this is a problem with what the user gave us, not with how chains is processing this.

pkg/chains/formats/intotoite6/extract/extract.go
```diff
@@ -62,6 +62,19 @@ func SubjectDigests(obj objects.TektonObject, logger *zap.SugaredLogger) []intot
 		})
 	}
 
+	ssts := artifacts.ExtractStructuredTargetFromResults(obj, artifacts.ArtifactsOutputsResultName, logger)
+	for _, s := range ssts {
+		splits := strings.Split(s.Digest, ":")
+		alg := splits[0]
+		digest := splits[1]
```
wlynch, line 69: This can panic if digest isn't what we expect. (we should add tests for this )

wlynch (commented):
Good start!

chuangw6:
> We might also need to change the function retrieveAllArtifactIdentifiers in https://github.com/tektoncd/chains/blob/main/pkg/chains/storage/grafeas/grafeas.go to get artifacts from StructuredTargets. Also please check if changes are needed for other storage as well.

This is a great catch! Thanks @jagathprakash .
I just changed that function in grafeas to [get URIs directly from intoto subjects](https://github.com/tektoncd/chains/pull/590/files#diff-18bb01f60078ffb3318ebfab5c57b00d5d38b8e55930d07e5c19cc88cb233516R456-R459) instead of calling individual helpers. In future, we don't need to worry the changes made to the Subject function in other places then.
So let's keep this out of scope of this PR for now.

jagathprakash:
> > We might also need to change the function retrieveAllArtifactIdentifiers in https://github.com/tektoncd/chains/blob/main/pkg/chains/storage/grafeas/grafeas.go to get artifacts from StructuredTargets. Also please check if changes are needed for other storage as well.
> 
> This is a great catch! Thanks @jagathprakash . I just changed that function in grafeas to [get URIs directly from intoto subjects](https://github.com/tektoncd/chains/pull/590/files#diff-18bb01f60078ffb3318ebfab5c57b00d5d38b8e55930d07e5c19cc88cb233516R456-R459) instead of calling individual helpers. In future, we don't need to worry the changes made to the Subject function in other places then. So let's keep this out of scope of this PR for now.

Thanks @chuangw6 for fixing this.

pkg/artifacts/signable.go
ywluogg: Can you explain a bit more on this? I'm trying to understand that maybe we are having this strategy that when users are not passing the right schema, Chains should just ignore it. But when it has the type hinting marker, should we expect the schema should be right and error / warn about it?

pkg/artifacts/signable.go
ywluogg: SG!

ywluogg: Ah yeah I'm not making it as a separate function anymore.

pkg/artifacts/signable.go
ywluogg: I've added godoc since I'm trying to make it clear that the type hinting is also distinguishing artifacts into inputs and outputs

pkg/chains/formats/intotoite6/taskrun/taskrun.go
ywluogg: Good catch thanks! Added here. The tests need to be added later as there is an update needed from Tekton Pipeline side for json marshall for object results. I've already talked to Yongxuan about it. Added a TODO for that as well

pkg/artifacts/signable_test.go
ywluogg: Ah so I added a TestValidateResults for testing them separately. In here, I'm trying to additionally test them in a single TaskRun to make sure they can be picked up into separate groups

pkg/artifacts/signable_test.go
ywluogg: Added

pkg/chains/formats/intotoite6/extract/extract.go
ywluogg: Yeah we should have separate tests for `SubjectDigests()`, but this particular piece's tested in taskrun/provenance_test.go. I've added a TODO for SubjectDigests() to add more tests in a separate PR

pkg/artifacts/signable.go
```diff
@@ -257,6 +268,104 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+}
+
+// MergeStructuredesultsIntoMaterials merge structure.
+func MergeStructuredResultsIntoMaterials(mats *[]slsa.ProvenanceMaterial, obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) {
+	// Retrieve structured provenance for inputs.
+	ssts := extractStructuredTargetFromResults(obj, ArtifactsInputsResultName, logger)
+	for _, s := range ssts {
+		splits := strings.Split(s.Digest, ":")
+		alg := splits[0]
+		digest := splits[1]
```
wlynch, line 306: This can panic if digest isn't properly formatted (we should add a test for this).

pkg/artifacts/signable.go
```diff
@@ -257,6 +268,104 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+		*mats = append(*mats, slsa.ProvenanceMaterial{
```
wlynch, line 307: I'm usually not a big fan of doing pointer mutation. Can we just return the materials list and let the caller do the append themselves?

wlynch (changes requested):

pkg/artifacts/signable.go
```diff
@@ -257,6 +268,104 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+}
+
+// extractStructuredTargetFromResults extracts structured signable targets aim to generate intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+// categoryMarker categorizes signable targets into inputs and outputs.
+func extractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []*StructuredSignable {
+	objs := []*StructuredSignable{}
+	// TODO(#592): support structured results using Run
+	results := []objects.Result{}
+	tr, isTaskRun := obj.GetObject().(*v1beta1.TaskRun)
+	if isTaskRun {
```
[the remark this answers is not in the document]
ywluogg: Or are you referring to something else here?

pkg/artifacts/signable.go
ywluogg: Good point! Added a test

pkg/artifacts/signable.go
ywluogg: SG! I just return a list and let caller to merge them 🍡

pkg/artifacts/signable.go
```diff
@@ -257,6 +268,104 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+				logger.Errorf("URI cannot be empty for %s", res.Name)
+				continue
+			}
+			objs = append(objs, &StructuredSignable{URI: res.Value.ObjectVal["uri"], Digest: res.Value.ObjectVal["digest"]})
+		}
+	}
+	return objs
+}
+
+// MergeStructuredesultsIntoMaterials merge structure.
```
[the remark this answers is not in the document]
ywluogg: Done!

pkg/artifacts/signable.go
```diff
@@ -257,6 +268,104 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+		*mats = append(*mats, slsa.ProvenanceMaterial{
+			URI:    s.URI,
+			Digest: map[string]string{alg: digest},
+		})
+	}
+}
+
+// extractStructuredTargetFromResults extracts structured signable targets aim to generate intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+// categoryMarker categorizes signable targets into inputs and outputs.
+func extractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []*StructuredSignable {
```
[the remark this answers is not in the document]
ywluogg: merged them together

pkg/artifacts/signable.go
```diff
@@ -257,6 +268,104 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+			}
+			if valid {
+				objs = append(objs, &StructuredSignable{URI: res.Value.ObjectVal["uri"], Digest: res.Value.ObjectVal["digest"]})
+			}
+		}
+	}
+	return objs
+}
+
+func validateResult(res objects.Result, categoryMarker string) (bool, error) {
```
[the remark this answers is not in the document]
ywluogg: Changed to isStructuredResult~

pkg/artifacts/signable.go
```diff
@@ -257,6 +268,104 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
 	return ss
 }
 
+// ExtractStructuredTargetFromResults extracts structured signable targets aim to generate intoto provenance as materials within TaskRun results and store them as StructuredSignable.
+// categoryMarker can either be ARTIFACT_INPUTS or ARTIFACT_OUTPUTS.
+func ExtractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []*StructuredSignable {
```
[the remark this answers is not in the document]
ywluogg: I'm curious what's the benefits of directly using `logging.FromContext`? I'm seeing that the places that are using `ExtractStructuredTargetFromResults` are already using logger directly, so before I go ahead and change those existing functions' arguments, I wanted to confirm that's what we plan to do

pkg/artifacts/signable.go
```diff
@@ -257,6 +268,104 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+	objs := []*StructuredSignable{}
+	for _, res := range obj.GetResults() {
+		if strings.HasSuffix(res.Name, categoryMarker) {
+			if res.Value.ObjectVal["uri"] == "" {
+				logger.Errorf("%s should have uri key", res.Name)
```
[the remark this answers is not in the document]
ywluogg: Done!

pkg/artifacts/signable.go
ywluogg: Good point! Added a check there

pkg/artifacts/signable.go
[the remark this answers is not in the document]
ywluogg: Marking this as resolved. Feel free to re-open this if you see anything else needs to be addressed here

pkg/artifacts/signable_test.go
ywluogg: SG!

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +381,171 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+	})
+	sort.Slice(gotOutputs, func(i, j int) bool {
+		return gotOutputs[i].Digest < gotOutputs[j].Digest
+	})
+	sort.Slice(wantInputs, func(i, j int) bool {
+		return wantInputs[i].Digest < wantInputs[j].Digest
+	})
+	sort.Slice(wantOutputs, func(i, j int) bool {
+		return wantOutputs[i].Digest < wantOutputs[j].Digest
+	})
```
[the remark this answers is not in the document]
ywluogg: Done!

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +381,171 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := validateResult(tt.obj, tt.categoryMarker)
+			if !cmp.Equal(got, tt.wantResult, ignore...) {
+				t.Errorf("Validation result is not as the expected: %s", cmp.Diff(got, tt.wantResult))
+			}
+			if !cmp.Equal(got, tt.wantResult, ignore...) {
+				t.Errorf("Validation error is not as the expected: %s", cmp.Diff(err, tt.wantErr))
+			}
```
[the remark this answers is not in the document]
ywluogg: Done!

pkg/chains/formats/intotoite6/extract/extract.go
ywluogg: The test is present for TaskRun. I added one for PipelineRun

ywluogg:
@wlynch I'm curious if I should add a user doc for this midland solution? I talked to the teams that are interested in adopting to this method, their main concern for adopting to the `artifacts` field in `status` is that they need to change a lot in their pipeline, and this current midland solution can help them migrating to a structured result, but the integration effort is not large comparing to the future solution that Prakash is working on.

I can add a user doc, but have a warning that we are doing a new design and implementation

pkg/artifacts/signable.go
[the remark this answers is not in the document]
wlynch: Hmmm that looks like an oversight (especially since [TaskRuns seem to support it](https://github.com/tektoncd/chains/blob/780521656ab0ea2a72d7a969a13040848ac6b969/pkg/chains/objects/objects.go#L98-L99)). Can you open up an issue? I think I'd lean towards adding it to the object types long term.

cc @bcaton85 @lcarva fyi (I think this also affects array results)

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +382,206 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+	}
+
+	wantInputs := []*StructuredSignable{
+		{URI: "gcr.io/foo/bar", Digest: digest3},
+	}
+	gotInputs := ExtractStructuredTargetFromResults(objects.NewTaskRunObject(tr), ArtifactsInputsResultName, logtesting.TestLogger(t))
+	gotInputsEqual := cmp.Equal(gotInputs, wantInputs, cmpopts.SortSlices(func(x, y *StructuredSignable) bool { return x.Digest < y.Digest }))
+	if !gotInputsEqual {
+		t.Errorf("Equal = %v\n reason: %v", gotInputsEqual, cmp.Diff(wantInputs, gotInputs, ignore...))
+	}
```
wlynch, line 432:
```suggestion
```suggestion
	if diff := cmp.Diff(gotInputs, wantInputs, cmpopts.SortSlices(func(x, y *StructuredSignable) bool { return x.Digest < y.Digest })); diff != "" {
		t.Errorf(...)
	}
```

You should use diff here instead of equal - 

cmp.Equal is doing a similar operation as cmp.Diff, so by using cmp.Equal into cmp.Diff you're effectively doing the same thing twice. This also ensures that you're using the same cmpoptions for the comparison and the diff.

pkg/artifacts/signable.go
wlynch: My last comment was vague - my bad 😅

We should check and handle it here, since we don't know what users will hand us.

pkg/artifacts/signable.go
```diff
@@ -257,6 +263,84 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
 	return ss
 }
 
+// RetrieveMeterialsFromStructuredResults retrieves structured results from Tekton Object, and convert them into materials.
+func RetrieveMeterialsFromStructuredResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []slsa.ProvenanceMaterial {
```
wlynch, line 267:
```suggestion
func RetrieveMaterialsFromStructuredResults(obj objects.TektonObject, categoryMarker string, logger *zap.SugaredLogger) []slsa.ProvenanceMaterial {
```

wlynch (changes requested):

pkg/artifacts/signable.go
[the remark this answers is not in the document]
ywluogg: Created #595 for tracking that

pkg/artifacts/signable_test.go
ywluogg: Ah good point!

pkg/artifacts/signable.go
ywluogg: I added extra check here. I previously think that ExtractStructuredTargetFromResults() has handled that once, so it could be unnecessary to add it here as well

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +382,205 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+			URI:    "gcr.io/foo/bar",
+			Digest: map[string]string{"sha256": "05f95b26ed10668b7183c1e2da98610e91372fa9f510046d4ce5812addad86b7"},
+		},
+	}
+
+	gotMaterials := RetrieveMaterialsFromStructuredResults(objects.NewTaskRunObject(tr), ArtifactsInputsResultName, logtesting.TestLogger(t))
+
+	if !cmp.Equal(gotMaterials, wantMaterials, ignore...) {
+		t.Fatalf("Meterials not the same %s", cmp.Diff(gotMaterials, wantMaterials, ignore...))
+	}
```
wlynch, line 485:
```suggestion
	if diff := cmp.Diff(gotMaterials, wantMaterials, ignore...); diff != "" {
		t.Fatalf("Materials not the same %s", diff)
	}
```

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +382,205 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := isStructuredResult(tt.obj, tt.categoryMarker)
+			if got != tt.wantResult {
+				t.Errorf("Validation result is not as the expected: got %v and wanted %v", got, tt.wantResult)
+			}
+			if !cmp.Equal(got, tt.wantResult, ignore...) {
+				t.Errorf("Validation error is not as the expected: %s", cmp.Diff(err, tt.wantErr))
+			}
```
wlynch, line 579:
```suggestion
			if diff := cmp.Diff(got, tt.wantResult, ignore...); diff != "" {
				t.Errorf("Validation error is not as the expected: %s", diff)
			}
```

pkg/chains/formats/intotoite6/pipelinerun/provenance_test.go
```diff
@@ -403,10 +406,28 @@ func TestMetadata(t *testing.T) {
...
+			Name:   "test.io/test/image",
+			Digest: slsa.DigestSet{"sha256": "827521c857fdcd4374f4da5442fbae2edb01e7fbae285c3ec15673d4c1daecb7"},
+		},
+	}
+
+	gotSubjects := extract.SubjectDigests(pro, logtesting.TestLogger(t))
+	gotEqual := cmp.Equal(gotSubjects, wantSubjects, cmpopts.SortSlices(func(x, y intoto.Subject) bool { return x.Name < y.Name }))
+	if !gotEqual {
+		t.Errorf("Equal = %v\n reason: %v", gotEqual, cmp.Diff(wantSubjects, gotSubjects, ignore...))
+	}
```
wlynch, line 432:
```suggestion
	opts := append(ignore,  cmpopts.SortSlices(func(x, y intoto.Subject) bool { return x.Name < y.Name })
	if diff := cmp.Equal(gotSubjects, wantSubjects, opts...); diff != "" {
		t.Errorf(diff)
	}
```

pkg/artifacts/signable_test.go
```diff
@@ -381,6 +382,205 @@ func TestExtractSignableTargetFromResults(t *testing.T) {
...
+
+	wantOutputs := []*StructuredSignable{
+		{URI: "projects/test-project/locations/us-west4/repositories/test-repo/mavenArtifacts/com.google.guava:guava:31.0-jre", Digest: digest1},
+		{URI: "com.google.guava:guava:31.0-jre.pom", Digest: digest2},
+	}
+	gotOutputs := ExtractStructuredTargetFromResults(objects.NewTaskRunObject(tr), ArtifactsOutputsResultName, logtesting.TestLogger(t))
+	gotOutputsEqual := cmp.Equal(gotOutputs, wantOutputs, cmpopts.SortSlices(func(x, y *StructuredSignable) bool { return x.Digest < y.Digest }))
+	if !gotOutputsEqual {
+		t.Errorf("Equal = %v\n reason: %v", gotOutputsEqual, cmp.Diff(wantOutputs, gotOutputs, ignore...))
+	}
```
wlynch, line 441:
```suggestion
	opts := append(ignore, cmpopts.SortSlices(func(x, y *StructuredSignable) bool { return x.Digest < y.Digest }))
	if diff := cmp.Diff(gotOutputs, wantOutputs, opts...); diff != "" {
		t.Error(diff)
	}
```

Also combine opts with the input compare above. (would this make sense to add to ignore?)

wlynch (changes requested):
Almost there. Few more instances of the diff change + waiting on docs.

ywluogg:
@wlynch the doc is added. PTAL

PR #495, opened 2022-07-20
Make grafeas notehint configurable and change the BuildType field

chuangw6:
### Changes
- Add new configmap field - `storage.grafeas.notehint` to allow customization
for the hint field configurable in grafeas ATTESTATION Note
- Change the `BuildType` field in the intoto attestation to the GroupVersionKind
of the run object i.e. `tekton.dev/v1beta1/kind=TaskRun`
- This commit also cleans up the testing code.

### Motivation
- The reason why making field in grafeas note configurable is to allow the users of Chains to customize the information in Grafeas note i.e. they can capture useful information about their service in the note.
- The reason why changing `BuildType` field is that the old value set for this field `https://tekton.dev/attestations/chains@v2` doesn't align with what SLSA spec describes about this field. According to [SLSA spec](https://slsa.dev/provenance/v0.2), `BuildType` field should be an URI indicating **what type of build was performed**. It determines the meaning of invocation, buildConfig and materials. To differentiate the taskrun/pipelinerun, therefore, we use GroupVersionKind information here.

### Release Note
```release-note
Add `storage.grafeas.notehint` field in configmap that allows the note hint field in the grafeas attestation note to be customizable.
```


pkg/config/store_test.go
```diff
@@ -91,6 +91,35 @@ var defaultSigners = SignerConfigs{
 	},
 }
 
+var defaultBuilder = BuilderConfig{
+	"https://tekton.dev/chains/v2",
+	"https://tekton.dev/attestations/chains@v2",
```
wlynch, line 96: Let's include field names so there's no ambiguity.

docs/config.md
```diff
@@ -78,6 +78,8 @@ Supported keys include:
 | `storage.docdb.url` | The go-cloud URI reference to a docstore collection | `firestore://projects/[PROJECT]/databases/(default)/documents/[COLLECTION]?name_field=name`| |
 |`storage.grafeas.projectid`|The project ID to store occurrences|||
 |`storage.grafeas.noteid` (optional)|The note ID to link occurrences. If noteid is not provided, a name in the format of `tekton-<NAMESPACE>` will be used.|||
+|`storage.grafeas.notehint` (optional)|The note hint is a field in the ATTESTATION note. If it is not provided, the default `This attestation note was generated by Tekton Chains` will be used.|||
+|`storage.grafeas.builderversion` (optional)|The builder version is a field in the BUILD note. If it is not provided, `tekton.dev/v1beta1` will be used.|||
```
wlynch, line 82: This probably needs a better description i.e.

- we should probably distinguish between Grafeas attestations/builders vs SLSA/intoto attestations/builders.
- what other values could this field be set to? any consequences that users should look out for?

docs/config.md
```diff
@@ -94,6 +96,7 @@ Tekton Chains supports the following docstore services:
 | Key | Description | Supported Values | Default |
 | :--- | :--- | :--- | :--- |
 | `builder.id` | The builder ID to set for in-toto attestations | | `https://tekton.dev/chains/v2`|
+| `builder.type` | The builder type to set for in-toto attestations | | `https://tekton.dev/attestations/chains@v2`|
```
wlynch, line 99: Do we actually want this to be configurable at the configmap level? Based on https://slsa.dev/provenance/v0.2, the description of this field is:

```
URI indicating what type of build was performed. It determines the meaning of invocation, buildConfig and materials.
```

I'd expect this to identity the data included in the provenance as a TaskRun or PipelineRun

This raises an interesting question though - the current default value may not actually be correct here (particularly as we think about supporting pipelineruns) and we may want to change it. 🤔  cc @lcarva

wlynch (changes requested):

docs/config.md
chuangw6: > Do we actually want this to be configurable at the configmap level?

I think making this field configurable will enable the builders who are built on the top of tekton to set this `builder.type` to whatever they want just like the `builder.id` field.

> I'd expect this to identity the data included in the provenance as a TaskRun or PipelineRun
> This raises an interesting question though - the current default value may not actually be correct here (particularly as we think about supporting pipelineruns) and we may want to change it. 🤔 cc @lcarva

Sounds interesting. Based on https://slsa.dev/provenance/v0.2, this field should be typeUri. Any idea on what the identifying uri should look like? Should it still be something like `https://tekton.dev/...`?

pkg/config/store_test.go
chuangw6: SG!

docs/config.md
chuangw6: > * we should probably distinguish between Grafeas attestations/builders vs SLSA/intoto attestations/builders.

AFAIK, the builder referred in grafeas notes is the same as the builder referred in the intoto statement. Grafeas note needs the `builderVersion` information, whereas builder.type and builder.id fields are needed in the intoto statement.

> * what other values could this field be set to? any consequences that users should look out for?

No. For the notehint and builderVersion, there are no rules on those string values.

wlynch: > AFAIK, the builder referred in grafeas notes is the same as the builder referred in the intoto statement. Grafeas note needs the `builderVersion` information, whereas builder.type and builder.id fields are needed in the intoto statement.

👍 I think where I got tripped up was I just assumed this was the intoto attestation, then was confused when I couldn't find an builder version field. Being explicit about what types this affects will help short-circuit the search.

> No. For the notehint and builderVersion, there are no rules on those string values.

https://github.com/grafeas/grafeas/blob/cd23d4dc1bef740d6d6d90d5007db5c9a2431c41/proto/v1/build.proto#L31 says

```
// Required. Immutable. Version of the builder which produced this build.
```

Should we be generating this from the Pipeline/Chains version rather than letting this be configurable via configmap?

docs/config.md
wlynch: > I think making this field configurable will enable the builders who are built on the top of tekton to set this `builder.type` to whatever they want just like the `builder.id` field.

It's a bit confusing though since the type we're setting in the intoto attestation is about the build (`buildType`), not the builder. I'd expect this to match the type we're attesting to, so I'm not sure if we want this to be configurable by service providers.
 
> > I'd expect this to identity the data included in the provenance as a TaskRun or PipelineRun
> > This raises an interesting question though - the current default value may not actually be correct here (particularly as we think about supporting pipelineruns) and we may want to change it. 🤔 cc @lcarva
> 
> Sounds interesting. Based on https://slsa.dev/provenance/v0.2, this field should be typeUri. Any idea on what the identifying uri should look like? Should it still be something like `https://tekton.dev/...`?

¯\_(ツ)_/¯ I think we have some flexibility here - maybe `tekton://<GroupVersionKind>`?

chuangw6:
Offline discussed with @wlynch.

We decided to just allow the hint field in grafeas note to be configurable.

- The `buildType` in the slsa predicate of intoto attestation will be set to run object's groupversion kind.
- The `builderVersion` in the grafeas build note will be set to an annotation field of the run object that will be added in pipeline. See https://github.com/tektoncd/pipeline/issues/5303 for more information.

docs/config.md
chuangw6: Updated this to be `tekton.dev/v1beta1/kind=TaskRun`.

I didn't use [`GetGroupVersionKind().String()`](https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/runtime/schema/group_version.go#L161-L163) directly since it will give us a string like `tekton.dev/v1beta1, Kind=TaskRun`.

Please have a look and let me know what you think.
Thanks

pkg/chains/formats/intotoite6/intotoite6.go
```diff
@@ -83,7 +82,7 @@ func (i *InTotoIte6) generateAttestationFromTaskRun(tr *v1beta1.TaskRun) (interf
 			Builder: slsa.ProvenanceBuilder{
 				ID: i.builderID,
 			},
-			BuildType:   tektonID,
+			BuildType:   fmt.Sprintf("%s/kind=%s", tr.GetGroupVersionKind().GroupVersion().String(), tr.GetGroupVersionKind().Kind),
```
wlynch, line 85: Should we just simplify this to `<group>/<version>/<kind>`? Does `kind=` serve a special purpose anywhere else (i.e. parsing, etc)?

wlynch (commented):
lgtm! one last comment

pkg/chains/formats/intotoite6/intotoite6.go
chuangw6: > Does kind= serve a special purpose anywhere else (i.e. parsing, etc)?

No off the top of my head. I added `kind=` to just try to match what [GetGroupVersionKind().String()](https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/runtime/schema/group_version.go#L161-L163) returns but no space there. 

> Should we just simplify this to <group>/<version>/<kind>? 

SGTM! Changed.

docs/config.md
lcarva: Is `tekton.dev/v1beta1/kind=TaskRun` a valid URI? Not sure how strict this needs to be.

I also agree that this should not be configurable because it would imply the format of the intoto attestation is also configurable, which is certainly not the case today.

chuangw6: > Is tekton.dev/v1beta1/kind=TaskRun a valid URI? Not sure how strict this needs to be.

Changed to `<group>/<version>/<kind>` i.e. tekton.dev/v1beta1/TaskRun. wdyt?


> I also agree that this should not be configurable because it would imply the format of the intoto attestation is also configurable, which is certainly not the case today.

Agree but interestingly Chains today allows `predicate.builder.id` to be [configurable](https://github.com/tektoncd/chains/blob/main/docs/config.md#in-toto-configuration).

ctl.txt
```diff
@@ -0,0 +1,635 @@
+2022/08/03 08:54:16 Registering 3 clients
```
wlynch, line 1: Looks like a debug log made it's way into the PR 😱

docs/config.md
```diff
@@ -76,8 +76,10 @@ Supported keys include:
 | `storage.gcs.bucket` | The GCS bucket for storage | | |
 | `storage.oci.repository` | The OCI repo to store OCI signatures and attestation in | If left undefined _and_ one of `artifacts.{oci,taskrun}.storage` includes `oci` storage, attestations will be stored alongside the stored OCI artifact itself. ([example on GCP](../images/attestations-in-artifact-registry.png)) Defining this value results in the OCI bundle stored in the designated location _instead of_ alongside the image. See [cosign documentation](https://github.com/sigstore/cosign#specifying-registry) for additional information. | |
 | `storage.docdb.url` | The go-cloud URI reference to a docstore collection | `firestore://projects/[PROJECT]/databases/(default)/documents/[COLLECTION]?name_field=name`| |
+<<<<<<< HEAD
```
wlynch, line 79: Merge conflict leftover

wlynch (changes requested):

ctl.txt
chuangw6: my bad... 😭

wlynch:
Looks good! Last thing is we should add a release-notes section to the PR (since I think the release tool will pick up on this) with the user facing changes.

chuangw6:
> Looks good! Last thing is we should add a release-notes section to the PR (since I think the release tool will pick up on this) with the user facing changes.

SG! Added!

wlynch (approved):
/lgtm

PR #509, opened 2022-08-05
Use industry standard names for payloadType field

chuangw6:
- `application/vnd.dev.cosign.simplesigning.v1+json` for simplesigning
payload
- `application/vnd.in-toto+json` for intoto payload


chuangw6:
/assign @wlynch 

PTAL. Thanks!

wlynch (approved):
/lgtm

PR #510, opened 2022-08-09
Clarify documentation for `storage.oci.repository`

bendory:


bendory:
@chuangw6 PTAL

bendory:
@Yongxuanzhang PTAL

docs/config.md
```diff
@@ -74,7 +74,7 @@ Supported keys include:
 | Key | Description | Supported Values | Default |
 | :--- | :--- | :--- | :--- |
 | `storage.gcs.bucket` | The GCS bucket for storage | | |
-| `storage.oci.repository` | The OCI repo to store OCI signatures in  | | |
+| `storage.oci.repository` | The OCI repo to store OCI signatures in | If left undefined _and_ one of `artifacts.{oci,taskrun}.storage` includes `oci` storage, attestations will be stored alongside the stored OCI artifact itself. ([example on GCP](../images/attestations-in-artifact-registry.png)) Defining this value results in the OCI bundle stored in the designated location _instead of_ alongside the image. Note that for OCI storage using Artifact Registry, this field should be a full image name like `$LOCATION-docker.pkg.dev/$PROJECT/$REPO/$STORAGE_IMAGE`, not just a repository. | |
```
Yongxuanzhang, line 77: > Note that for OCI storage using Artifact Registry
 
Artifact Registry is GCP's service, can we add a link to this term? I think for OSS we cannot assume the users know GCP

Yongxuanzhang (commented):
Thanks for adding this clarification. And for your screenshot maybe we should hide some personal information like your project id etc?

docs/config.md
priyawadhwa: We could point to [cosign docs](https://github.com/sigstore/cosign#specifying-registry) here as well since this is using that feature

Yongxuanzhang: From cosign's doc it looks like for GCR we don't have to specify the image name.

Yongxuanzhang: Not sure if this is a bug? Supposedly we should only provide a repo and cosign should be able to push the attestation and sig to that repo?

chuangw6: > From cosign's doc it looks like for GCR we don't have to specify the image name.

GCR doesn't have the concept of repository. When you specify a GCR path, the image name is included. For example, in the cosign doc, it says `gcr.io/dlorenc-vmtest2/demo` in which demo is the image name.

bendory:
> Thanks for adding this clarification! And for your screenshot maybe we should hide some personal information like your project id etc?

I thought about that; the project is already deleted, so it should be harmless.

docs/config.md
bendory: Updated here to point to cosign docs at https://github.com/sigstore/cosign#specifying-registry.

I'm heading over there now to clarify Artifact Registry usage.

Yongxuanzhang: > > From cosign's doc it looks like for GCR we don't have to specify the image name.
> 
> GCR doesn't have the concept of repository. When you specify a GCR path, the image name is included. For example, in the cosign doc, it says `gcr.io/dlorenc-vmtest2/demo` in which demo is the image name.

It says export COSIGN_REPOSITORY=gcr.io/my-new-repo, there is no image name in it

docs/config.md
```diff
@@ -74,7 +74,7 @@ Supported keys include:
 | Key | Description | Supported Values | Default |
 | :--- | :--- | :--- | :--- |
 | `storage.gcs.bucket` | The GCS bucket for storage | | |
-| `storage.oci.repository` | The OCI repo to store OCI signatures in  | | |
+| `storage.oci.repository` | The OCI repo to store OCI signatures in | If left undefined _and_ one of `artifacts.{oci,taskrun}.storage` includes `oci` storage, attestations will be stored alongside the stored OCI artifact itself. ([example on GCP](../images/attestations-in-artifact-registry.png)) Defining this value results in the OCI bundle stored in the designated location _instead of_ alongside the image. See [cosign documentation](https://github.com/sigstore/cosign#specifying-registry) for additional information. | |
```
Yongxuanzhang, line 77: Sorry just one more, 
```suggestion
| `storage.oci.repository` | The OCI repo to store OCI signatures and attestation in | If left undefined _and_ one of `artifacts.{oci,taskrun}.storage` includes `oci` storage, attestations will be stored alongside the stored OCI artifact itself. ([example on GCP](../images/attestations-in-artifact-registry.png)) Defining this value results in the OCI bundle stored in the designated location _instead of_ alongside the image. See [cosign documentation](https://github.com/sigstore/cosign#specifying-registry) for additional information. | |
```

Yongxuanzhang (commented):
.

PR #512, opened 2022-08-09
Pinning sigstore/scaffolding/actions/setup version

bcaton85:


.github/workflows/kind-e2e.yaml
```diff
@@ -67,6 +67,8 @@ jobs:
 
     - name: Install mirror, kind, knative + sigstore
       uses: sigstore/scaffolding/actions/setup@main
+      with:
+        version: v0.3.0
```
vaikas, line 71: Also please add this:
k8s-version 1.22

so like this:
```
with:
  version: "v0.3.0"
  k8s-version: "v1.22.x"
```

vaikas: Because 23 is being goofy:
https://github.com/sigstore/scaffolding/issues/81

bcaton85: Gotcha, updated.

.github/workflows/kind-e2e.yaml
```diff
@@ -67,6 +67,9 @@ jobs:
...
+        k8s-version: "v1.22.x"
```
wlynch, line 72:
```suggestion
        k8s-version: ${{ matrix.k8s-version }}
```

wlynch (changes requested):

bcaton85:
@vaikas seems `v1.21.x` is failing. We're currently using `k8s-version: ${{ matrix.k8s-version }}`. Is there an easy fix for that? Not sure if we want to remove it from the test

vaikas:
@bcaton85 So, it seems like the tests were passing fine just a few days ago here:

https://github.com/tektoncd/chains/runs/7697397547?check_suite_focus=true

What's curious is that it was running very old version

```
SIGSTORE_SCAFFOLDING_RELEASE_VERSION: v0.1.19
```

So I wonder if you could just use an older version of the action, instead of using `main`, maybe grab this hash:
https://github.com/sigstore/scaffolding/releases/tag/v0.3.0

I guess your uses would look like this:
```
uses: sigstore/scaffolding/actions/setup@74d1900f01c1f717f121e0436c4e1996fa55e986
```

And take out the with. At least I think that should run the version that was working just fine before.

wlynch:
@vaikas - not sure if that version was actually used - https://github.com/tektoncd/chains/runs/7697397547?check_suite_focus=true#step:7:114 looks like it ignored the env var and used 0.3.0.

I'm half tempted to just drop 1.21 - it's EOL anyway. 🤔

bcaton85:
Sounds good, updated to use the hash.

PR #513, opened 2022-08-09
Pin to scaffolding v0.4.2, test with k8s 1.22, 1.23, and 1.24.

vaikas:

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -67,9 +67,6 @@ jobs:
 
     - name: Install mirror, kind, knative + sigstore
       uses: sigstore/scaffolding/actions/setup@main
-      with:
-        version: v0.3.0
-        k8s-version: v1.22.x
 
     - name: Install Tekton pipelines
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/kind-e2e.yaml
```diff
@@ -66,7 +66,7 @@ jobs:
         path: ./src/github.com/tektoncd/chains
 
     - name: Install mirror, kind, knative + sigstore
-      uses: sigstore/scaffolding/actions/setup@main
+      uses: sigstore/scaffolding/actions/setup@74d1900f01c1f717f121e0436c4e1996fa55e986
```
priyawadhwa, line 69: think this might need

```
with:
  version: v0.3.0
```

if we're cool with pinning to an older version?

vaikas: Well, I think this might do it (first try ;) )

https://github.com/sigstore/scaffolding/pull/277

That would then allow us to update the chains tests to use proper TUF roots.

wlynch (approved):
Thanks @vaikas! 

/lgtm

PR #519, opened 2022-08-12
Fix for issue #515, This PR ensures only those signers which are specified as signers for any artifact are initialized.

### LINKED ISSUE #515: Noisy error message in controller log
# Expected Behavior

If users only configures KMS as the signer for both taskrun and oci artifact and `signing-secrets` is not configured, controller log doesn't complain anything about x509.

# Actual Behavior

If the kms is chose as the signer for both taskrun and oci artifact and `signing-secrets` is not configured, controller log contains the following error message. This error message might be misleading for users who look into log and find the error message is not related to the bug they are investigating.

```
{"level":"warn","ts":"2022-08-03T20:30:18.025Z","logger":"watcher","caller":"chains/signing.go:64","msg":"error configuring x509 signer: no valid private key found, looked for: [x509.pem, cosign.key]","commit":"b78ee5e","knative.dev/controller":"github.com.tektoncd.chains.pkg.reconciler.taskrun.Reconciler","knative.dev/kind":"tekton.dev.TaskRun","knative.dev/traceid":"4b39eff0-bdca-424d-a2f8-25c9c991bfbf","knative.dev/key":"default/kaniko-run-mmczh"}
```

1. install chains & pipeline
2. (do not create `signing-secret`)
3. configure kms as the signer for both oci and taskrun artifact
4. configure other appropriate fields for formatter and storage. 
5. run task&taskrun of your choice
6. check chains controller log

jagathprakash:
# Changes
Fixes #515
This PR ensures only those signers which are specified as signers for any artifact are initialized. 
This change ensures that we do not initialize unwanted signers for which we might not have all necessary configs, which was resulting in unwanted errors in logs.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/chains/signing_test.go
+++ b/pkg/chains/signing_test.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/sigstore/rekor/pkg/generated/models"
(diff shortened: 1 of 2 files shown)


pkg/chains/signing_test.go
```diff
@@ -325,6 +326,89 @@ func TestTaskRunSigner_Transparency(t *testing.T) {
...
+				Artifacts: config.ArtifactConfigs{
+					TaskRuns: config.Artifact{
+						Format:         "tekton",
+						StorageBackend: sets.NewString("mock"),
+						Signer:         "x509",
+					},
+				},
+				Transparency: config.TransparencyConfig{
+					Enabled: false,
+				},
```
wlynch, line 351: This can be omitted (this will 0 value default to false anyway)

pkg/chains/signing_test.go
```diff
@@ -325,6 +326,89 @@ func TestTaskRunSigner_Transparency(t *testing.T) {
...
+	logger := logging.FromContext(ctx)
+	tests := []struct {
+		name       string
+		signers    []string
+		config     config.Config
+		SecretPath string
+	}{
+		{
+			name:    "x509",
+			signers: [](string){signing.TypeX509},
```
wlynch, line 340: nit: `[]string{signing.TypeX509}`

wlynch (changes requested):
LGTM!

Main piece of feedback is let's update the commit message / PR body to add more details on what the change does - https://github.com/tektoncd/community/blob/main/standards.md#commits

My rule of thumb is you should be able to have a rough idea of what the change is doing even if you only have access to `git log` / `git blame`.

pkg/chains/signing_test.go
jagathprakash: Done.

jagathprakash:
> LGTM!
> 
> Main piece of feedback is let's update the commit message / PR body to add more details on what the change does - https://github.com/tektoncd/community/blob/main/standards.md#commits
> 
> My rule of thumb is you should be able to have a rough idea of what the change is doing even if you only have access to `git log` / `git blame`.

@wlynch thanks for the review! Agree to the requirement for detailed messaging. I have updated the PR message as well as the commit to be more detailed.

wlynch (approved):
/lgtm

PR #527, opened 2022-08-24
Fix param value ordering in materials

chuangw6:
# Changes

Prior to the fix, when generating `materials` section of the intoto
attestation, Chains gets the param value from TaskRun spec first and then
default from Task spec. This overwriting will cause the problem that
the actual value used in run is from taskrun spec, but the value recorded
in `materials` is the value defined in task spec's `default`.

In this PR, we fix the issue by making sure that the default parameters are 
always overwritten by values from the taskrun spec.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/chains/formats/intotoite6/intotoite6_test.go
+++ b/pkg/chains/formats/intotoite6/intotoite6_test.go
@@ -62,12 +62,12 @@ func TestCreatePayload1(t *testing.T) {
 				BuildFinishedOn: &e1BuildFinished,
 			},
 			Materials: []slsa.ProvenanceMaterial{
-				{URI: "git+https://git.test.com.git", Digest: slsa.DigestSet{"sha1": "abcd"}},
+				{URI: "git+https://git.test.com.git", Digest: slsa.DigestSet{"sha1": "sha:taskrun"}},
 			},
 			Invocation: slsa.ProvenanceInvocation{
 				Parameters: map[string]v1beta1.ArrayOrString{
 					"IMAGE":             {Type: "string", StringVal: "test.io/test/image"},
-					"CHAINS-GIT_COMMIT": {Type: "string", StringVal: "abcd"},
+					"CHAINS-GIT_COMMIT": {Type: "string", StringVal: "sha:taskrun"},
 					"CHAINS-GIT_URL":    {Type: "string", StringVal: "https://git.test.com"},
 					"filename":          {Type: "string", StringVal: "/bin/ls"},
 				},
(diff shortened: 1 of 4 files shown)


pkg/chains/formats/intotoite6/intotoite6_test.go
```diff
@@ -62,13 +62,13 @@ func TestCreatePayload1(t *testing.T) {
 				BuildFinishedOn: &e1BuildFinished,
 			},
 			Materials: []slsa.ProvenanceMaterial{
-				{URI: "git+https://git.test.com.git", Digest: slsa.DigestSet{"sha1": "abcd"}},
+				{URI: "git+https://git1.test.com.git", Digest: slsa.DigestSet{"sha1": "sha:taskrun"}},
 			},
 			Invocation: slsa.ProvenanceInvocation{
 				Parameters: map[string]v1beta1.ArrayOrString{
 					"IMAGE":             {Type: "string", StringVal: "test.io/test/image"},
-					"CHAINS-GIT_COMMIT": {Type: "string", StringVal: "abcd"},
```
lcarva, line 70: It would be great to improve the test coverage to ensure the default value is indeed used if there are results and no explicit parameters.

lcarva (approved):

pkg/chains/formats/intotoite6/testdata/taskrun1.json
```diff
@@ -7,11 +7,11 @@
             },
             {
                 "name": "CHAINS-GIT_COMMIT",
-                "value": "abcd"
+                "value": "sha:taskrun"
             },
             {
                 "name": "CHAINS-GIT_URL",
-                "value": "https://git.test.com"
```
Yongxuanzhang, line 14: Silly question, why change this to git1?

Yongxuanzhang (commented):
/LGTM 
This looks like a bug! Great work for fixing it!

wlynch (approved):

pkg/chains/formats/intotoite6/intotoite6_test.go
chuangw6: > ensure the default value is indeed used if there are results and no explicit parameters.

Sorry do you mean if there are **NO** results and no explicit params? If so, that makes sense to me. I just added the case to `taskrun2.json`. Please take a look.
Thanks

pkg/chains/formats/intotoite6/testdata/taskrun1.json
chuangw6: Aha, I just wanted to differentiate it with the value defined in default. Changed this back and named value in default `defaultgit.test.com` haha.

Issue #522, opened 2022-08-17
Upgrade to pipeline v1beta1 types


wlynch:
A bunch of v1alpha1 types have been removed in [Pipelines 0.38](https://github.com/tektoncd/pipeline/releases/tag/v0.38.0). This is blocking upgrade of the Pipelines dep - https://github.com/tektoncd/chains/pull/516

We should move our dependency to v1beta1.

wlynch:
@ywluogg Interested in picking this up? 👀

ywluogg:
> @ywluogg Interested in picking this up? 👀

Yes!

PR #530, opened 2022-08-24
Remove Tekton Pipelines v1alpha1 type dependencies

ywluogg:
# Changes

Remove Tekton Pipelines v1alpha1 type dependencies since v0.38 is removing some v1alpha1 types.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Remove Tekton Pipelines v1alpha1 type dependencies.
```

### DIFF
--- a/pkg/artifacts/signable.go
+++ b/pkg/artifacts/signable.go
@@ -88,13 +88,6 @@ type StructuredSignable struct {
 
 func (oa *OCIArtifact) ExtractObjects(tr *v1beta1.TaskRun) []interface{} {
 	imageResourceNames := map[string]*image{}
-	if tr.Status.TaskSpec != nil && tr.Status.TaskSpec.Resources != nil {
-		for _, output := range tr.Status.TaskSpec.Resources.Outputs {
-			if output.Type == v1beta1.PipelineResourceTypeImage {
-				imageResourceNames[output.Name] = &image{}
-			}
-		}
-	}
 
 	for _, rr := range tr.Status.ResourcesResult {
 		img, ok := imageResourceNames[rr.ResourceName]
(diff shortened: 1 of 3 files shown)


ywluogg:
#522

pkg/artifacts/signable.go
```diff
@@ -88,13 +88,6 @@ type StructuredSignable struct {
 
 func (oa *OCIArtifact) ExtractObjects(tr *v1beta1.TaskRun) []interface{} {
 	imageResourceNames := map[string]*image{}
-	if tr.Status.TaskSpec != nil && tr.Status.TaskSpec.Resources != nil {
-		for _, output := range tr.Status.TaskSpec.Resources.Outputs {
-			if output.Type == v1beta1.PipelineResourceTypeImage {
-				imageResourceNames[output.Name] = &image{}
-			}
-		}
-	}
```
Yongxuanzhang, line 97: Is `tr.Status.TaskSpec.Resources.Outputs` v1alpha1?
Just feel a bit confused what v1alpha1 dependency we are removing here.

Yongxuanzhang (commented):
Do you mind including the specific removed type dependencies in your PR description, commit msg and release note?

pkg/artifacts/signable.go
```diff
@@ -88,13 +88,6 @@ type StructuredSignable struct {
 
 func (oa *OCIArtifact) ExtractObjects(tr *v1beta1.TaskRun) []interface{} {
 	imageResourceNames := map[string]*image{}
-	if tr.Status.TaskSpec != nil && tr.Status.TaskSpec.Resources != nil {
```
chuangw6, line 91: We should also remove this in `func materials(tr *v1beta1.TaskRun)` in which resource input params will overwrite others.

chuangw6 (commented):
Thanks @ywluogg for removing these!

pkg/artifacts/signable.go
ywluogg: So the logic mainly focuses on v1beta1.PipelineResourceTypeImage. When this is removed, there's no logic left within this for loop anymore

pkg/artifacts/signable.go
ywluogg: That's a good catch.

PR #564, opened 2022-09-22
Use event updates for PR reconciliation

lcarva:
# Changes

A PipelineRun should only be processed when the underlying TaskRun tasks have been signed. This prevents timing issues where the Chains controller may attempt to push both TaskRun and PipelineRun attestations to an OCI repository at the same time, causing one of them to be lost.

Prior to this change, a PipelineRun that was not ready to be processed was requeued so it could be re-processed in 15 seconds. Given that this scenario happens often, this caused the Chains controller to be overwhelmed if there were too many TaskRun and/or PipelineRun resources on the cluster.

This commit modifies the logic so when Chains signs a TaskRun, it also updates an annotation, `chains.tekton.dev/child-signed`, on the parent PipelineRun, if any. The annotation update triggers the PipelineRun to be reconciled again. Thus the logic is driven by "events" instead of a polling mechanism.

Notify parent object after child is reconciled


pkg/chains/objects/objects.go
```diff
@@ -107,6 +108,20 @@ func (tro *TaskRunObject) GetServiceAccountName() string {
 	return tro.Spec.ServiceAccountName
 }
 
+// GetParent returns the PipelineRun object for the TaskRun, if one exists.
+func (tro *TaskRunObject) GetParent(ctx context.Context, clientSet versioned.Interface) (TektonObject, error) {
+	prName, ok := tro.GetLabels()["tekton.dev/pipelineRun"]
```
wlynch, line 113: nit: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference is the more - authoritative source for this.

pkg/chains/signing.go
```diff
@@ -247,8 +248,22 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
...
+	if err := MarkSigned(ctx, tektonObj, o.Pipelineclientset, extraAnnotations); err != nil {
+		return err
+	}
+
+	// Notify parent object that child object has been signed. This allows PipelineRun resources
+	// to be reconciled when the status of its TaskRuns use ChildReferences.
+	if parent, err := tektonObj.GetParent(ctx, o.Pipelineclientset); err != nil {
+		return err
+	} else if parent != nil {
+		if err := MarkParentSigned(ctx, parent, o.Pipelineclientset); err != nil {
```
wlynch, line 261: Do we need to wait until all Tasks are done before doing this? 

In particular I'm thinking about if we have a matrix pipeline that builds images for different arch in different Tasks - if we set this too early the Pipeline sig might still clash with other Tasks. 

Maybe we invert this relationship and have objects query for their parent status? 🤔

wlynch:
Also a few general questions:

1. Is this primarily affecting OCI artifacts w/ OCI storage since we're trying to update the same `.sig` file, or did you observe this for tekton artifacts as well?
2. Are we actually losing data / events when this happens (i.e. the events are successfully writing over each other), or are we just getting a conflict error and we'll eventually get to the right state?

pkg/chains/signing.go
lcarva: > Do we need to wait until all Tasks are done before doing this?

In reality, we just need to wait until the last Task within the Pipeline that has signable subjects. We could enhance the logic to handle that, not sure how much more beneficial that would be though.

> In particular I'm thinking about if we have a matrix pipeline that builds images for different arch in different Tasks - if we set this too early the Pipeline sig might still clash with other Tasks.

I think in this case, each of those Tasks produces a different image (unique digest). So the Tasks themselves wouldn't interfere with each other's signatures and attestations.

> Maybe we invert this relationship and have objects query for their parent status?

Interesting idea. It would mean that signatures and attestations may have a perceivable delay in being created (maybe?). Not sure if that's an issue.

lcarva:
>     1. Is this primarily affecting OCI artifacts w/ OCI storage since we're trying to update the same `.sig` file, or did you observe this for tekton artifacts as well?

Unlike the taskrun reconciler, the pipelinerun reconciler doesn't sign images, it only creates attestations. But, yes, the same concept applies. This is not an issue when other types of storage are used, afaik.
 
>     2. Are we actually losing data / events when this happens (i.e. the events are successfully writing over each other), or are we just getting a conflict error and we'll eventually get to the right state?

Yes, we lose data and events. If both the taskrun and pipelinerun reconcilers run at the same time, either the taskrun or pipelinerun attestation is lost. We saw this early on in the POC. This is why the check was added to only reconcile pipelineruns once their taskruns have been reconciled. If we don't poll (like the code used to), and don't update the parent resource (like this PR does), then there's nothing that triggers another pipelinerun reconciliation, thus the pipelinerun attestation is not created. (Oddly enough, if you do delete the pipelinerun then the attestation is created due to the use of finalizers. I don't think we should rely on that though.)

pkg/chains/objects/objects.go
lcarva: Thanks for the suggestion. Updated.

pkg/chains/signing.go
wlynch: > In reality, we just need to wait until the last Task within the Pipeline that has signable subjects. We could enhance the logic to handle that, not sure how much more beneficial that would be though.

Makes sense! This is just waiting until 1 Task is done before annotating the PipelineRun to go though, right? That's why I'm wondering if we should just wait until the pipeline clears first before we handle the TaskRuns 🤔 

> I think in this case, each of those Tasks produces a different image (unique digest). So the Tasks themselves wouldn't interfere with each other's signatures and attestations.

Hmmm I think I'm not fully grasping the problem then - if TaskRuns are pushing unique digests that won't interfere with each other, how does this interfere with the PipelineRun? 👀 

> Interesting idea. It would mean that signatures and attestations may have a perceivable delay in being created (maybe?). Not sure if that's an issue.

I'm not super concerned - chains is already async so there's an expectation of some delay. We could use https://pkg.go.dev/knative.dev/pkg/controller#Impl.EnqueueKey to force an immediate requeue of sub resources to help speed things up if needed.

wlynch:
> Unlike the taskrun reconciler, the pipelinerun reconciler doesn't sign images, it only creates attestations. But, yes, the same concept applies. This is not an issue when other types of storage are used, afaik.

Could this be avoided if we used a different naming scheme for OCI storage for Run attestations to avoid collision?

pkg/chains/signing.go
lcarva: > Makes sense! This is just waiting until 1 Task is done before annotating the PipelineRun to go though, right? 

This updates the PipelineRun whenever a TaskRun finishes. So if the PipelineRun contains 5 TaskRuns, the annotation on the PipelineRun will get updated 5 times.

> That's why I'm wondering if we should just wait until the pipeline clears first before we handle the TaskRuns

I think to make this work, we would need to update an annotation on the TaskRun to indicate whether or not it can be proceed by Chains, right? That might work. I'm not sure what the main benefit would be though. Can you help me understand?

> Hmmm I think I'm not fully grasping the problem then - if TaskRuns are pushing unique digests that won't interfere with each other, how does this interfere with the PipelineRun?

Because if you want a PipelineRun attestation for each one of those images, they need to become Pipeline results. Let's say a PipelineRun contains three TaskRuns, each building a unique image (e.g. multi-arch). Since they're unique, they have different digests, so even if they're pushed to the same OCI repo, they'll each have their own signature and attestation images. (This is also the case prior to TEP-84). Now let's say the PipelineRun produces multiple images in its result, which makes sense since at the scope of a PipelineRun, multiple images were built. The PipelineRun reconciler is going to generate the PipelineRun attestation (which is the same for all three images), and push them not only to the same OCI repo, but also to the same `sha256-<digest>.att` tag. cosign is smart enough to not overwrite the attestations that are already there, so you end up with attestation images that contain two layers. The first layer contains the TaskRun attestation (unique across all three images), and the second containing the PipelineRun attestation (the same across all three images).

lcarva:
> Could this be avoided if we used a different naming scheme for OCI storage for Run attestations to avoid collision?

Something like `sha256-<digest>.pr.at` ? Yes, it would absolutely avoid this sort of issue. I'm not sure if that's a good idea though since cosign wouldn't know about it.

wlynch:
> > Could this be avoided if we used a different naming scheme for OCI storage for Run attestations to avoid collision?
> 
> Something like `sha256-<digest>.pr.at` ? Yes, it would absolutely avoid this sort of issue. I'm not sure if that's a good idea though since cosign wouldn't know about it.

I was misunderstanding the problem. 🙃  I get it now - we can hit this case any time we try to attach things to an OCI image, which can happen for simple signing, but also when we try and attach the build provenance to the image as well. I was getting tripped up thinking about the case where we are just using an OCI registry as a storage bucket for build provenance w/o attaching it to an image, which should be safe from this.

+1 to staying consistent with cosign.

This could be solved with a bit of retries if we could pass an etag, though looks like this is still WIP - https://github.com/google/go-containerregistry/issues/1333. /cc @imjasonh - any OCI wisdom you can pass down here?

pkg/chains/signing.go
wlynch: Okay, now that I'm caught up and grok everything, I think we can actually do this with less changes!

> I think to make this work, we would need to update an annotation on the TaskRun to indicate whether or not it can be proceed by Chains, right? That might work. I'm not sure what the main benefit would be though. Can you help me understand?

I think we already have this in the `chains.tekton.dev/signed` annotation + the reconciled check here: https://github.com/tektoncd/chains/blob/ad5a650eca7cc0d3019d5712bc945f906c2a4d1b/pkg/chains/annotations.go#L36-L43

I think what we probably want to do is swap out the direct client in https://github.com/tektoncd/chains/blob/ad5a650eca7cc0d3019d5712bc945f906c2a4d1b/pkg/reconciler/pipelinerun/pipelinerun.go#L94 for the lister cache via the informer from https://pkg.go.dev/github.com/tektoncd/pipeline@v0.40.0/pkg/client/injection/informers/pipeline/v1beta1/taskrun . This should be safe to query as much as we want and should generally have the latest-ish values.

To trigger the resync from the TaskRun, you can just requeue the parent object key using https://pkg.go.dev/knative.dev/pkg@v0.0.0-20220921024409-d1d5c849073b/controller#Impl.EnqueueKey

This should block the PipelineRun reconcile until all the TaskRuns are done and minimize the # of API calls the PipelineRun reconciler is making to query TaskRun state.

> The reason I chose the current approach is because I wanted to reconcile resources as early as possible. Waiting to reconcile TaskRuns until the whole PipelineRun completes goes against that. I'm not saying that we shouldn't do that, just clarifying my thought process.

Got it. Makes sense!

lcarva: ooh I like this idea.

I replaced the line with:
```go
taskruninformer.Get(ctx).Lister().TaskRuns(pr.Namespace).Get(name)
```

But I'm getting a panic when that line is run:
```go
Unable to fetch github.com/tektoncd/pipeline/pkg/client/informers/externalversions/pipeline/v1beta1.TaskRunInformer from context.
```

I'm not well-versed here, but it sounds like maybe I also need to do some setup on the informer.

wlynch: Yeah I always forget where the client injection needs to go. Try passing it when the controller is created similar to the pipelinerun informer. https://github.com/tektoncd/chains/blob/ad5a650eca7cc0d3019d5712bc945f906c2a4d1b/pkg/reconciler/pipelinerun/controller.go#L33

lcarva: Thanks for the tip :pray:  Pushed a new commit that takes this approach. It should further improve performance.

For the second part of your suggestion, I'm not sure if it'll work. It would require the TaskRun controller/reconciler to add something on the queue of the PipelineRun controller/reconciler. I'm likely missing something obvious here though.

wlynch: Shout out to @vaikas for this!

We should be able to use a tracker for this to watch a PipelineRun's children:

https://github.com/sigstore/policy-controller/blob/main/pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go#L204 for an example

lcarva: Kudos to @vaikas!

Added a commit that takes this approach.

@wlynch, if the current changes look good to you, I can squash the commits if you prefer.

(The failing e2e test seems to be unrelated to these changes because it's also failing on my PR that updates just the docs.)

wlynch (approved):
Looks great!

/lgtm

vaikas:
So, this looks right to me! Glad the Tracker worked out :)

The test seems to be failing (I think?) because of this:
The attestation that we get back has this:
```
        {
          "entryPoint": "/ko-app/imagedigestexporter",
          "arguments": [
            "-images",
            "[{\"name\":\"builtImage\",\"type\":\"image\",\"url\":\"gcr.io/foo/bar\",\"digest\":\"\",\"OutputImageDir\":\"/workspace/sourcerepo\"}]"
          ],
          "environment": {
            "container": "image-digest-exporter-4nq2b",
            "image": "docker-pullable://gcr.io/tekton-nightly/github.com/tektoncd/pipeline/cmd/imagedigestexporter@sha256:ea087a3978cde45884aaa01354939d82352d9c49b3d6c44a2fb1c22a681e59aa"
          },
          "annotations": null
        }
```

And what's is expected is this:
```
                        {
                            "entryPoint": "",
                            "arguments": null,
                            "environment": {
                                "container": "image-digest-exporter-4nq2b",
                                "image": "docker-pullable://gcr.io/tekton-nightly/github.com/tektoncd/pipeline/cmd/imagedigestexporter@sha256:ea087a3978cde45884aaa01354939d82352d9c49b3d6c44a2fb1c22a681e59aa"
                            },
                            "annotations": null
                        }

```

Notice the entryPoint and arguments are different. Could it be a newer version of ko, or something else?

lcarva:
@wlynch, lgtm label removed because I had to rebase to pickup the e2e test workaround :pray:

lcarva:
@wlynch, @priyawadhwa, this PR is approved but needs lgtm label. Can I just add it to my own PRs? Or is that not a recommended practice?

imjasonh:
> This could be solved with a bit of retries if we could pass an etag, though looks like this is still WIP - [google/go-containerregistry#1333](https://github.com/google/go-containerregistry/issues/1333). /cc @imjasonh - any OCI wisdom you can pass down here?

Not all registries support ETags today, and though there's guidance to recommend it, there's currently no way to tell whether the ETag you pass will be honored.

_That being said_, you should still set them optimistically. When/if that ggcr PR lands you should just get it For Free™️, but I don't know at this time whether or when it will land. I'll poke Jon to give his feedback.

wlynch (approved):
Sorry about the delay! Was OOO Mon-Tue 🙏 

/lgtm
/approve

PR #579, opened 2022-09-28
allow reading the pull secret from the podTemplate

joejstuart:
# Changes
This change is to allow chains to get the `imgPullSecret` from the pod template. This will allow users to configure repo auth on a per pipelineRun/taskRun basis and not modify the service account.

### DIFF
--- a/pkg/chains/objects/objects_test.go
+++ b/pkg/chains/objects/objects_test.go
@@ -23,68 +23,70 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const (
-	namespace  = "objects-test"
-	pullSecret = "pull-secret"
-)
-
-var (
-	tr = &v1beta1.TaskRun{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "foo",
-			Namespace: namespace,
-		},
-		Spec: v1beta1.TaskRunSpec{},
-	}
-	pr = &v1beta1.PipelineRun{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "foo",
-			Namespace: namespace,
-		},
-		Spec: v1beta1.PipelineRunSpec{},
-	}
-	templatePullSecret = &pod.PodTemplate{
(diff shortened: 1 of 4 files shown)


lcarva:
@joejstuart, seems like a good enhancement. Would you be willing to add unit tests?

joejstuart:
> @joejstuart, seems like a good enhancement. Would you be willing to add unit tests?

absolutely.

pkg/chains/objects/objects_test.go
```diff
@@ -0,0 +1,129 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	namespace  = "objects-test"
+	pullSecret = "pull-secret"
+)
+
+var (
+	tr = &v1beta1.TaskRun{
```
lcarva, line 32: I think this is problematic because the pointer is shared across multiple unit tests. This can cause unintended consequences.

Can we maybe do away with these variables and just have some factory-like functions that create them as needed? e.g.

```go
...
		{
			name:     "Test pull secret found",
			taskRun:  fakeTaskRunObject(true),
			want:     []string{pullSecret},
		}
...


func fakeTaskRunObject(includePullSecret bool) *v1beta1.TaskRun {
  tr := &v1beta1.TaskRun{
	ObjectMeta: metav1.ObjectMeta{
			Name:      "foo",
			Namespace: namespace,
		},
		Spec: v1beta1.TaskRunSpec{},
	}
  
  if includePullSecret {
    tr.Spec.PodTemplate = &pod.PodTemplate{
		ImagePullSecrets: []corev1.LocalObjectReference{
			{
				Name: pullSecret,
			},
		},
	}
  }

  return tr
}

```

joejstuart:
> LGTM!
> 
> Main nit is we should document this here: https://github.com/tektoncd/chains/blob/main/docs/authentication.md#authenticating-to-an-oci-registry

Docs added. Thanks!

wlynch (approved):
/lgtm

Issue #476, opened 2022-06-29
Set the StorageOpts's key for OCI artifact to be the full identifier


chuangw6:
### Feature request

There is a clear need in Grafeas backend storage to get the full identifier for an OCI artifact, that is `IMAGE_URL@IMAGE_DIGEST`.

Currently, the [`StorageOpts` key](https://github.com/tektoncd/chains/blob/7894a226f1628727ecbcd77fa7a8400b83b22339/pkg/config/options.go#L21-L26) is [set](https://github.com/tektoncd/chains/blob/7894a226f1628727ecbcd77fa7a8400b83b22339/pkg/chains/signing.go#L193) to be just [the first 12 chars of the digest](https://github.com/tektoncd/chains/blob/7894a226f1628727ecbcd77fa7a8400b83b22339/pkg/artifacts/signable.go#L200-L203) for OCI artifact. That means in grafeas backend implementation, we have to do some extra work to loop over all the images generated from a taskrun, and compare the first 12 chars of every single image's digest to the key field of `StorageOpts` in order to find the full representation `IMAGE_URL@IMAGE_DIGEST`. That's not good.

Therefore, it would be great if the `StorageOpts` key is set to be the full representation `IMAGE_URL@IMAGE_DIGEST` for OCI artifact beforehand instead of just first 12 chars of the digest.

### Use case

When creating grafeas occurrences, we need to set the `resourceUri` field to be the full path of an image.

Thanks @wlynch for the idea.

wlynch:
This sounds fine - I'll need to dig in a bit more to see how this is referenced in other places, but refactoring how this is passed around within chains so that Grafeas can have access to the full digest sgtm.

PR #589, opened 2022-10-04
Full access to image url and digest in StorageOpts

chuangw6:
# Changes

Related to https://github.com/tektoncd/chains/issues/476

Prior, the StorageOpts key for OCI artifact was the first 12 characters of the digest.

Now, we set it to be the full representation of the OCI artifact in the format of 
`IMAGE_URL@IMAGE_DIGEST` so that storage backends can have the full 
access to the identifier of the OCI artifact directly.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
StorageOpts have the full access to the image identifier.
```

### DIFF
--- a/pkg/artifacts/signable.go
+++ b/pkg/artifacts/signable.go
@@ -33,7 +33,14 @@ type Signable interface {
 	StorageBackend(cfg config.Config) sets.String
 	Signer(cfg config.Config) string
 	PayloadFormat(cfg config.Config) formats.PayloadType
-	Key(interface{}) string
+	// FullKey returns the full identifier for a signable artifact.
+	// - For OCI artifact, it is the full representation in the format of `<NAME>@sha256:<DIGEST>`.
+	// - For TaskRun/PipelineRun artifact, it is `<GROUP>-<VERSION>-<KIND>-<UID>`
+	FullKey(interface{}) string
+	// ShortKey returns the short version  of an artifact identifier.
+	// - For OCI artifact, it is first 12 chars of the image digest.
+	// - For TaskRun/PipelineRun artifact, it is `<KIND>-<UID>`.
+	ShortKey(interface{}) string
 	Type() string
 	Enabled(cfg config.Config) bool
 }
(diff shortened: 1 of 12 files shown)


pkg/artifacts/signable.go
```diff
@@ -293,7 +293,7 @@ func (oa *OCIArtifact) Signer(cfg config.Config) string {
 
 func (oa *OCIArtifact) Key(obj interface{}) string {
 	v := obj.(name.Digest)
-	return strings.TrimPrefix(v.DigestStr(), "sha256:")[:12]
+	return v.Name()
```
wlynch, line 296: We should double check the behavior here for annotation based storage - we should probably preserve the old behavior there since IIRC labels/annotations have a fixed key size, and including the full digest may cause issues.

wlynch (changes requested):

pkg/artifacts/signable.go
chuangw6: Great point!! Thanks for catching this @wlynch. You are right. I found the [source](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set).

Now, I maintained both ShortKey and FullKey so that downstream code can use whichever is needed. Please let me know what you think or if you have better suggestions.

Thank you!!

pkg/artifacts/signable.go
```diff
@@ -33,7 +33,8 @@ type Signable interface {
 	StorageBackend(cfg config.Config) sets.String
 	Signer(cfg config.Config) string
 	PayloadFormat(cfg config.Config) formats.PayloadType
-	Key(interface{}) string
+	FullKey(interface{}) string
```
wlynch, line 36: bikeshed: I'm okay with Key/ShortKey. I know we've been lax about this, but we should also add godoc describing the difference between the 2 (you can generalize instead of going into the specifics of each impl).

wlynch (changes requested):
Looks good! One minor nit.

pkg/artifacts/signable.go
wlynch: We want docs here as well since this is where Signable implementors / IDEs are going to look for documentation for the type.

chuangw6: Oh I see. Added! Please take a look! Thanks

wlynch (approved):
/lgtm

PR #590, opened 2022-10-05
Grafeas storage for PipelineRun-level provenance

chuangw6:
# Changes

Support storing pipelinerun level provenance in grafeas
backend.

There are 3 main aspects of the PR:
- created different BUILD Notes for storing both taskrun and pipelinerun
level provenance/occurrence in grafeas that attest the same artifact.
- extracted artifact identifiers from intoto subjects instead of calling
individual helpers to extract from results.
- changed TektonObject "GetKind" method to "GetGVK" to get the whole
  GroupVersionKind information.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Support Grafeas storage for PipelineRun level provenance
```

### DIFF
--- a/pkg/chains/formats/intotoite6/extract/extract.go
+++ b/pkg/chains/formats/intotoite6/extract/extract.go
@@ -17,6 +17,7 @@ limitations under the License.
 package extract
 
 import (
+	"fmt"
 	"sort"
 	"strings"
 
@@ -30,8 +31,13 @@ import (
 	"go.uber.org/zap"
 )
 
-// SubjectDigests extracts OCI images and other structured results from the TaskRun and PipelineRun based on standard hinting set up
-// It also goes through looking for any PipelineResources of Image type
+// SubjectDigests returns software artifacts produced from the TaskRun/PipelineRun object
+// in the form of standard subject field of intoto statement.
+// The type hinting fields expected in results help identify the generated software artifacts.
+// Valid type hinting fields must:
+//   - have suffix `IMAGE_URL` & `IMAGE_DIGEST` or `ARTIFACT_URI` & `ARTIFACT_DIGEST` pair.
+//   - the `*_DIGEST` field must be in the format of "<algorithm>:<actual-sha>" where the algorithm must be "sha256" and actual sha must be valid per https://github.com/opencontainers/image-spec/blob/main/descriptor.md#sha-256.
+//   - the `*_URL` or `*_URI` fields cannot be empty.
 func SubjectDigests(obj objects.TektonObject, logger *zap.SugaredLogger) []intoto.Subject {
 	var subjects []intoto.Subject
 
@@ -117,4 +123,20 @@ func SubjectDigests(obj objects.TektonObject, logger *zap.SugaredLogger) []intot
 		return subjects[i].Name <= subjects[j].Name
 	})
 	return subjects
+}
+
+// RetrieveAllArtifactURIs returns all the URIs of the software artifacts produced from the run object.
+// - It first extracts intoto subjects from run object results and converts the subjects
+// to a slice of string URIs in the format of "NAME" + "@" + "ALGORITHM" + ":" + "DIGEST".
+// - If no subjects could be extracted from results, then an empty slice is returned.
+func RetrieveAllArtifactURIs(obj objects.TektonObject, logger *zap.SugaredLogger) []string {
+	result := []string{}
+	subjects := SubjectDigests(obj, logger)
+
+	for _, s := range subjects {
+		for algo, digest := range s.Digest {
+			result = append(result, fmt.Sprintf("%s@%s:%s", s.Name, algo, digest))
+		}
+	}
+	return result
 }
--- a/pkg/chains/formats/intotoite6/extract/extract_test.go
+++ b/pkg/chains/formats/intotoite6/extract/extract_test.go
@@ -1,0 +1,165 @@
+/*
+Copyright 2022 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extract_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	"github.com/tektoncd/chains/pkg/chains/formats/intotoite6/extract"
+	"github.com/tektoncd/chains/pkg/chains/objects"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	logtesting "knative.dev/pkg/logging/testing"
+)
+
+const (
+	artifactURL1    = "gcr.io/test/kaniko-chains1"
+	artifactDigest1 = "a2e500bebfe16cf12fc56316ba72c645e1d29054541dc1ab6c286197434170a9"
+	artifactURL2    = "us-central1-maven.pkg.dev/test/java"
+	artifactDigest2 = "b2e500bebfe16cf12fc56316ba72c645e1d29054541dc1ab6c286197434170a9"
+)
+
+func TestSubjectDigestsAndRetrieveAllArtifactURIs(t *testing.T) {
+	var tests = []struct {
+		name string
+		// a map of url:digest pairs for type hinting results
+		results      map[string]string
+		wantSubjects []intoto.Subject
+		wantFullURLs []string
+	}{
+		{
+			name: "valid type hinting result fields",
+			results: map[string]string{
+				artifactURL1: "sha256:" + artifactDigest1,
+				artifactURL2: "sha256:" + artifactDigest2,
+			},
+			wantSubjects: []intoto.Subject{
+				{
+					Name: artifactURL1,
+					Digest: map[string]string{
+						"sha256": artifactDigest1,
+					},
+				},
+				{
+					Name: artifactURL2,
+					Digest: map[string]string{
+						"sha256": artifactDigest2,
+					},
+				},
+			},
+			wantFullURLs: []string{
+				fmt.Sprintf("%s@sha256:%s", artifactURL1, artifactDigest1),
+				fmt.Sprintf("%s@sha256:%s", artifactURL2, artifactDigest2),
+			},
+		},
+		{
+			name: "invalid/missing digest algorithm name",
+			results: map[string]string{
+				artifactURL1: "sha1:" + artifactDigest1,
+				artifactURL2: artifactDigest2,
+			},
+			wantSubjects: nil,
+			wantFullURLs: []string{},
+		},
+		{
+			name: "invalid digest sha",
+			results: map[string]string{
+				artifactURL1: "sha256:a123",
+			},
+			wantSubjects: nil,
+			wantFullURLs: []string{},
+		},
+		{
+			name: "invalid url value",
+			results: map[string]string{
+				"": "sha256:" + artifactDigest1,
+			},
+			wantSubjects: nil,
+			wantFullURLs: []string{},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			// test both taskrun object and pipelinerun object
+			runObjects := []objects.TektonObject{
+				createTaskRunObjectWithResults(tc.results),
+				createPipelineRunObjectWithResults(tc.results),
+			}
+
+			for _, o := range runObjects {
+				gotSubjects := extract.SubjectDigests(o, logtesting.TestLogger(t))
+				if diff := cmp.Diff(tc.wantSubjects, gotSubjects, cmpopts.SortSlices(func(x, y intoto.Subject) bool { return x.Name < y.Name })); diff != "" {
+					t.Errorf("Wrong subjects extracted, diff=%s", diff)
+				}
+
+				gotURIs := extract.RetrieveAllArtifactURIs(o, logtesting.TestLogger(t))
+				if diff := cmp.Diff(tc.wantFullURLs, gotURIs, cmpopts.SortSlices(func(x, y string) bool { return x < y })); diff != "" {
+					t.Errorf("Wrong URIs extracted, diff=%s", diff)
+				}
+			}
+
+		})
+	}
+}
+
+func createTaskRunObjectWithResults(results map[string]string) objects.TektonObject {
+	trResults := []v1beta1.TaskRunResult{}
+	prefix := 0
+	for url, digest := range results {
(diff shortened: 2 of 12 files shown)


pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -13,486 +13,486 @@ limitations under the License.
...
-		// each statement looks like `resourceUrl="foo"`
-		// 2. extract `foo`
-		uriWithQuotes := strings.Split(statement, "=")[1]
-		uri := uriWithQuotes[1 : len(uriWithQuotes)-1]
-		results = append(results, uri)
-	}
-
-	return results
-}
+// import (
```
bcaton85, line 16: Were tests going to done in a follow up PR?

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -240,13 +260,14 @@ func (b *Backend) createNote(ctx context.Context, tr *v1beta1.TaskRun, opts conf
...
-func (b *Backend) createOccurrence(ctx context.Context, tr *v1beta1.TaskRun, payload []byte, signature string, opts config.StorageOpts) ([]*pb.Occurrence, error) {
+func (b *Backend) createOccurrence(ctx context.Context, obj objects.TektonObject, payload []byte, signature string, opts config.StorageOpts) ([]*pb.Occurrence, error) {
 	occs := []*pb.Occurrence{}
 
 	// create Occurrence_Attestation for OCI
 	if opts.PayloadFormat == formats.PayloadTypeSimpleSigning {
-		uri := b.retrieveSingleOCIURI(tr, opts)
-		occ, err := b.createAttestationOccurrence(ctx, tr, payload, signature, uri)
+		uri := b.retrieveSingleOCIURI(obj, opts)
+		b.logger.Info("chuangw6 - uri for att occ ", uri)
```
bcaton85, line 269: Is this a debugging comment that was accidentally left in?

pkg/chains/storage/grafeas/grafeas.go
```diff
@@ -214,21 +213,42 @@ func (b *Backend) createNote(ctx context.Context, tr *v1beta1.TaskRun, opts conf
...
-	return b.client.CreateNote(ctx,
-		&pb.CreateNoteRequest{
-			Parent: b.getProjectPath(),
-			NoteId: fmt.Sprintf(buildNoteNameFormat, noteID),
-			Note: &pb.Note{
-				ShortDescription: "Build Provenance Note",
-				Type: &pb.Note_Build{
-					Build: &pb.BuildNote{
-						BuilderVersion: tr.GetGroupVersionKind().GroupVersion().String(),
+	switch obj.GetObject().(type) {
```
bcaton85, line 216: Seems the only fields that differ here are `NoteId` and `ShortDescription`, would we want to switch on just those to reduce duplicated code? This is very much a nitpick, i'm also fine with how it is now.

bcaton85 (commented):
looks good, added minor comments. Glad this got worked on. Given i haven't ran this since I don't have a Grafeas backend to test with.

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: Yes! The tests will be done in this PR.

pkg/chains/storage/grafeas/grafeas.go
chuangw6: Yes, thanks @bcaton85!

pkg/chains/storage/grafeas/grafeas.go
chuangw6: Good point. Happy to extract this out. Thanks @bcaton85 for reviewing the PR.

pkg/chains/objects/objects.go
```diff
@@ -61,15 +62,17 @@ type TaskRunObject struct {
 	*v1beta1.TaskRun
 }
 
+var _ TektonObject = &TaskRunObject{}
```
lcarva, line 65: Oh good idea. Thanks for adding this!

pkg/chains/storage/grafeas/grafeas_test.go
chuangw6: The unit tests are finished. Please take a look. I also tested the backend with real grafeas server (container analysis). It works as expected. Thank you!

chuangw6:
This PR is ready for review. Please take a look @wlynch, @jagathprakash, @lcarva when you get a chance. Thank you very much for your time and feedback in advance!

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -294,87 +442,122 @@ func testStoreAndRetrieve(ctx context.Context, t *testing.T, test testConfig, ba
...
-					IntotoStatement: &pb.InTotoStatement{
-						Predicate: &pb.InTotoStatement_SlsaProvenanceZeroTwo{
-							SlsaProvenanceZeroTwo: &pb.SlsaProvenanceZeroTwo{
-								Builder: &pb.SlsaProvenanceZeroTwo_SlsaBuilder{},
-								Invocation: &pb.SlsaProvenanceZeroTwo_SlsaInvocation{
-									ConfigSource: &pb.SlsaProvenanceZeroTwo_SlsaConfigSource{},
+	}
+}
+
+func getCIPipelineRunBUILDOcc(t *testing.T, identifier string) *pb.Occurrence {
```
wlynch, line 509:
```suggestion
func getPipelineRunBuildOcc(t *testing.T, identifier string) *pb.Occurrence {
```

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -408,71 +591,126 @@ type mockGrafeasServer struct {
...
+func (s *mockGrafeasServer) getOccurrencesByFilter(filter string, occurrences []*pb.Occurrence) []*pb.Occurrence {
+	sort.Slice(occurrences, func(i, j int) bool {
+		return occurrences[i].ResourceUri+occurrences[i].NoteName < occurrences[j].ResourceUri+occurrences[j].NoteName
+	})
+
 	// if filter string is empty, the expected behaviour will be to return all.
-	if len(req.GetFilter()) == 0 {
-		return &pb.ListOccurrencesResponse{Occurrences: sortedOccurrencesInServer}, nil
+	if len(filter) == 0 {
+		return occurrences
```
wlynch, line 695: I'd move the sorting back up to ListOccurrences - there's no mention that `getOccurrencesByFilter` is also sorting, and this might cause issues if this is removed / refactored later if the next change doesn't realize this did more than just filter.

wlynch (changes requested):
Mostly LGTM!

Just some small tweaks.

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -234,34 +371,45 @@ func TestGrafeasBackend_StoreAndRetrieve(t *testing.T) {
 	if err != nil {
 		t.Fatal("Failed to call ListOccurrences. error ", err)
 	}
-	wantedResponse := &pb.ListOccurrencesResponse{Occurrences: getExpectedOccurrences()}
-	if diff := cmp.Diff(got, wantedResponse, protocmp.Transform()); diff != "" {
+
+	sort.Slice(allOccurrencesInServer, func(i, j int) bool {
```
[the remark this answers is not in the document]
chuangw6: Yeah I think it's doable. But the question here is what we want to test here.
Currently, it's testing if all occurrences from different requests are created correctly in the server i.e. testing if two occurrences with same resourceUri (taskrun level provenance and pipelinerun level provenance) are indeed created in the server. 
IIUC, what you described seems more like testing if the occurrences in one request are created correctly, but not testing the potential conflicts caused by storing multiple occurrences from multiple requests.

Please let me know what you think. I am happy to change to the second case too!

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -294,87 +442,122 @@ func testStoreAndRetrieve(ctx context.Context, t *testing.T, test testConfig, ba
...
-					IntotoStatement: &pb.InTotoStatement{
-						Predicate: &pb.InTotoStatement_SlsaProvenanceZeroTwo{
-							SlsaProvenanceZeroTwo: &pb.SlsaProvenanceZeroTwo{
-								Builder: &pb.SlsaProvenanceZeroTwo_SlsaBuilder{},
-								Invocation: &pb.SlsaProvenanceZeroTwo_SlsaInvocation{
-									ConfigSource: &pb.SlsaProvenanceZeroTwo_SlsaConfigSource{},
-								},
+// ------------------ occurrences for taskruns and pipelineruns --------------
+// BUILD Occurrence for the build taskrun that stores the slsa provenance
+func getBuildTaskRunBUILDOcc(t *testing.T, identifier string) *pb.Occurrence {
```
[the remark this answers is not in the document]
chuangw6: Done!

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -15,41 +15,170 @@ package grafeas
...
 	opts      config.StorageOpts
 }
 
 type testConfig struct {
-	name    string
-	args    args
-	wantErr bool
+	name              string
+	args              args
+	wantedOccurrences []*pb.Occurrence
```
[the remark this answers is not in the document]
chuangw6: Good catch!

pkg/chains/storage/grafeas/grafeas_test.go
```diff
@@ -15,41 +15,170 @@ package grafeas
...
 	gstatus "google.golang.org/grpc/status"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	logtesting "knative.dev/pkg/logging/testing"
 	rtesting "knative.dev/pkg/reconciler/testing"
 )
 
+const (
+	PROJECTID = "test-project"
+	NOTEID    = "test-note"
```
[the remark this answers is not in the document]
chuangw6: Thanks for catching this. This is helpful!

wlynch (commented):
Mostly LGTM! One last question

pkg/chains/formats/intotoite6/extract/extract.go
```diff
@@ -105,3 +106,15 @@ func SubjectDigests(obj objects.TektonObject, logger *zap.SugaredLogger) []intot
...
 	return subjects
 }
+
+// SubjectsToURIs converts intoto subjects to a slice of string URIs in the format
+// of "NAME" + "@" + "ALGORITHM" + ":" + "DIGEST"
+func SubjectsToURIs(subjects []intoto.Subject) []string {
+	result := []string{}
+	for _, s := range subjects {
+		for algo, digest := range s.Digest {
+			result = append(result, fmt.Sprintf("%s@%s:%s", s.Name, algo, digest))
```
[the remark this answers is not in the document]
chuangw6: Happy to add the tests!
Regarding whether return error or empty slice, my thought is that if the intoto statement has empty subject or subject field is missing, it just indicates the build does not produce any artifact/output rather than a conversion error. As such we may just want to return empty slice as well? wdyt?
When it comes storage, 
- grafeas will not create any occurrences for the statement with empty subjects, and just log the information saying no occurrences were created in the controller log.
- [OCI storage](https://github.com/tektoncd/chains/blob/43615bdd7efbae5d7245d8f22bba6718f0ecc3bd/pkg/chains/storage/oci/oci.go#L94-L106) seems to have similar behaviour. And we may want to refactor OCI storage later to use this helper function as well.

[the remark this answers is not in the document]
chuangw6: [tests](https://github.com/tektoncd/chains/pull/590/files#diff-852530124ab21d01f48bb7e3a76cb200a0c23d5e876844fb45dec65c044530df) are added. PTAL. Thank you! @wlynch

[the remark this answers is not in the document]
chuangw6: To make the functionality reusable of being able to extract artifact URIs, I moved the `RetrieveAllArtifactURIs` function from grafeas package to this extract package so that it can be reused in oci or any other places as needed. The [tests](https://github.com/tektoncd/chains/pull/590/files#diff-852530124ab21d01f48bb7e3a76cb200a0c23d5e876844fb45dec65c044530df) are added.

PTAL. Thank you @wlynch !!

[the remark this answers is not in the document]
wlynch: 👍  empty slice is fine with me - make sure the behavior is documented (e.g. what's considered valid)!

pkg/chains/formats/intotoite6/extract/extract_test.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+	)
+)
+
+func TestSubjectDigests(t *testing.T) {
+	var tests = []struct {
+		name         string
+		runObj       objects.TektonObject
+		wantSubjects []intoto.Subject
+	}{
```
wlynch, line 75: I'd add some test cases where there are malformed / missing digests as well. e.g.

- Result with name but no result value
- sha1 (or any other non-sha256)
- Hash without `sha256:` prefix
- whatever else you can think of!

wlynch (commented):
Almost there! Just some missing test coverage.

pkg/chains/formats/intotoite6/extract/extract_test.go
chuangw6: Good points. I've added all of them including some other edge cases i.e. typo in type hinting field name, invalid sha256 value etc. PTAL! Thanks!

chuangw6:
/assign @wlynch 

added you as assignee. Thank you for the thorough review!!

pkg/chains/formats/intotoite6/extract/extract_test.go
```diff
@@ -0,0 +1,284 @@
...
+	)
+
+	prInvalidTypeHintingValue = objects.NewPipelineRunObject(
+		&v1beta1.PipelineRun{
+			Status: v1beta1.PipelineRunStatus{
+				PipelineRunStatusFields: v1beta1.PipelineRunStatusFields{
+					PipelineResults: []v1beta1.PipelineRunResult{
+						// invalid digest algorithm name
+						{Name: "invalidDigestAlgo_IMAGE_DIGEST", Value: *v1beta1.NewStructuredValues("sha1:" + artifactDigest2)},
+						{Name: "invalidDigestAlgo_IMAGE_URL", Value: *v1beta1.NewStructuredValues(artifactURL2)},
```
wlynch, line 137: We may have went too far in the other direction with too many tests 😅 

i.e. what's the difference between these?

```go
{Name: "invalidDigestAlgo_IMAGE_URL", Value: *v1beta1.NewStructuredValues(artifactURL2)},
{Name: "invalidDigestAlgo_ARTIFACT_URI", Value: *v1beta1.NewStructuredValues(artifactURL2)},
{Name: "invalidDigestSHA_IMAGE_URL", Value: *v1beta1.NewStructuredValues(artifactURL2)},
{Name: "invalidDigestSHA_ARTIFACT_URI", Value: *v1beta1.NewStructuredValues(artifactURL2)},
...
```

On the surface they all look like the same test case with different names - can we cut this down so we're only testing each thing once?

chuangw6: Sorry for the confusion. Since uri and digest are expected to appear in pair, I tried to cover the case that no subjects are extracted when the digest part is invalid even if the corresponding url in the pair meets the requirement 🤣.

wlynch (approved):
/lgtm

PR #599, opened 2022-10-20
Support structured results for PipelineRunObject

### LINKED ISSUE #595: PipelineRunObject adding supported to structured results
### Feature request

Adding support of structured results in PipelineRunObject:
Currently TaskRunObject already have the structured results [being supported](https://github.com/tektoncd/chains/blob/780521656ab0ea2a72d7a969a13040848ac6b969/pkg/chains/objects/objects.go#L98-L99) but not PipelineRunObject. It seems straightforward to add those, just curious if there's anything concerning for adding those? @bcaton85 @lcarva fyi.

### Use case

Use Objects for structured results in PipelineRunObjects

ywluogg:
# Changes

Support structured results for PipelineRunObject. Previously PipelineRunObject only supports string types

Fixes #595 

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Support structured results for PipelineRunObject.
```

### DIFF
--- a/pkg/artifacts/signable.go
+++ b/pkg/artifacts/signable.go
@@ -266,6 +266,10 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
 
 	for _, res := range obj.GetResults() {
 		if strings.HasSuffix(res.Name, identifierSuffix) {
+			if res.Value.StringVal == "" {
+				logger.Debugf("error getting string value for %s", res.Name)
+				continue
+			}
 			marker := strings.TrimSuffix(res.Name, identifierSuffix)
 			if v, ok := ss[marker]; ok {
 				v.URI = strings.TrimSpace(res.Value.StringVal)
@@ -276,6 +280,10 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
 			// TODO: add logic for Intoto signable target as input.
 		}
 		if strings.HasSuffix(res.Name, digestSuffix) {
+			if res.Value.StringVal == "" {
+				logger.Debugf("error getting string value for %s", res.Name)
+				continue
+			}
 			marker := strings.TrimSuffix(res.Name, digestSuffix)
 			if v, ok := ss[marker]; ok {
 				v.Digest = strings.TrimSpace(res.Value.StringVal)
@@ -319,23 +327,11 @@ func ExtractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker
 
 	// TODO(#592): support structured results using Run
 	results := []objects.Result{}
-	tr, isTaskRun := obj.GetObject().(*v1beta1.TaskRun)
-	if isTaskRun {
-		for _, res := range tr.Status.TaskRunResults {
-			results = append(results, objects.Result{
-				Name:  res.Name,
-				Value: res.Value,
-			})
-		}
-
-	} else {
-		pr := obj.GetObject().(*v1beta1.PipelineRun)
-		for _, res := range pr.Status.PipelineResults {
-			results = append(results, objects.Result{
-				Name:  res.Name,
-				Value: res.Value,
-			})
-		}
+	for _, res := range obj.GetResults() {
+		results = append(results, objects.Result{
+			Name:  res.Name,
+			Value: res.Value,
+		})
 	}
 	for _, res := range results {
 		if strings.HasSuffix(res.Name, categoryMarker) {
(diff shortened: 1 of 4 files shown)


wlynch (commented):
Looks good! Can we add some tests? 🙏

wlynch:
@bcaton85 following up on https://github.com/tektoncd/chains/issues/595#issuecomment-1282582145

>  IIRC there is places where the pipelinerun results is expected to be string.

Have a link for where this is happening? 👀

bcaton85:
This happens in two places, when trying to extract values from results:
https://github.com/tektoncd/chains/blob/main/pkg/artifacts/signable.go#L220
https://github.com/tektoncd/chains/blob/main/pkg/artifacts/signable.go#L267-L287

ywluogg:
> > Have a link for where this is happening? 👀
> 
> @wlynch This happens in two places, when trying to extract values from results: https://github.com/tektoncd/chains/blob/main/pkg/artifacts/signable.go#L220 https://github.com/tektoncd/chains/blob/main/pkg/artifacts/signable.go#L267-L287

I added a few more checks for the string values

ywluogg:
> Looks good! Can we add some tests? 🙏

Added those tests and a few more checks for the string values

bcaton85:
I don't have access to approve or merge but LGTM

pkg/artifacts/signable.go
```diff
@@ -268,18 +272,34 @@ func extractTargetFromResults(obj objects.TektonObject, identifierSuffix string,
...
+				}
 				v.Digest = strings.TrimSpace(res.Value.StringVal)
 			} else {
+				if res.Value.StringVal == "" {
+					logger.Errorf("error getting string value for %s", res.Name)
+					continue
+				}
 				ss[marker] = &StructuredSignable{Digest: strings.TrimSpace(res.Value.StringVal)}
 			}
 		}
```
chuangw6, line 305: nit: Seems like we only accept string type result here right? If so, is it possible that we move the `if res.Value.StringVal == "" {}` block right after the line `for _, res := range obj.GetResults() {` so that we don't repeat the check?

chuangw6 (commented):
Thanks @ywluogg! Mostly LGTM! Just a minor comment.

pkg/artifacts/signable.go
ywluogg: That's a good point! My only concern is that we probably don't want to log this when the results don't have the right suffices. I can check the StringVal at where you mention, but don't log at all and keep continuing, but I definitely feel users would find the logged error being helpful

wlynch: That sgtm! If you want to keep the log in, perhaps downgrade to log.Debug? This definitely isn't log.Error worthy since it isn't really an error and it's not particularly actionable. log.Debug will remove it from production logs, but will still keep it around for local debugging, etc.

chuangw6: Got it. +1 to log this error.

In that case, maybe we can just check the result type only after checking suffix each time instead of checking result type in both inner `if v, ok := ss[marker]; ok {` and its else block?

ywluogg: That's a good point! Change is done!

ywluogg:
> /lgtm

Thanks Chuang!

pkg/chains/objects/objects_test.go
```diff
@@ -127,3 +156,43 @@ func TestPipelineRun_ImagePullSecrets(t *testing.T) {
 		})
 	}
 }
+
+func TestPipelineRun_GetResults(t *testing.T) {
+
+	t.Run("TestPipelineRun_GetResults", func(t *testing.T) {
```
wlynch, line 162: The t.Run is redundant here - you can remove it

```suggestion
```

pkg/artifacts/signable.go
```diff
@@ -217,6 +217,10 @@ func ExtractOCIImagesFromResults(obj objects.TektonObject, logger *zap.SugaredLo
 		if key.Name != "IMAGES" {
 			continue
 		}
+		if key.Value.StringVal == "" {
```
wlynch, line 220: Can we add some tests to signable_test.go to verify behavior for the different types of result formats?

I suspect we might not even need this check at all because of the `if trimmed == 
""` below, but it would be great to have tests to confirm.

wlynch (changes requested):

pkg/artifacts/signable.go
ywluogg: Good point! I added the tests and removed this check

wlynch (approved):
/lgtm

ywluogg:
@chuangw6 Hi Chuang, do you mind giving another LGTM again? Thanks!

Issue #602, opened 2022-10-28
Chains can stop PipelineRun's from being deleted if the TaskRun has been deleted


stuartwdouglas:
# Expected Behavior

If a TaskRun has been removed it should still be possible to remove a PipelineRun.

# Actual Behavior

The Tekton Chains controller sees the TaskRun is missing, and returns an error without removing the finaliser. This means it is now impossible to delete the PipelineRun without manually removing the finaliser.  

# Steps to Reproduce the Problem

1. Create a PipelineRun
2. Delete a child TaskRun before the PR is signed
3. Attempt to delete the PipelineRun

- Kubernetes version:

  ```
  Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0+dc5a2fd", GitCommit:"0a57f1f59bda75ea2cf13d9f3b4ac5d202134f2d", GitTreeState:"clean", BuildDate:"2022-09-29T16:45:01Z", GoVersion:"go1.18.4", Compiler:"gc", Platform:"linux/amd64"}

  ```

- Tekton Pipeline version:
Client version: 0.27.0
Chains version: v0.9.0
Pipeline version: v0.37.4
Triggers version: v0.20.2
Operator version: v0.60.1

lcarva:
I'm surprised by this behavior in Chains v0.9.0. That version of Chains doesn't watch PipelineRun resources. For my own sanity, can you verify that it's using the "official" Chains image?

The next release of Chains, likely v0.13.0, does include support for PipelineRun attestations so this may be an issue there. 

https://github.com/tektoncd/chains/issues/560 seems related and could be a potential fix.

stuartwdouglas:
Actually I just realised I left out a critical piece of info and just blindly filled out the version field as per the instructions(sorry, it had been a long day).

Pipeline service is actually using https://github.com/hacbs-contract/chains/tree/poc-tep-84


That said from what I could see the problem was here:

https://github.com/tektoncd/chains/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L97

Because the TR is missing the Sign method never gets called, and the finaliser is never removed.

lcarva:
Update: I can reproduce with the following:

# Scenario 4

Stop Chains controller.
Start PR.
Wait for PR to finish.
Delete one of the TRs from the PR.
Start the Chains controller.

I see the error message:
> Unable to get reconciled status of taskrun...

And I'm unable to delete the PR.

PR #607, opened 2022-11-01
Stop reconciliation on unrecoverable error

lcarva:
Fixes: #602

# Changes

When reconciling a PipelineRun, if one of the TaskRuns is not found, do not retry reconciling the resource later.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -23,6 +23,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	pipelinerunreconciler "github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun"
 	listers "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"knative.dev/pkg/logging"
 	pkgreconciler "knative.dev/pkg/reconciler"
 	"knative.dev/pkg/tracker"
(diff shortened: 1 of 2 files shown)


lcarva:
@wlynch , wdyt of this approach?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -95,6 +96,11 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		tr, err := r.TaskRunLister.TaskRuns(pr.Namespace).Get(name)
 		if err != nil {
 			logging.FromContext(ctx).Errorf("Unable to get reconciled status of taskrun %s within pipelinerun", name)
+			if errors.IsNotFound(err) {
```
wlynch, line 99: Can we add a test to make sure that we don't try and write an attestation for a PipelineRun that has a missing TaskRun?

We probably should just drop the PipelineRun and refuse to sign it.

wlynch (changes requested):
Change LGTM, but a test would be great 🙏

pkg/reconciler/pipelinerun/pipelinerun.go
lcarva: Thanks. Added the test. Let me know what you think.

wlynch: Looks great! Thanks.

wlynch (approved):
/lgtm

lcarva:
@wlynch, this is passing now, but since a rebase was needed your lgtm label got removed. :pray:

wlynch (approved):
/lgtm

PR #608, opened 2022-11-01
Pin PipelineRun e2e bash image.

wlynch:
# Changes

This is currently breaking because it's pulling in a newer version of bash, causing the templated the checksums to fail.

Long term we should look to pull the image digests from the TaskRun status via the PipelineRun status.

Using distroless.dev/busybox instead of DockerHub busybox, since there appears to be some issue using busybox as is (Pipelines refuses to mount the script when executing).

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


priyawadhwa (approved):
Thanks for fixing this!

PR #629, opened 2022-11-28
Remove `predicate.buildConfig.tasks[i].ref` field

chuangw6:
# Changes

Related to https://github.com/tektoncd/chains/pull/436#discussion_r973143588.

Prior, pipeline level provenance had a `ref` field in the `predicate.buildConfig.tasks[i]` section.

In this commit, we remove this `ref` field from buildConfig in favor of `buildConfig.tasks[i].invocation.configSource` https://github.com/tektoncd/chains/pull/554 because configsource should be enough to record the source information that we need to track the origin of the remote definitions.

Since `predicate.buildConfig` is pretty much free text, this deletion shouldn't introduce backward incompatible changes.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Remove the `ref` field from free text field `predicate.buildConfig.tasks[i]` in the Pipeline level provenance
```

### DIFF
--- a/pkg/chains/formats/intotoite6/pipelinerun/pipelinerun.go
+++ b/pkg/chains/formats/intotoite6/pipelinerun/pipelinerun.go
@@ -140,10 +140,6 @@ func buildConfig(pro *objects.PipelineRunObject, logger *zap.SugaredLogger) Buil
 			Results:    tr.Status.TaskRunResults,
 		}
 
-		if t.TaskRef != nil {
-			task.Ref = *t.TaskRef
-		}
-
 		tasks = append(tasks, task)
 		if i < len(pSpec.Tasks) {
 			last = task.Name
--- a/pkg/chains/formats/intotoite6/intotoite6_test.go
+++ b/pkg/chains/formats/intotoite6/intotoite6_test.go
@@ -172,12 +172,8 @@ func TestPipelineRunCreatePayload(t *testing.T) {
 			BuildConfig: pipelinerun.BuildConfig{
 				Tasks: []pipelinerun.TaskAttestation{
 					{
-						Name:  "git-clone",
-						After: nil,
-						Ref: v1beta1.TaskRef{
-							Name: "git-clone",
-							Kind: "ClusterTask",
-						},
+						Name:       "git-clone",
+						After:      nil,
 						StartedOn:  e1BuildStart,
 						FinishedOn: e1BuildFinished,
 						Status:     "Succeeded",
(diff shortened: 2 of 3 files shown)


pkg/chains/formats/intotoite6/pipelinerun/pipelinerun.go
```diff
@@ -140,10 +140,6 @@ func buildConfig(pro *objects.PipelineRunObject, logger *zap.SugaredLogger) Buil
 			Results:    tr.Status.TaskRunResults,
 		}
 
-		if t.TaskRef != nil {
-			task.Ref = *t.TaskRef
-		}
-
```
chuangw6, line 146: Hey @bcaton85 @lcarva ,
This is related to our previous discussion https://github.com/tektoncd/chains/pull/436#discussion_r973143588. Please take a look and let me know what you think. Hope this is not a backward incompatible change.

lcarva: How do I determine which Tasks were used from the PipelineRun attestation? This is a pretty important piece of information that we need.

chuangw6: `buildConfig.tasks[i].invocation.configSource` should be able to tell that b/c it includes fields like the uri, digest and entrypoint. For example, [the configsource for in-cluster task](https://github.com/tektoncd/pipeline/pull/5687#issue-1423152223) will be the api path which has the information like namespace, kind, resource name and uid.

wlynch: Is there any disadvantage to leaving this field in?

I'm tempted to leave this as-is:

1. For compatibility (we would probably want a more phased deprecation if we wanted to remove it completely)
2. I feel like there's some additional value in keeping the value the user gave (e.g. user said to checkout `main`, which was resolved to `34f4b3`) even if it doesn't really give us

chuangw6: Thank you for sharing your thoughts @wlynch @lcarva. Sorry I didn't know there is a policy looking at this field. @lcarva 

> Is there any disadvantage to leaving this field in?

I don't think there is disadvantage to leave this field in other than sort of duplication with configsource. Keeping this for now SGTM.

chuangw6: I'll close this PR for now. We can come back when we need it.

PR #644, opened 2022-12-13
Added GHA to run go vet and go fmt checks.

bendory:
# Changes

Added GHA to run go vet and go fmt checks.

# Submitter Checklist

- [N/A] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [N/A] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
  functionality, content, code)
- [N/A] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [N/A] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/.github/workflows/goclean.yml
+++ b/.github/workflows/goclean.yml
@@ -1,0 +1,38 @@
+name: goclean
+
+on:
+  push:
+    paths:
+      - '**.go'
+      - 'go.mod'
+      - 'go.sum'
+    branches: [ main ]
+  pull_request:
+    paths:
+      - '**.go'
+      - 'go.mod'
+      - 'go.sum'
+    branches: [ main ]
+  schedule:
+    - cron: '41 20 * * 1'
+
+jobs:
+  goclean:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3 # check-out repository
+    
+    - name: Setup Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: '1.18.0'
+    
+    - name: go vet
+      run: |
+        # Use in-place vendored dependencies.
+        go vet -mod=vendor ./...
+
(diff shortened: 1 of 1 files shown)


bendory:
Happily, `go vet` and `go fmt` are clean in this repo. That was not the case over in pipelines, so I'm adding this GHA here too to prevent us from losing integrity down the road.

.github/workflows/goclean.yml
```diff
@@ -0,0 +1,38 @@
...
+      - 'go.mod'
+      - 'go.sum'
+    branches: [ main ]
+  pull_request:
+    paths:
+      - '**.go'
+      - 'go.mod'
+      - 'go.sum'
+    branches: [ main ]
+  schedule:
```
lcarva, line 16: Shouldn't `on.push` make this irrelevant?

bendory: I pondered that too, and I **think** so? The one corner case I can think of is that a change in the go version may cause new warnings to materialize, and it's best to catch those up front rather than in an arbitrary unrelated PR.

But perhaps that corner case is best handled with a simple comment in the file reminding anyone who revs the `go-version` to ensure `vet` and `fmt` are clean prior to commit?

WDYT?

lcarva: I didn't consider that. I think it's "cheap" enough to run the workflow on a schedule. If it becomes an issue for whatever reason, we can always disable it.

lcarva (approved):
/lgtm

.github/workflows/goclean.yml
bendory: SGTM

PR #655, opened 2022-12-20
Document release compatibility policy.

wlynch:
# Changes

This defines the compatibility policy for chains as we work towards a beta.

This policy is notably different than other tekton projects (i.e. pipelines) - it's aimed to be a simplified version that better suits Chains development and release cycles.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/releases.md
+++ b/releases.md
@@ -38,6 +38,65 @@ Further documentation available:
 - [Installing Tekton][tekton-installation]
 - Standard for [release notes][release-notes-standards]
 
+## Backwards Incompatible Changes
+
+In general we aim for new features to be added in a backwards compatible way,
+but sometimes we will need to make breaking changes. This policy outlines how we
+will make and communicate these.
+
+**NOTE**: Tekton Chains is working towards a formal beta release. Until then,
+all features are technically considered `alpha` (though we continue to do our
+best to retain backwards compatibility).
+
+### Stability levels
+
+| Stability Level | Deprecation Window                                       |
+| --------------- | -------------------------------------------------------- |
+| alpha           | none                                                     |
+| beta            | 3 months or 3 minor releases (which ever is longer)      |
+| stable          | 12 months or 3 LTS minor releases (which ever is longer) |
+
+Example: if a beta feature is announced as deprecated in v0.10.0, it can be
+completely removed in v0.13.0.
+
+### What's in scope
+
+- [User Configuration](docs/config.md)
+- Input mechanisms from Pipelines (e.g.
+  [Type Hinting](https://github.com/tektoncd/chains/blob/main/docs/config.md#chains-type-hinting))
+- Chains generated provenance output formats
+- Exported Client Libraries (https://pkg.go.dev/github.com/tektoncd/chains)
+
+<!-- TODO(wlynch): if/when config type aliases are a thing, document policy around how these can change -->
+
+#### Exceptions
+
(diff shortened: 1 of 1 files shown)


lcarva (approved):
/lgtm

releases.md
```diff
@@ -38,6 +38,65 @@ Further documentation available:
...
+<!-- TODO(wlynch): if/when config type aliases are a thing, document policy around how these can change -->
+
+#### Exceptions
+
+We reserve the right to make breaking changes (regardless of stability level)
+under certain conditions:
+
+- Security
+
+  We will make breaking changes for security reasons if necessary.
```
pritidesai, line 79: Suggest adding CVE or any other vulnerabilities to be a little more specific in addition to `security`.

releases.md
```diff
@@ -38,6 +38,65 @@ Further documentation available:
...
+- Backwards incompatible changes from dependencies
+
+  We try to keep Chains up-to-date with minor versions of its dependencies with
+  Dependabot. Chains assumes its dependencies adhere to
+  [Go module versioning](https://go.dev/doc/modules/developing#versioning) for
+  backwards compatibility. In cases where dependencies break this expectation,
+  we will try and work around these changes as best as we can and make breaking
+  changes as a last resort.
+
+  Example: If tektoncd/pipeline makes a breaking change to its client that
```
pritidesai, line 90: This has happened a few times with Go client but the changes did not intend to introduce a backward incompatibility for the consumers. It's reasonable to document this here just in case. With Pipelines V1, the expectation is not to experience this kind of breaking change outside of the pipelines API policy.

releases.md
```diff
@@ -38,6 +38,65 @@ Further documentation available:
...
+**NOTE**: Tekton Chains is working towards a formal beta release. Until then,
+all features are technically considered `alpha` (though we continue to do our
+best to retain backwards compatibility).
+
+### Stability levels
+
+| Stability Level | Deprecation Window                                       |
+| --------------- | -------------------------------------------------------- |
+| alpha           | none                                                     |
+| beta            | 3 months or 3 minor releases (which ever is longer)      |
```
pritidesai, line 56: How often are we creating minor releases?

Doesn't this sound a little aggressive for `beta`? At least increase the deprecation window to 6 months.

wlynch: > How often are we creating minor releases?

We aim for ~monthly releases.

> Doesn't this sound a little aggressive for beta? At least increase the deprecation window to 6 months.

A quarter heads up feels reasonable to me as a heads up that some thing not in stable is changing. This is also a minimum window (i.e. chains has technically not had a deprecation policy and been under alpha, but we've rarely made breaking changes). We can re-evaluate this in the future if needed - extending a deprecation window is a backwards compatible change.

releases.md
wlynch: Done.

releases.md
wlynch: The [pipelines compatibility policy calls out that breaking changes to language clients are not in scope](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#go-libraries-compatibility-policy) and we've been burned by this a few times.  😢 

I'd love for this to change (https://github.com/tektoncd/pipeline/issues/2748), but I don't think this has really gotten traction on the pipelines side.

lcarva (approved):
/lgtm

Issue #642, opened 2022-12-08
`predicate.materials` needs to record more information.


chuangw6:
# Expected Behavior

`predicate.materials` should record all the artifacts that influence the build according to [slsa definition](https://slsa.dev/provenance/v0.2#materials)

> The collection of artifacts that influenced the build including sources, dependencies, build tools, ***base images***, and so on.
> This is considered to be incomplete unless metadata.completeness.materials is true.

In tekton, this means we should record the following information in materials (feel free to suggest more if missed)
- [x] the source code repo from which artifacts are built.
- [x] step base image
- [x] sidecar base image
- [x] remote tasks referenced in a remote pipeline.yaml (this artifact can be a repo, a Tekton OCI bundle or a hub resource).

# Actual Behavior

[predicate.materials](https://slsa.dev/provenance/v0.2#materials) now only records the source code repo.

wlynch:
SGTM!

lcarva:
@chuangw6, that makes sense! I do think there's value in knowing which image was used in each step of each task because it allows for very targeted policies. Maybe we can add to ``predicate.materials` without changing `predicate.buildConfig` ?

chuangw6:
> Maybe we can add to predicate.materials without changing `predicate.buildConfig` ?

That SGTM. 

I think that the idea of [the plan](https://github.com/tektoncd/chains/issues/597#issuecomment-1282918607) for the configurable formatting / aliasing  will  
- make sure `intoto` / `slsav0.1` option still keeps the current `buildConfig`.
- introduce `slsav0.2` option that will generate the provenance whose `buildConfig` is the full spec.

chuangw6:
cc @chitrangpatel

chuangw6:
Other than the step base images, there are few more things that should be captured in materials.
1. source of remote tasks in a remote pipeline.yaml
In the special case of a remote pipeline definition referencing remote tasks that came from other places i.e. other repos, other tekton bundles, I just realized that those remote tasks' source information should be the build materials as well while the source information of the remote pipeline definition is captured in `predicate.invocation.configSource`. 
2. sidecar base images 
Pointed out by @chitrangpatel. Similar to step base images, sidecar base images should also be captured in `predicate.materials`. 


Please comment if there is any questions. Otherwise, I'll update the issue description to reflect this later. Thanks!!

chitrangpatel:
[...]

PR #658, opened 2022-12-22
Issue 642: Add sidecar, step and remote tasks to pipelinerun materials

chitrangpatel:
This PR addresses issue https://github.com/tektoncd/chains/issues/642 by adding steps and sidecars image uri and digest information to predicate.materials for a PipelineRun.

Prior to reviewing this PR, review #649 
# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
predicate.materials includes image uri and digest information from all steps, sidecars and also config source information for remote tasks. 
```
/kind bug

### DIFF
--- a/pkg/chains/formats/intotoite6/pipelinerun/pipelinerun.go
+++ b/pkg/chains/formats/intotoite6/pipelinerun/pipelinerun.go
@@ -21,6 +21,8 @@ import (
 	"github.com/tektoncd/chains/pkg/artifacts"
 	"github.com/tektoncd/chains/pkg/chains/formats/intotoite6/attest"
 	"github.com/tektoncd/chains/pkg/chains/formats/intotoite6/extract"
+	"github.com/tektoncd/chains/pkg/chains/formats/intotoite6/internal/material"
+	"github.com/tektoncd/chains/pkg/chains/formats/intotoite6/taskrun"
 	"github.com/tektoncd/chains/pkg/chains/objects"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"go.uber.org/zap"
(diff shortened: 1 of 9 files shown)


pkg/chains/formats/intotoite6/pipelinerun/pipelinerun.go
```diff
@@ -182,8 +188,39 @@ func metadata(pro *objects.PipelineRunObject) *slsa.ProvenanceMetadata {
...
+				return mats, nil
+			}
+
+			// add sidecar images
+			if err := taskrun.AddSidecarImagesToMaterials(tr.Status.Sidecars, &mats); err != nil {
+				return mats, nil
+			}
+
+			// add remote task configsource information in materials
+			if tr.Status.Provenance != nil && tr.Status.Provenance.ConfigSource != nil {
```
lcarva, line 215: Should we also add to materials the Provenance info for the PipelineRun itself?

chitrangpatel: you mean the config source info of the pipeline run?

lcarva: Right. My reasoning is that if the Pipeline definition comes from a Tekton Bundle, for example, then that image is yet another artifact that influenced the build. Wdyt?

chitrangpatel: We can include it here. Though that information would be captured in configSource as well. If we are ok with that duplication then I'm happy to include it here.

lcarva: I think duplication is ok. In fact, the image references from the TaskRun status are already included in the provenance under buildConfig, as the discussion in https://github.com/tektoncd/chains/issues/642 shows. To be honest, I'm not sure what the right call is in this case.

chitrangpatel: SGTM! I will include it here.

chitrangpatel: Done!

lcarva (approved):
/lgtm

Issue #636, opened 2022-12-06
`ARTIFACT_INPUTS` type hinting cannot work with digest "sha1:xxxx"


chuangw6:
# Expected Behavior

`ARTIFACT_INPUTS` type hinting works for git uri and digest for which algorithm is "sha1".

# Actual Behavior

`ARTIFACT_INPUTS` type hinting cannot recognize git uri and digest because algorithm is not "sha256" and the sha1 hex value cannot pass the validation as well.

# Background
We recently introduced new type hinting approach using structure type result . For example, to identify the source repo information i.e. uri and digest that will be recorded in `predicate.materials` section, Chains looks for an object result that has suffix `ARTIFACT_INPUTS` with two keys `uri` and `digest`. 

For git sources, the digest algorithm will be `sha1`. That implies the value for the `digest` key of `ARTIFACT_INPUTS` has to be "sha1:xxxxx". But Chains has digest validation and the validation is mainly for sha256 digest i.e. length, and the algorithm has to be "sha256". "sha1" will not be recognized.

# Steps to Reproduce the Problem

1. enable alpha feature flag in pipeline to use object result
2. generate a key pair https://github.com/tektoncd/chains/blob/main/docs/signing.md#generate-cosign-keypair
3. apply chains configmap
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: chains-config
  namespace: tekton-chains
  labels:
    app.kubernetes.io/instance: default
    app.kubernetes.io/part-of: tekton-chains
data:
  artifacts.taskrun.storage: ""
  artifacts.oci.storage: ""
  artifacts.pipelinerun.format: in-toto
  artifacts.pipelinerun.storage: tekton
```
3. run the following yaml
```yaml
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: clone-pipeline
spec:
  workspaces:
    - name: shared
  params:
    - name: git-repo
      description: The source repo to clone.
    - name: git-revision
      description: The revision of the source repo.
  results:
    - name: ARTIFACT_INPUTS
      type: object
      value:
        uri: $(tasks.clone-source.results.url)
        digest: sha1:$(tasks.clone-source.results.commit)
  tasks:
    - name: clone-source
      taskRef:
        resolver: bundles
        params:
          - name: bundle
            value: gcr.io/tekton-releases/catalog/upstream/git-clone:0.7
          - name: name
            value: git-clone
          - name: kind
            value: task
      params:
        - name: url
[...]

PR #676, opened 2023-01-12
Fix checks to allow passing in of sha1 digests through StructuredResults.

jagathprakash:
Fixes https://github.com/tektoncd/chains/issues/636

This PR addresses the issue of artifacts which have sha1 digest not being added into materials.

Before this fix checkDigest would only check if the digest was sha256 and would reject other digests.
This commit fixes checkDigest to ensue that it accepts all valid digests of type sha256, sha512, sha384 and sha1.

Signed-off-by: jagathprakash <31057312+jagathprakash@users.noreply.github.com>

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/artifacts/signable.go
+++ b/pkg/artifacts/signable.go
@@ -15,7 +15,9 @@ package artifacts
 
 import (
 	_ "crypto/sha256" // Recommended by go-digest.
+	_ "crypto/sha512" // Recommended by go-digest.
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -31,6 +33,10 @@ import (
 const (
 	ArtifactsInputsResultName  = "ARTIFACT_INPUTS"
 	ArtifactsOutputsResultName = "ARTIFACT_OUTPUTS"
+)
+
+var (
+	Sha1Regexp *regexp.Regexp = regexp.MustCompile(`^[a-f0-9]{40}$`)
 )
 
 type Signable interface {
@@ -339,6 +345,7 @@ func ExtractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker
 				logger.Debugf("ExtractStructuredTargetFromResults: %v", err)
 			}
 			if valid {
+				logger.Debugf("Extracted Structured data from Result %s, %s", res.Value.ObjectVal["uri"], res.Value.ObjectVal["digest"])
 				objs = append(objs, &StructuredSignable{URI: res.Value.ObjectVal["uri"], Digest: res.Value.ObjectVal["digest"]})
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/artifacts/signable.go
```diff
@@ -366,13 +369,27 @@ func isStructuredResult(res objects.Result, categoryMarker string) (bool, error)
 }
 
 func checkDigest(dig string) error {
-	prefix := digest.Canonical.String() + ":"
-	if !strings.HasPrefix(dig, prefix) {
-		return fmt.Errorf("unsupported digest algorithm: %s", dig)
+	// 93c56eeba9ec70f74c9bfd297d9516642d366cb5
```
lcarva, line 372: Leftover?

pkg/artifacts/signable.go
```diff
@@ -366,13 +369,27 @@ func isStructuredResult(res objects.Result, categoryMarker string) (bool, error)
...
+	algo_string := strings.ToLower(strings.TrimSpace(parts[0]))
+	algo := digest.Algorithm(algo_string)
+	hex := strings.TrimSpace(parts[1])
+
+	switch {
+	case algo.Available():
+		if err := algo.Validate(hex); err != nil {
+			return err
+		}
+	case algo_string == "sha1":
```
lcarva, line 386: It might be worth adding a comment here to explain why sha1 is treated differently.

pkg/artifacts/signable.go
```diff
@@ -366,13 +369,27 @@ func isStructuredResult(res objects.Result, categoryMarker string) (bool, error)
...
+		if !regexp.MustCompile(`^[a-f0-9]{40}$`).MatchString(hex) {
```
lcarva, line 387: Maybe `regexp.MustCompile` should be called at the package level? Minor performance boost.

pkg/artifacts/signable.go
```diff
@@ -366,13 +369,27 @@ func isStructuredResult(res objects.Result, categoryMarker string) (bool, error)
...
+			return fmt.Errorf("sha1 digest %s does not match regexp %s", dig, "^[a-f0-9]{40}$")
```
lcarva, line 388: Let's use `Regexp.String()` so we don't have to worry about keeping these strings in sync.

lcarva (commented):
A few minor comments, but overall it's pretty straight forward.

Do we actually want to support SHA1 since it's been deprecated since 2011?

pkg/artifacts/signable.go
jagathprakash: +1.
Done.

pkg/artifacts/signable.go
jagathprakash: Done.

lcarva:
/lgtm

> Do we actually want to support SHA1 since it's been deprecated since 2011?

I guess adding this support is fine since it's simply used to detect information about artifacts. @wlynch, any objections?

jagathprakash:
> /lgtm
> 
> > Do we actually want to support SHA1 since it's been deprecated since 2011?
> 
> I guess adding this support is fine since it's simply used to detect information about artifacts. @wlynch, any objections?

To provide more context on this.
Though there is experimental support for SHA256, by default Git commit hash is still SHA1. The git-clone task also emits SHA1.

pkg/artifacts/signable_test.go
```diff
@@ -412,6 +418,48 @@ func TestExtractStructuredTargetFromResults(t *testing.T) {
 							"digest": digest3,
 						}),
 					},
+					{
+						Name: "img2_input" + "_" + ArtifactsInputsResultName,
```
wlynch, line 422: nit: I'd include the `sha#` in the name to distinguish these.

pkg/artifacts/signable.go
```diff
@@ -336,9 +342,10 @@ func ExtractStructuredTargetFromResults(obj objects.TektonObject, categoryMarker
 		if strings.HasSuffix(res.Name, categoryMarker) {
 			valid, err := isStructuredResult(res, categoryMarker)
 			if err != nil {
-				logger.Debugf("ExtractStructuredTargetFromResults: %v", err)
```
wlynch, line 339: I don't think we want to error here - I think it's fine if a result happens to have the suffix but doesn't conform to the structured result format.

wlynch (changes requested):
> I guess adding this support is fine since it's simply used to detect information about artifacts. @wlynch, any objections?

Yeah SHA1 is unavoidable with git. AFAIK no git hosting provider supports the sha256 mode for repos yet (and even then it will take a long time to get repos to move over).

As long as we're only using it for input and not our own digest output, it's fine.

pkg/artifacts/signable_test.go
jagathprakash: Done.

pkg/artifacts/signable.go
jagathprakash: +1 agree that we should not error here.
Currently we are just logging a message (although at debug level which does not show up in the logs which is not useful and as such would prefer WARN level atleast)  and continuing.

wlynch (approved):
/lgtm

PR #692, opened 2023-01-26
Add labels and annotations to attestation

### LINKED ISSUE #591: Provide TaskRun annotations in SLSA provenance
### Feature request

I'm mostly interested in having the `experimental.tekton.dev/execution-mode` annotation available in the SLSA provenance. It might be useful to place all TaskRun annotations in there as well. A good place might be `invocation.environment.context`, similar to how this is done for GitHub Workflows in the [this example](https://slsa.dev/provenance/v0.2#github-actions).

### Use case

I want to see from the SLSA provenance if a build was done in hermetic environment or not. This allows me to evaluate the built result with higher level of confidence that there wasn't outside tampering with certain TaskRun. For example, for the build, having fetched the source code and having all the dependencies vendored I don't want to allow any external traffic, and I'd like to attest to that in the SLSA provenance generated for that TaskRun.

This was already proposed in #278, though it was closed as stale.

lcarva:
Fixes #591

# Changes

Labels and annotations are added to the PipelineRun and TaskRun attestations in `.predicate.invocation.environment`. Additionally, for PipelineRun attestations, each task also contains its own `.invocation.environment` with the corresponding labels and annotations.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
- Annotations and labels are included in the attestation.
```

### DIFF
--- a/pkg/chains/formats/slsa/attest/attest.go
+++ b/pkg/chains/formats/slsa/attest/attest.go
@@ -22,6 +22,8 @@ import (
 
 	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -55,7 +57,7 @@ func Step(step *v1beta1.Step, stepState *v1beta1.StepState) StepAttestation {
 	return attestation
 }
 
-func Invocation(source *v1beta1.ConfigSource, params []v1beta1.Param, paramSpecs []v1beta1.ParamSpec) slsa.ProvenanceInvocation {
+func Invocation(source *v1beta1.ConfigSource, params []v1beta1.Param, paramSpecs []v1beta1.ParamSpec, meta metav1.Object) slsa.ProvenanceInvocation {
 	i := slsa.ProvenanceInvocation{
 		ConfigSource: convertConfigSource(source),
 	}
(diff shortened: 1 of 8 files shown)


test/testdata/intoto/pipeline-output-image.json
```diff
@@ -42,7 +45,12 @@
                     ],
                     "invocation": {
                         "configSource": {},
-                        "parameters": {}
+                        "parameters": {},
+                        "environment": {
+                            "annotations": {
+                                "pipeline.tekton.dev/release": "9221ac3-dirty"
```
chitrangpatel, line 51: Is this value hard-coded `9221ac3-dirty`? Just checking. Probably don't want this to change if we update the installation of Tekton pipelines in CI.

chitrangpatel (approved):
Just one nit to double check.

test/testdata/intoto/pipeline-output-image.json
lcarva: You're right... I think I'll have to delete this attribute from the annotations before the call to Diff.

wlynch: You can have cmp.Diff ignore fields with this! https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreFields

pkg/chains/formats/slsa/attest/attest.go
```diff
@@ -74,6 +76,30 @@ func Invocation(source *v1beta1.ConfigSource, params []v1beta1.Param, paramSpecs
 	}
 
 	i.Parameters = iParams
+
+	environment := map[string]map[string]string{}
+
+	annotations := map[string]string{}
+	for name, value := range meta.GetAnnotations() {
+		// Ignore annotations that are not relevant to provenance information
+		if name == corev1.LastAppliedConfigAnnotation || strings.HasPrefix(name, "chains.tekton.dev/") {
```
wlynch, line 85: Thanks for thinking of this! This seems reasonable, but we should probably document this behavior somewhere so it doesn't catch someone by surprise.

wlynch (commented):
Mostly LGTM! Just some minor comments.

pkg/chains/formats/slsa/attest/attest.go
lcarva: Good idea. Added

test/testdata/intoto/pipeline-output-image.json
lcarva: Done

wlynch (approved):
/lgtm

Issue #703, opened 2023-02-13
Same attestation created multiple times


lcarva:
# Expected Behavior

For a given PipelineRun, Chains should produce at most **one** attestation for the PipelineRun, and **one** attestation for the TaskRun (depending on its configuration, and Pipeline/Task structure).

# Actual Behavior

Sometimes Chains produces **two** attestations for the PipelineRun, and **two** attestations for the TaskRun.

# Steps to Reproduce the Problem

1. Configure Chains to create PipelineRun attestations as usual.
2. Trigger a PipelineRun.
3. Just before the PipelineRun completes/succeeds, set an annotation on the PipelineRun.

The following script can be used to more easily trigger step 3:
```bash
namespace='default'
pr="$(kubectl -n "${namespace}" get pr --no-headers | awk '{printf $1 " "} ')"

while :
do
    kubectl -n "${namespace}" annotate --overwrite pr ${pr} "spam=$(date +%s.%N)"
done
```

# Additional Info

- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.2", GitCommit:"f66044f4361b9f1f96f0053dd46cb7dce5e990a8", GitTreeState:"clean", BuildDate:"2022-06-15T14:22:29Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.0", GitCommit:"a866cbe2e5bbaa01cfd5e969aa3e033f3282a8a2", GitTreeState:"clean", BuildDate:"2022-09-01T23:30:43Z", GoVersion:"go1.19", Compiler:"gc", Platform:"linux/amd64"}
```

I was able to reproduce the issue in `kind`, but I first observed it in OpenShift running `Kubernetes Version: v1.24.6+5157800`.

- Tekton Pipeline version:
```
Client version: 0.29.0
Chains version: v0.13.1
Pipeline version: v0.44.0

```

lcarva:
This happens when there are multiple controllers (Chains and something else) operating on the same resource. I noticed this on a cluster running Chains and [Pipeline as Code](https://pipelinesascode.com/) (aka PaC). However, it's obvious from the reproducer that this could be any other controller or client.

Order of events:

1. PipelineRun succeeds. A copy of the resource is added to the Chains controller event queue, and another copy is added to the PaC controller event queue.
[...]

PR #704, opened 2023-02-14
Prevent duplicate attestations

lcarva:
# Changes

See context in #703.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


wlynch (approved):
/lgtm

We're moving ahead with this knowing that this will introduce more load on the API server.

1. This is only done iff the Run is done && the signed annotation is missing. We won't query during execution and we won't query if the annotation is already set.
2. While we could update the lister cache on write, we would need to guarantee that the next reconcile hits the same controller, which we would need to be more mindful about for HA configurations. This is not complexity we think is necessary at the moment, though we're open to changing into this later.

Issue #698, opened 2023-02-03
Can not verify signatures after following instructions in "Getting Started Tutorial"


jpace121:
# Expected Behavior

I'm currently investigating chains and installed it on my local cluster and tried the instruction from the ["Getting Started Tutorial"](https://tekton.dev/docs/chains/getting-started-tutorial/) and I'm not able to successfully verify the resulting payload.

Specifically running:
```
cosign verify-blob --key k8s://tekton-chains/signing-secrets --signature ./signature ./payload
``` 
returns:
```
Error: verifying blob [./payload]: failed to verify signature                                                                                         
main.go:46: error during command execution: verifying blob [./payload]: failed to verify signature
````
not 
```
Verified OK

```

This may be a PEBKAK error, but the documentation on how to debug this is a little thin.

The only thing I can think I've done differently then just blindly follow the installation guide and then the tutorial is I loaded the following `ConfigMap`, though I believe I'm basically specifying some things to default values.

```
apiVersion: v1                                                                                                                                        
kind: ConfigMap                                                                                                                                       
metadata:                                                                                                                                             
  labels:                                                                                                                                             
    app.kubernetes.io/instance: default                                                                                                               
    app.kubernetes.io/part-of: tekton-chains                                                                                                          
  name: chains-config                                                                                                                                 
  namespace: tekton-chains                                                                                                                            
data:                                                                                                                                                 
   artifacts.taskrun.format: in-toto                                                                                                                  
   artifacts.taskrun.storage: tekton                                                                                                                  
   artifacts.pipelinerun.format: in-toto                                                                                                              
   artifacts.pipelinerun.storage: tekton                                                                                                              
   artifacts.oci.storage: tekton                                                                                                                      
   builder.id: http://tekton.internal.jpace121.net
```

# Actual Behavior
```
cosign verify-blob --key k8s://tekton-chains/signing-secrets --signature ./signature ./payload
```
returns `Error: verifying blob [./payload]: failed to verify signature                                                                                         
main.go:46: error during command execution: verifying blob [./payload]: failed to verify signature`.

# Steps to Reproduce the Problem

Blindly follow instructions https://tekton.dev/docs/chains/getting-started-tutorial/ with the exception of using the config map specified above.
Verify the payload signature.

# Additional Info

- Kubernetes version:
```
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
[...]

PR #706, opened 2023-02-16
Retain material ordering when pruning duplicates.

wlynch:
# Changes

This modifies how we are generating the material list by keeping the original order of materials as we see them. This removes some non-determinism in the provenance output caused by the map.

Encountered this when looking into #698, though I'm not sure if this was root cause (technically we should be able to still verify a signature, but multiple runs of signing may produce slightly different provenance).

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
In-toto materials now preserve the order in which they are read in. This should result in deterministic material output.
```


pkg/chains/formats/slsa/v1/internal/material/material_test.go
```diff
@@ -118,13 +117,41 @@ func TestRemoveDuplicates(t *testing.T) {
...
+				},
+			},
+		},
 	}}
 	for _, tc := range tests {
 		mat, err := RemoveDuplicateMaterials(tc.mats)
 		if err != nil {
 			t.Fatalf("Did not expect an error but got %v", err)
 		}
-		if diff := cmp.Diff(tc.want, mat, test.OptSortMaterial); diff != "" {
```
chitrangpatel, line 127: if `test.OptSortMaterial` is no longer needed now then can we remove it from the test folder and the e2e tests as well (since we are cleaning it up) ?

chitrangpatel: ie. https://github.com/tektoncd/chains/blob/main/test/sortoptions.go
and https://github.com/tektoncd/chains/blob/9950cb5dbc93493daab02af348754fc0515a11c1/test/examples_test.go#L139

wlynch: Good call! Done.

chitrangpatel:
I'm not GROOT!! (@lcarva can you review this as well. Your LGTM and approval will be allowed 😄 )

lcarva (approved):
/lgtm

nice!

PR #720, opened 2023-02-22
[TEP0122] Add feature flags field to Invocation.Environment

chitrangpatel:
For completeness of build instructions, Tekton Chains needs to pick up the feature flags used for the build. This PR adds it to `invocation.Environment` section of the slsav0.2 predicate. Like ConfigSource, this field is only populated if the feature flag: `enable-provenance-in-status` is set to `"true"`.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Add feature flags field to Invocation.Environment
```
/kind feature

### DIFF
--- a/pkg/chains/formats/slsa/v2/taskrun/taskrun.go
+++ b/pkg/chains/formats/slsa/v2/taskrun/taskrun.go
@@ -70,7 +70,7 @@ func GenerateAttestation(builderID string, payloadType config.PayloadType, tro *
 // which material the Task definition came from
 func invocation(tro *objects.TaskRunObject) slsa.ProvenanceInvocation {
 	i := slsa.ProvenanceInvocation{}
-	if p := tro.Status.Provenance; p != nil {
+	if p := tro.Status.Provenance; p != nil && p.ConfigSource != nil {
 		i.ConfigSource = slsa.ConfigSource{
 			URI:        p.ConfigSource.URI,
 			Digest:     p.ConfigSource.Digest,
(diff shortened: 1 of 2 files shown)


pkg/chains/formats/slsa/v2/taskrun/taskrun.go
```diff
@@ -70,17 +70,32 @@ func GenerateAttestation(builderID string, payloadType config.PayloadType, tro *
...
 	return i
 }
 
+// invocationEnv adds the tekton feature flags that were enabled
+// for the taskrun. In the future, we can populate versioning information
+// here as well.
+func invocationEnv(tro *objects.TaskRunObject) map[string]any {
+	var iEnv map[string]any = make(map[string]any)
+	if tro.Status.Provenance != nil && tro.Status.Provenance.FeatureFlags != nil {
+		iEnv["tekton-config-feature-flags"] = tro.Status.Provenance.FeatureFlags
```
wlynch, line 94: nit:

```suggestion
		iEnv["tekton-pipelines-feature-flags"] = tro.Status.Provenance.FeatureFlags
```

so there's no confusion about what component these are coming from.

wlynch (changes requested):

pkg/chains/formats/slsa/v2/taskrun/taskrun.go
chitrangpatel: Agreed! Thanks for the suggestion.

wlynch:
We should also start thinking about how we want to document the provenance format: having an example in the readme might go a long way!

(long term I want to look into generating an OpenAPI schema for the provenance we're producing)

chitrangpatel:
> We should also start thinking about how we want to document the provenance format: having an example in the readme might go a long way!
> 
> (long term I want to look into generating an OpenAPI schema for the provenance we're producing)

Sounds good! Now that this is complete, I can add a section in https://github.com/tektoncd/chains/blob/main/pkg/chains/formats/slsa/README.md to show an example. After we move it from slsav2alpha1 to slsav2, we can surface the example in the docs.

wlynch (approved):
/lgtm

PR #730, opened 2023-03-01
Add support for SBOM attestations

lcarva:
# Changes

~~The idea is a TaskRun produces an unsigned SBOM and pushes it to the OCI registry. The reference to the SBOM is then added to the result IMAGE_SBOM_URL. When Chains sees this result, it downloads the SBOM from the OCI registry, signs it, then adds to the image attestations with an SBOM-specific predicate.~~

~~This is also possible when using the IMAGES result, and the corresponding SBOMS result. Both must have the exact number of items.~~

~~It's a draft because it's still missing important things like unit tests and documentation. Some of the TODOs also need to be addressed. Looking for early feedback on the overall approach.~~

This commit teaches Chains to create SBOM attestations based on type hinting results from TaskRuns.
    
The general idea is that a Task generates the SBOM, then uploads it as a blob to an OCI registry. Chains then downloads the blob and uses it as the payload when creating the SBOM in-toto attestation.

Multiple SBOMs per image are allowed. They may be of different formats.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
* Added support for signing SBOMs
```


imjasonh:
(Mostly a drive-by comment, didn't review the code) I love it! If you're interested in getting `ko` to integrate with this I'd be happy to help in any way I can. `ko` currently only produces unsigned SBOMs but it'd be great if they got signed by Chains along the way. 👍

pkg/chains/formats/slsa/v2/slsav2.go
```diff
@@ -48,7 +49,7 @@ func (s *Slsa) Wrap() bool {
 	return true
 }
 
-func (s *Slsa) CreatePayload(ctx context.Context, obj interface{}) (interface{}, error) {
+func (s *Slsa) CreatePayload(ctx context.Context, kc kubernetes.Interface, obj interface{}) (interface{}, error) {
```
chitrangpatel, line 52: Any reason why we don't sign SBOMs if users use this format?

pkg/chains/formats/sbom/sbom.go
```diff
@@ -0,0 +1,106 @@
+/*
```
chitrangpatel, line 1: Wondering if this should go into `slsa/internal/sbom`?

pkg/chains/formats/sbom/sbom.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
+	"github.com/tektoncd/chains/pkg/chains/objects"
+	"k8s.io/client-go/kubernetes"
+)
+
+const maxSBOMBytes = 5 * 1024 * 1024 // 5 Megabytes
```
chitrangpatel, line 32: Why 5 Megabytes? Are most SBOMs much smaller than this?

chitrangpatel (commented):
Overall, looks great! Left some minor nits.

It would be great to see an e2e test to demonstrate how this runs. May be also an example in `tutorials`?

pkg/chains/formats/sbom/sbom.go
lcarva: I don't have the stats. The reasoning here was if it's over 5MBs, there's something wrong with your SBOM :) 

This should probably be configurable though.

pkg/chains/formats/sbom/sbom.go
lcarva: SBOM is not related to the SLSA Provenance format.

lcarva:
> Overall, looks great! Left some minor nits.
> 
> It would be great to see an e2e test to demonstrate how this runs. May be also an example in `tutorials`?

Yes, good idea! I think that's a must.

> 
> Does the SBOMURI also contain the digest? I'm wondering what would happen if the SBOM was updated by the time Chains downloaded and signed it. This is where an example/documentation would clarify things a bit.

It must include a digest. I updated the docs. Thank you!

docs/config.md
```diff
@@ -33,9 +33,18 @@ The list of images can be separated by commas or by newlines.
   value: img1@sha256:digest1, img2@sha256:digest2
 ```
 
-When processing a `TaskRun`, Chains will parse through the list, then sign and attest each image.
+Chains also suports signing SBOMs. If using the `IMAGE_URL` and `IMAGE_DIGEST` combination, the
```
arewm, line 36: Chains would be signing/attesting to something that it doesn't generate itself. While this is better than no signing, would it be worth it to call out that an individual may want to prefer to have the SBOM signed by its generator (if possible)?

The preferred approach would involve additional work in the deployment including the possibility of leveraging keyless signing and spiffe/spire.

pkg/chains/formats/sbom/sbom.go
mmalina: @lcarva FWIW, I have a testing cyclonedx sbom that has around 1700 components and the size is 5.2M ;)

docs/config.md
lcarva: +1

I do think the same argument applies to singing images. Chains is signing something that it didn't generate. Could you file an issue?

pkg/chains/formats/sbom/sbom.go
lcarva: My goodness.. So 5mb is not _big_ enough. I'll update this PR to make it configurable and set a modest default value to 10MB.

docs/config.md
arewm: I cannot resolve this thread, but feel free to do so as I created https://github.com/tektoncd/chains/issues/791.

lcarva: Thank you!

pkg/chains/formats/sbom/sbom.go
mmalina: Yeah, I was surprised myself when I checked. That sbom was for a Service Registry image. 10MB sounds good. But of course it's hard to say how often users will have sboms bigger than that.

lcarva:
If possible, I'd like to address some of the comments as a follow up. (I can file issues so we don't lose track of them).

So far:
* Add tutorial on how to use Chains to sign SBOMs
* Make max SBOM size configurable and bump default value to 10MB

Does that seem reasonable? I'm also happy to address those here otherwise.

chitrangpatel:
> If possible, I'd like to address some of the comments as a follow up. (I can file issues so we don't lose track of them).
> 
> So far:
> 
> * Add tutorial on how to use Chains to sign SBOMs
> * Make max SBOM size configurable and bump default value to 10MB
> 
> Does that seem reasonable? I'm also happy to address those here otherwise.

That sounds good to me!

sallyom:
I don't have pull around here :smile: but /LGTM too!

pkg/chains/formats/sbom/sbom.go
```diff
@@ -0,0 +1,106 @@
...
+
+	var blob bytes.Buffer
+	if _, err := io.Copy(&blob, io.LimitReader(layer, maxSBOMBytes)); err != nil {
+		return nil, err
+	}
+
+	var data json.RawMessage
+	if err := json.Unmarshal(blob.Bytes(), &data); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
wlynch: This would also solve the 5MB image problem, since we'd sign the the manifest rather than the direct SBOM content.

wlynch (changes requested):

pkg/artifacts/signable.go
```diff
@@ -436,3 +463,119 @@ func (oa *OCIArtifact) FullKey(obj interface{}) string {
...
+		return objs
+	}
+
+	if len(sboms) > 0 && sbomsFormat == "" {
+		logger.Warnf("SBOMS_FORMAT not specified")
+		return objs
+	}
+
+	for i, sbom := range sboms {
+		img := images[i]
```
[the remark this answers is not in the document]
wlynch: Should we try to discover .sbom images from the base URLs instead of using params? 🤔

docs/config.md
```diff
@@ -33,9 +33,19 @@ The list of images can be separated by commas or by newlines.
...
+`TaskRun` can also emit the `IMAGE_SBOM_URL` and the `IMAGE_SBOM_FORMAT` results. The first contains
```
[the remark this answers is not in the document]
lcarva: The values must be a digest. I believe there are some unit tests that verify this.

pkg/chains/formats/sbom/sbom.go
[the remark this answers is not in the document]
lcarva: @wlynch, I don' think we should be signing the .sbom images, mostly because of the problem outlined in [sigstore/cosign#2755](https://github.com/sigstore/cosign/issues/2755). If someone does want to sign those sbom images, then they can simply use the existing `IMAGE_URL` result type hint. Chains will happily sign those.

Using attestations to store SBOMs seems like the right approach here, and embedding the SBOM in the attestation payload seems correct as well.

Good call on documenting that this only supports JSON format. I think with a little tweak, it could support anything though.

lcarva:
@wlynch, thank you very much for the thorough review! I'll be away for the next couple of weeks, but I'll address them as soon as I'm back.

pkg/chains/formats/sbom/sbom.go
lcarva: Updated the PR to make it configurable via the key `artifacts.sbom.maxbytes`. Set the default to 10 MB. Documented.

pkg/chains/formats/slsa/v1/intotoite6.go
```diff
@@ -51,12 +53,16 @@ func (i *InTotoIte6) Wrap() bool {
 	return true
 }
 
-func (i *InTotoIte6) CreatePayload(ctx context.Context, obj interface{}) (interface{}, error) {
+func (i *InTotoIte6) CreatePayload(ctx context.Context, kc kubernetes.Interface, obj interface{}) (interface{}, error) {
```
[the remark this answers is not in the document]
lcarva: Done. Let me know what you think!

pkg/artifacts/signable.go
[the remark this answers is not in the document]
lcarva: > This is making the assumption that SBOMs are 1:1 with images - is there a way we can enforce that this pairing is correct?

Not unless we crack open the SBOM and add support to specific formats. For example, if it's SPDX, we can look at `.name`. If it's cyclonedx, we can look at `.metadata.component`. I'm trying to avoid having to teach Chains about any specific format.

> Should we try to discover .sbom images from the base URLs + digest instead of using params? 

I think that's worth discussing. The reason I didn't take that approach is because the relationship between an SBOM attachment and the image is weak. It's simply done by tag convention. It would be trivial for someone to trick Chains into signing an SBOM  that was created elsewhere without leaving any trace in the pipeline run. I think it's important to include the digest of the SBOM image in the SLSA Provenance so it can be properly audited and verified later on.

Thoughts?

lcarva:
I'm gonna move this PR to draft because there are still some open questions around the design of the feature. I've put together a [document](https://docs.google.com/document/d/1Rsc_Ph8DBkxjmnfp8uwIQS7fekTlUgB3snYegevQmeM/edit#heading=h.p3phcq4zb9gv) to facilitate driving this forward.

(Instructions on accessing the Tekton Google drive are [here](https://github.com/tektoncd/community/blob/main/contact.md#shared-drive).)

pkg/artifacts/signable.go
```diff
@@ -265,7 +267,7 @@ func (s *StructuredSignable) FullRef() string {
 	return fmt.Sprintf("%s@%s", s.URI, s.Digest)
 }
 
-func extractTargetFromResults(ctx context.Context, obj objects.TektonObject, identifierSuffix string, digestSuffix string) map[string]*StructuredSignable {
+func extractTargetFromResults(ctx context.Context, obj objects.TektonObject, identifierSuffix, digestSuffix, sbomSuffix, sbomFormatSuffix string) map[string]*StructuredSignable {
```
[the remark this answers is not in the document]
lcarva: Yeah, this wasn't great :smile: 

I introduced the concept of `structuredSignableExtractor` which I think makes this more sustainable. I think this is sufficient for now. Let me know what you think!

pkg/artifacts/signable.go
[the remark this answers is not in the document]
lcarva: I've reworked this section. It is much simplified by using blobs directly.

pkg/chains/formats/sbom/sbom.go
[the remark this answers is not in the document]
lcarva: I think this is addressed now.

lcarva:
It's been a while. I've re-worked this PR to address the comments and to focus more on SBOM attestations (instead of SBOM attachments). I updated the description to reflect these changes. It is now out of draft.

It also now includes a tutorial, `docs/tutorials/signed-sbom.md`, which will, obviously, be useful to users, but may also help reviewers.

I think this PR is pretty complete right now, but let me know if there's something missing!

(Apologies for the huge PR. I could not find a reasonable way to break it up.)

chitrangpatel:
Thanks for adding the examples and tutorials. I'm wondering if we might benefit from an e2e test here (our e2e tests are already taking ~45 mins 😅).

lcarva:
> Thanks for adding the examples and tutorials. I'm wondering if we might benefit from an e2e test here (our e2e tests are already taking ~45 mins 😅).

Yeah, for sure! Is this something that should be part of this PR or a follow up? I'm happy with either way. Just trying to not make this PR even harder to review.

docs/config.md
chitrangpatel, line 37: It can be in a separate PR but we should probably also add support for Object Results, like we have for Artifact Inputs.

```yaml
xxxSBOM:
    url:...
    format:... 
```

pkg/chains/formats/slsa/v1/intotoite6_test.go
```diff
@@ -783,3 +784,7 @@ func TestCorrectPayloadType(t *testing.T) {
 		t.Errorf("Invalid type returned: %s", i.Type())
 	}
 }
+
+func appContext() context.Context {
```
wlynch, line 788: Might as well inline context.Background directly. ¯\_(ツ)_/¯

pkg/chains/signing.go
```diff
@@ -44,6 +45,15 @@ type ObjectSigner struct {
 	Backends          map[string]storage.Backend
 	SecretPath        string
 	Pipelineclientset versioned.Interface
+	appContext        context.Context
```
wlynch, line 48: I don't think we want to attach contexts to the signer like this - this would prevent us from reusing the signer for multiple requests if there's a deadline attached to it. We already plumb through a context with Sign - we should reuse that if possible.

pkg/chains/formats/sbom/sbom.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	rawLayer, err := remote.Layer(ref, opt)
+	if err != nil {
+		return nil, err
+	}
+
+	layer, err := rawLayer.Uncompressed()
```
wlynch, line 70: This might pull in a large blob to memory. idk how good of a job the compression does though so idk how worried to be about this 🤷

pkg/chains/objects/objects.go
```diff
@@ -65,6 +68,7 @@ type TektonObject interface {
 	SupportsTaskRunArtifact() bool
 	SupportsPipelineRunArtifact() bool
 	SupportsOCIArtifact() bool
+	SupportsSBOMArtifact() bool
```
wlynch, line 71: I know this is following the pattern of existing funcs, but does this make sense to add to the Object interface? 🤔 

I don't know if "whether SBOM artifacts are supported" is a property of an Object we want to enforce on all implementations. Should this be broken out into its own interface?

pkg/artifacts/sbom.go
```diff
@@ -0,0 +1,148 @@
...
+		case "SBOMS":
+			for _, img := range strings.FieldsFunc(key.Value.StringVal, split) {
+				trimmed := strings.TrimSpace(img)
+				if trimmed == "" {
+					continue
+				}
+				if _, err := name.NewDigest(trimmed); err != nil {
+					logger.Errorf("error getting digest for SBOM image %s: %v", trimmed, err)
+					continue
+				}
```
chitrangpatel, line 117: Everything until here is completely the same as the above case. Wrap it into a func?

pkg/chains/formats/sbom/sbom.go
chitrangpatel: Should we have a max limit that users cannot exceed?

wlynch: I was worried about the uncompressed fetch, but digging in a bit more, I think this is okay and it's doing the proper reader piping 👍 (https://github.com/google/go-containerregistry/blob/190ad0e4d556f199a07951d55124f8a394ebccd9/pkg/v1/partial/compressed.go#L51-L78)

pkg/artifacts/sbom.go
```diff
@@ -0,0 +1,148 @@
...
+		return objs
+	}
+
+	if len(sboms) > 0 && sbomsFormat == "" {
+		logger.Warnf("SBOMS_FORMAT not specified")
+		return objs
+	}
+
+	for i, sbom := range sboms {
+		img := images[i]
```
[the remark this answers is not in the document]
lcarva: That's a good point. Added a commit that removes this support for now. We can add it later when the need arises, and make sure we add some sort of assurance regarding the order.

pkg/chains/signing.go
lcarva: The context here has a different purpose (which is probably a code smell).

The SBOM formater needs access to the OCI registry so it can read the SBOM (consider a private repo). To do this, we need a kube client to initialize the k8schain.

One way to get a kube client is to get it from the context, e.g. `kubeclient.Get(ctx)`. This, however, does not work with the context given to a reconcile iteration - the one `ObjectSigner.Sign` gets. It only works for the "application" context - the one `NewController` gets.

Instead of storing the application context, we could store the kube client directly in the ObjectSigner. That might make things more obvious.

@wlynch, wdyt?

pkg/artifacts/structured.go
```diff
@@ -0,0 +1,98 @@
...
+type structuredSignableExtractor struct {
+	uriSuffix        string
+	digestSuffix     string
+	sbomURISuffix    string
+	sbomFormatSuffix string
+	isValid          func(StructuredSignable) bool
+}
+
+func (b *structuredSignableExtractor) extract(ctx context.Context, obj objects.TektonObject) []StructuredSignable {
+	logger := logging.FromContext(ctx)
```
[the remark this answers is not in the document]
lcarva: Extracted the StructuredSignable piece to https://github.com/tektoncd/chains/pull/924. Hopefully, this will make it easier to also do this for the object param/results.

lcarva:
The changes in this PR means that Chains would sign any SBOM mentioned in the results of a TaskRun. 

In a system where users are able to run their own Tasks, this can be easily abused. It turns out that that's the use case I was trying to address. Thus, I have been struggling in accepting this is a good approach.

After some further reflection, I'm going to abandon this PR. 

I don't plan on removing the branch nor my fork, so if you want to take a shot at this and address the security concerns, you are welcome to do so.

If you're curious, for my particular case, I'm going with an approach that relies on the information from the SLSA Provenance since that's already signed (and generated!) by Chains. A task that produces the sbom will use something like [oras](https://oras.land/) to push the SBOM to the registry as an OCI blob and record the digest reference as a Task result. On the verification side, we can assert that the task that produced the SBOM is trusted, fetch the SBOM given the reference, and process it as needed.

PR #732, opened 2023-03-06
Remove PipelineResources usage in the examples

vdemeester:
# Changes

PipelineResources are deprecated for a long time and will be removed
very soon.

Signed-off-by: Vincent Demeester <vdemeest@redhat.com>

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Removing the PipelineResources usage in the examples as they are
deprecated and will be soon removed
```


vdemeester:
Depends on https://github.com/tektoncd/chains/pull/733 as well.

wlynch:
Created https://github.com/tektoncd/chains/issues/734 to track the rest of the removal.

wlynch (approved):
/lgtm

vdemeester:
> /remove-approve
> 
> (i suspect prow doesn't know about the github actions workflows, which is probably why the last PR was able to submit even with the workflow failing).

Yeah, I think this could be fixed by having the github action as required (because I think even if it doesn't "own" those, it should respect their status)..

vdemeester:
hum.. the e2e tests are failing.. 😓

lcarva (approved):
/lgtm

vdemeester:
Updating the json (in `testdata`) wasn't the easiest task 🙃 I wonder if it would make sense to use `golden` file instead maybe ? 🙃

lcarva (approved):
/lgtm

lcarva:
> Updating the json (in `testdata`) wasn't the easiest task upside_down_face I wonder if it would make sense to use `golden` file instead maybe ?

I'm not familiar with the `golden` file. Is that something that comes from plumbing?

wlynch:
> I'm not familiar with the `golden` file. Is that something that comes from plumbing?

@lcarva it's a general term for a static expected output file - https://ieftimov.com/posts/testing-in-go-golden-files/

So instead of having the templated json we would just have several `<testcase>.golden.json`/`<testcase>.out.json` files instead.

---

I'm neutral here - I think part of the reason why we do things this way is to insert runtime values like timestamps, but if we have a way to stabilize these with golden files that seems fine!

vdemeester:
> I'm neutral here - I think part of the reason why we do things this way is to insert runtime values like timestamps, but if we have a way to stabilize these with golden files that seems fine!

Yeah that was my guess as well. Honestly I am neutral as well, I just thought of this when editing them 🙃

PR #776, opened 2023-04-18
Cleanup: generics replace deprecated sets.String.

bendory:
# Changes

There are no expected functional changes in this PR.

In the k8s.io/apimachinery/pkg/util/sets package, the specified Set types like `sets.String` have been deprecated in favor of generics like `sets.Set[string]`. This PR changes to the generic type across our repository.

See: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets

# Submitter Checklist

  functionality, content, code)

/kind cleanup

# Release Notes

``` release-note
ACTION REQUIRED: users who build against Tekton Chains may need to adjust their code to use
generic `sets.Set[string]` types in place of `sets.String` types. See https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets for details.
```

### DIFF
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -31,6 +31,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	versioned "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/logging"
 )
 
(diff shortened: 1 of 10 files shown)


pkg/chains/signing.go
```diff
@@ -171,7 +172,7 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
 			}
 
 			// Now store those!
-			for _, backend := range signableType.StorageBackend(cfg).List() {
+			for _, backend := range sets.List[string](signableType.StorageBackend(cfg)) {
```
lcarva, line 175: Maybe use `signableType.StorageBackend(cfg).UnsortedList()` here? If so, let's also do it in `pkg/chains/verifier.go` and `test/test_utils.go`.

lcarva (approved):
Thanks for this!

Just a minor comment for consideration.

pkg/chains/signing.go
bendory: Fair consideration... but that would be a (minor) functional change in that the current API returns a sorted string slice: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String.List

IDK if anyone relies on the list being sorted. WDYT?

lcarva: It probably doesn't matter, but let's avoid the surprise :)

PR #792, opened 2023-05-02
Conform provenance.materials.uri to URI syntax

zregvart:
# Changes

According to SLSA Provenance v0.2 the `provenance.materials.uri`[1] needs to be a `ResourceURI` which is defined[2] as a _"Uniform Resource Identifier as specified in RFC 3986"_.

The `uri` field of `provenance.materials` is populated with the image reference which does not contain the scheme and therefore is not a valid URI.

This changes that to prepend the `oci://` prefix to Task and sidecar image references.

[1] https://slsa.dev/provenance/v0.2#materials.uri
[2] https://github.com/in-toto/attestation/blob/main/spec/v0.1.0/field_types.md#ResourceURI

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
SLSA Provenance v0.2 provenance.materials.uri field syntax for image references has the `oci://` prefix.
```

### DIFF
--- a/pkg/artifacts/signable.go
+++ b/pkg/artifacts/signable.go
@@ -35,6 +35,8 @@ import (
 const (
 	ArtifactsInputsResultName  = "ARTIFACT_INPUTS"
 	ArtifactsOutputsResultName = "ARTIFACT_OUTPUTS"
+	OCIScheme                  = "oci://"
+	GitSchemePrefix            = "git+"
 )
 
 var (
--- a/pkg/chains/formats/slsa/internal/material/material.go
+++ b/pkg/chains/formats/slsa/internal/material/material.go
@@ -66,7 +66,8 @@ func AddImageIDToMaterials(imageID string, mats *[]common.ProvenanceMaterial) er
 		if len(digest) == 2 {
 			// no point in partially populating the material
 			// do it if both conditions are valid.
-			m.URI = uriDigest[0]
+			uri := strings.TrimPrefix(uriDigest[0], "docker-pullable://")
+			m.URI = artifacts.OCIScheme + uri
 			m.Digest[digest[0]] = digest[1]
 			*mats = append(*mats, m)
 		} else {
--- a/test/testdata/slsa/v1/pipeline-output-image.json
+++ b/test/testdata/slsa/v1/pipeline-output-image.json
@@ -35,7 +35,7 @@
                             "arguments": null,
                             "environment": {
                                 "container": "create-dockerfile",
-                                "image": "distroless.dev/busybox@sha256:186312fcf3f381b5fc1dd80b1afc0d316f3ed39fb4add8ff900d1f0c7c49a92c"
+                                "image": "oci://distroless.dev/busybox@sha256:186312fcf3f381b5fc1dd80b1afc0d316f3ed39fb4add8ff900d1f0c7c49a92c"
                             },
                             "annotations": null
                         }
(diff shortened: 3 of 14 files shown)


test/testdata/slsa/v1/task-output-image.json
```diff
@@ -44,7 +44,7 @@
         "materials": [
 	    {{range .URIDigest}}
 	    {
-		"uri": "{{.URI}}",
+		"uri": "oci://{{.URI}}",
```
chitrangpatel, line 47: May be we should capture this in the URI itself. The reason is that materials can also contain repos of the remote tasks and pipelines which could be on git/oci/artifact hub.

chitrangpatel: I think our example tests inline the spec (instead of fetching them remotely) so they are not throwing errors.

test/testdata/slsa/v2/task-output-image.json
```diff
@@ -79,7 +79,7 @@
         "materials": [
 	    {{range .URIDigest}}
 	    {
-		"uri": "{{.URI}}",
+		"uri": "oci://{{.URI}}",
```
chitrangpatel, line 82: same comment as above.

test/testdata/slsa/v1/pipeline-output-image.json
```diff
@@ -72,7 +72,7 @@
         "materials": [
 	    {{range .URIDigest}}
 	    {
-		"uri": "{{.URI}}",
+		"uri": "oci://{{.URI}}",
```
chitrangpatel, line 75: see comment below.

pkg/artifacts/signable.go
```diff
@@ -316,7 +316,7 @@ func RetrieveMaterialsFromStructuredResults(ctx context.Context, obj objects.Tek
 		alg := splits[0]
 		digest := splits[1]
 		mats = append(mats, common.ProvenanceMaterial{
-			URI:    s.URI,
+			URI:    fmt.Sprintf("oci://%s", s.URI),
```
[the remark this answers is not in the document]
chitrangpatel: See https://github.com/tektoncd/community/blob/main/teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md for context.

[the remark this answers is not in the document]
chitrangpatel: Basically its up to the task authors to declare what was fetched.

zregvart:
Thanks for the review @chitrangpatel I was overreaching by prefixing the results with `oci://`. I think I've addressed the comments in the latest change I just pushed. Can you have another look?

pkg/chains/formats/slsa/internal/material/material.go
```diff
@@ -66,7 +66,8 @@ func AddImageIDToMaterials(imageID string, mats *[]common.ProvenanceMaterial) er
 		if len(digest) == 2 {
 			// no point in partially populating the material
 			// do it if both conditions are valid.
-			m.URI = uriDigest[0]
+			uri := strings.TrimPrefix(uriDigest[0], "docker-pullable://")
+			m.URI = fmt.Sprintf("oci://%s", uri)
```
chuangw6, line 70: Thanks @zregvart for fixing this. Can we make the prefix a constant so that we can have a single place to define what the prefixes are? Perhaps we should do it for `git+` as well.

chitrangpatel (approved):
Thanks!

Issue #780, opened 2023-04-20
Chains v1.0.0


wlynch:
We want to cut a stable release for Chains.

Most of our actual feature set has been stable for some time - we don't expect any major changes to how chains operates.

However, there are still things we want to tackle before a v1 is cut, primarily around library stability and testing:

- [ ] Pipelines Regression Testing
  - [x] https://github.com/tektoncd/chains/issues/716
  - [ ] Add tests to catch changes in pipelines versions.
- [ ] https://github.com/tektoncd/chains/issues/640
- [ ] https://github.com/tektoncd/chains/issues/665
- [ ] Limit client libraries for external usage

Things that are not in scope:
- SLSA v1.0 support (additive feature)

aaron-prindle:
Is it possible to re-open this?  Seems it was auto-closed but some of the chains v1.0.0 TODO issues are still open.

aaron-prindle:
Created an issue for the client lib work:
https://github.com/tektoncd/chains/issues/1044

PR #842, opened 2023-06-26
v1: Rewrite OCI storage to use new interface

wlynch:
# Changes

This change is part of a larger effort for refactoring external facing libraries to make them easier to support for v1+: https://github.com/tektoncd/chains/issues/780

This introduces a new signing interface using generics. This allows us to cut out much of the type switching and parameter plumbing by making the key fields that storers need to make decisions on how to store signatures / attestations:

1. The original tekton object
2. The artifact we extracted from the tekton object
3. The formatted object we transformed the artifact into
4. The signature/cert details

This change marks the existing Backend implementation as deprecated, but is intended to be backwards compatible and still supported for the time being. There should be no change in user behavior.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/docs/v1-proposal.md
+++ b/docs/v1-proposal.md
@@ -1,0 +1,166 @@
+# v1 client changes proposal - storage refactoring
+
+With v1 approaching, I want to take a moment to look at changes we want to make
+to the existing client libraries to better set us up for long term maintenance.
+
+We already know that we want to reduce the external library surface of chains.
+But to do this, we need to define better interfaces between components that we
+expect external clients to use.
+
+Today, I think a lot of the codebase's complexity has come from a few places:
+
+1. Storage libraries
+
+   Each storage type needs different pieces of data - i.e. Grafeas and OCI need
+   to distinguish image signatures and attestation formats, and some clients
+   need the original object to extract out information like GVKs, names,
+   namespaces, etc. This has led to a organic growth of the chains libraries to
+   pass the different types of data around, and a lot of typecasting and other
+   generic object tricks.
+
+   Looking at the storage interfaces, I think this data roughly boils down to:
+
+   - The original Tekton object
+   - The formatted data object
+   - The signed payload + signature (with optional cert information)
+
+   Unlike when chains first started, we now have another useful tool available
+   to us: generics. I think we can use this to create clearer interfaces.
+
+2. Dependence on the config package
+
+   tkn depends on the chains server config, but it probably shouldn't. We should
+   aim to have better ways to initialize clients for others to use.
+
+Good news, I don't think we're far off, but we should make some changes
+
+## Signables
+
+At it's core, Chains is basically an ETL pipeline. We Extract artifacts from run
+objects, Transform and sign them, then Load them into storage.
+
+```go
+type Signable[T any] interface {
+	Extract(ctx context.Context, obj objects.TektonObject) []T{}
+}
+```
+
+## Payloaders
+
+I think payloaders are mostly in a good place, though we can introduce generics
+to start creating stricter type relationships between Signables and Payloaders.
+
+```go
+type Payloader[Input any, Output BinaryMarshaler] interface {
+	CreatePayload(ctx context.Context, in Input) (Output, error)
+}
+```
+
+tl;dr: Some type comes in, some type comes out.
+
+[BinaryMarshaler comes from the encoding package](https://pkg.go.dev/encoding#BinaryMarshaler),
+but basically all we're aiming for here is to make sure we can get a []byte for
+signing. For existing payload types, this may mean we need to wrap external
+types for this functionality.
+
+## Signers
+
+Signers are mostly in a good spot, though we should probably just embrace []byte
+instead of typecasting between string for cert details.
+
+```go
+type Signer interface {
+	signature.SignerVerifier
+	Cert() []byte
+	Chain() []byte
+}
+```
+
+## Storers
+
+Now that we have all the other pieces defined, we can now have stricter typing
+for storing:
+
+```go
+type Storer[Input any, Output any] interface {
+	Store(ctx context.Context, req *StoreRequest) (*StoreResponse, error)
+}
+
+type StoreRequest[Input any, Output any] struct {
+    Object objects.TektonObject
+    Artifact Input
+    Payload Output
+    Bundle *signing.Bundle
+}
+
+type StoreResponse struct {
+    // Some identifier for what we uploaded to reference later?
+    ID string
+}
+
+type Bundle struct {
+	Content   []byte
+	Signature []byte
+	Cert      []byte
+	Chain     []byte
+}
+```
+
+While the StoreRequest struct may not be necessary, it has a nice RPC-like
+quality in that it will make it easier to add/remove fields in the future.
+
+## Attestors
+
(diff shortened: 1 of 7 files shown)


docs/v1-proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+   tkn depends on the chains server config, but it probably shouldn't. We should
+   aim to have better ways to initialize clients for others to use.
+
+Good news, I don't think we're far off, but we should make some changes
+
+## Signables
+
+At it's core, Chains is basically an ETL pipeline. We Extract artifacts from run
+objects, Transform and sign them, then Load them into storage.
```
lcarva, line 40: This is a good explanation. It would be great to show how they map to the `Signable`, `Payloader`, `Signer` and `Storer`  interfaces below.

I think you mean something like this: We Extract artifacts from run objects (`Signable`), Transform (`Payloader`) and sign (`Signer`) them, then Load them into storage (`Storer`). If so, that makes perfect sense to me. 

Maybe in the future, we may want to rename these interfaces. `Signer` and `Storer` are clear. `Signable` is confusing because it has nothing to do with signing - maybe `Extractor`? `Payloader` is a bit ambiguous - `Transformer` then `Transformer.Transform`? I'm nitpicking. Just trying to make this easier for newcomers to understand.

docs/v1-proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+type Bundle struct {
+	Content   []byte
+	Signature []byte
+	Cert      []byte
+	Chain     []byte
+}
+```
+
+While the StoreRequest struct may not be necessary, it has a nice RPC-like
```
lcarva, line 109: +1

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/golangci/golangci-lint v1.53.3
```
lcarva, line 9: It would be great to minimize the changes in go.mod for this PR. Consider splitting it up.

pkg/chains/storage/oci/attestation.go
```diff
@@ -0,0 +1,100 @@
...
+// AttestationStorer stores in-toto Attestation payloads in OCI registries.
+type AttestationStorer struct {
+	// repo configures the repo where data should be stored.
+	// If empty, the repo is inferred from the Artifact.
+	repo *name.Repository
+	// remoteOpts are additional remote options (i.e. auth) to use for client operations.
+	remoteOpts []remote.Option
+}
+
+func NewAttestationStorer(cfg config.Config) (*AttestationStorer, error) {
```
lcarva, line 47: Should we process `cfg.Storage.OCI.Insecure` here as well? I believe if that's set to `true`, then it applies to any OCI repo, not just the one configured in `cfg.Storage.OCI.Repository`.

pkg/chains/storage/oci/attestation.go
```diff
@@ -0,0 +1,100 @@
...
+	return &AttestationStorer{
+		repo: repo,
+	}, nil
+}
+
+func (s *AttestationStorer) Store(ctx context.Context, req *api.StoreRequest[name.Digest, in_toto.Statement]) (*api.StoreResponse, error) {
+	logger := logging.FromContext(ctx)
+	// upload an attestation for each subject
+	logger.Info("Starting to upload attestations to OCI ...")
+	for _, subj := range req.Payload.Subject {
```
lcarva, line 66: Is looping over the subjects right? The for-loop only references `subj` for logging purposes.

pkg/chains/storage/oci/attestation.go
```diff
@@ -0,0 +1,100 @@
...
+			return nil, err
+		}
+
+		// Publish the signatures associated with this entity
+		if err := ociremote.WriteAttestations(repo, newImage, ociremote.WithRemoteOptions(s.remoteOpts...)); err != nil {
+			return nil, err
+		}
+		logger.Infof("Successfully uploaded attestation for %s", imageName)
+	}
+	return &api.StoreResponse{}, nil
```
lcarva, line 99: `return nil, nil` ?

pkg/chains/storage/oci/simple.go
```diff
@@ -0,0 +1,96 @@
...
+	// remoteOpts are additional remote options (i.e. auth) to use for client operations.
+	remoteOpts []remote.Option
+}
+
+var (
+	_ api.Storer[name.Digest, simple.SimpleContainerImage] = &SimpleStorer{}
+)
+
+func NewSimpleStorerFromConfig(cfg config.Config) (*SimpleStorer, error) {
+	var repo *name.Repository
```
lcarva, line 47: Same comment about handling `cfg.Storage.OCI.Insecure` here. Probably ok to duplicate given the other implementation is deprecated and will be removed.

lcarva (commented):
Love these changes! Just left a couple of minor comments.

docs/v1-proposal.md
wlynch: Yeah I agree. I'm somewhat biasing on existing interface names at the moment. Happy to rename these (though probably in a separate PR).

pkg/chains/storage/oci/attestation.go
wlynch: Good call. Done!

pkg/chains/storage/oci/attestation.go
wlynch: Good catch! This was copy-pasted from the outer loop. Removed.

pkg/chains/storage/oci/attestation.go
wlynch: I thought about this - this is inspired by gRPC which prefers this style of successful responses (it just so happens that there's no content).

The difference here is negligible though so 🤷

pkg/chains/storage/oci/simple.go
wlynch: Yeah long term I want to see if we can refactor this + attestation storer to use https://pkg.go.dev/github.com/google/go-containerregistry@v0.15.2/pkg/v1/remote#Pusher so we can consolidate these.

I think we need to make some upstream changes though (since options like Insecure are name options, not remote options) so I didn't try to tackle this in this PR.

pkg/chains/storage/oci/attestation.go
```diff
@@ -0,0 +1,98 @@
...
+	repo *name.Repository
+	// remoteOpts are additional remote options (i.e. auth) to use for client operations.
+	remoteOpts []remote.Option
+}
+
+func NewAttestationStorer(cfg config.Config) (*AttestationStorer, error) {
+	var repo *name.Repository
+	if r := cfg.Storage.OCI.Repository; r != "" {
+		var opts []name.Option
+		if cfg.Storage.OCI.Insecure {
```
lcarva, line 50: I believe this should be moved outside of the outer if-block. At least today, this config option applies to any repo, not just `Storage.OCI.Repository`.

wlynch: It's a bit confusing, but this was already happening outside of this block - this was only handling the "you asked us to store the objects in another repo" resolution.

I agree this isn't super clear though, so refactored this a bit more to try and consolidate this (and as an added bonus, removed the need to pass config.Config in!)

lcarva (approved):

Issue #846, opened 2023-06-28
Add resolved top-level build config yaml to external parameters


chitrangpatel:
In https://docs.google.com/document/d/1ewqtPXyg_y3MmU6Tc6l1X8nfzjt0AJHlP6VOnFsGNpQ/edit#heading=h.8hw7rf805u8r, we decided the add the complete `runSpec` to `externalParameters` and the resolved top-level task/pipeline yaml to resolved dependencies. This has caused some friction in alignment between the Slsa specification, builders and verifiers.

According to the [SLSA spec](https://slsa.dev/provenance/v1#provenance), the external parameters should capture the inputs to the build interface. In the case where the build config is located remotely (e.g. in version control), one of the key ingredients that **must be validated** is the repo from where the build config was fetched. This information is available in the `runSpec` under `pipelineRef` or the `taskRef` however, that syntax is very specific to Tekton which makes it very challenging for generic verifiers (that validate provenances from several CI/CD systems) to identify without making the verification policy understand the Tekton API. To make matters worse, the repo uri is not necessarily embedded in the [resolver parameters](https://github.com/tektoncd/pipeline/blob/main/pkg/resolution/resolver/git/params.go). This means that depending on the parameters, verifiers need to add logic to extract the URI for verification. 

In Tekton Chains, we have this information completely resolved and accessible. However, we add this to the resolved dependencies section of the provenance. 

Here, we also propose adding this resolved URI to externalParameters to make it easy for generic verifies (like binary Authorization) to access a field that can be easily verified.

**Note** that this will also make it synchronous with the suggested [migration path from SLSA v0.2 to v1.0](https://slsa.dev/provenance/v1#migrating-from-02) where it says that the externalParameters should also contain the `old invocation.ConfigSource`. 

cc @chuangw6, @bobcatfish, @MarkLodato

chuangw6:
Thanks @chitrangpatel !
The idea SGTM. Can you please be explicit about the extra field we want to introduce to the externalParameters? Is it just a single uri field, or is it an object containing key value pairs such as uri, digest and entryppoint? Also what names should we use for those fields?

chitrangpatel:
I was thinking of this:
```yaml
buildConfigSource:
  ref: either git reference (i.e. commit sha - sha1:xxxx) or image digest (e.g. sha256:91fb5e20325059e8feb48c966a83d616596054c5edf811b5bc673683e6ecabb6)
  repository: either git repository uri or image repository uri
[...]

PR #863, opened 2023-07-13
Add top-level build config to External Parameters

chitrangpatel:
In the [build type design](https://docs.google.com/document/d/1ewqtPXyg_y3MmU6Tc6l1X8nfzjt0AJHlP6VOnFsGNpQ/edit#heading=h.8hw7rf805u8r), we decided the add the complete runSpec to externalParameters and the resolved top-level task/pipeline yaml to resolved dependencies. This has caused some friction in alignment between the Slsa specification, builders and verifiers.

According to the [SLSA spec](https://slsa.dev/provenance/v1#provenance), the external parameters should capture the inputs to the build interface. In the case where the build config is located remotely (e.g. in version control), one of the key ingredients that must be validated is the repo from where the build config was fetched. This information is available in the runSpec under pipelineRef or the taskRef however, that syntax is very specific to Tekton which makes it very challenging for generic verifiers (that validate provenances from several CI/CD systems) to identify without making the verification policy understand the Tekton API. To make matters worse, the repo uri is not necessarily embedded in the [resolver parameters](https://github.com/tektoncd/pipeline/blob/main/pkg/resolution/resolver/git/params.go). This means that depending on the parameters, verifiers need to add logic to extract the URI for verification.

In Tekton Chains, we have this information completely resolved and accessible. However, we add this to the resolved dependencies section of the provenance. **Note** that this will also make it synchronous with the suggested [migration path from SLSA v0.2 to v1.0](https://slsa.dev/provenance/v1#migrating-from-02) where it says that the externalParameters should also contain the old invocation.ConfigSource.

This PR surfaces this information in `externalParameters` section of the SLSA 1.0 predicate and addresses issue https://github.com/tektoncd/chains/issues/846.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Add top-level build config to External Parameters
```

### DIFF
--- a/pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
+++ b/pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
@@ -94,6 +94,29 @@ func internalParameters(pro *objects.PipelineRunObject) map[string]any {
 // externalParameters adds the pipeline run spec
 func externalParameters(pro *objects.PipelineRunObject) map[string]any {
 	externalParams := make(map[string]any)
+
+	// add the origin of top level pipeline config
+	// isRemotePipeline checks if the pipeline was fetched using a remote resolver
+	isRemotePipeline := false
+	if pro.Spec.PipelineRef != nil {
+		if pro.Spec.PipelineRef.Resolver != "" && pro.Spec.PipelineRef.Resolver != "Cluster" {
+			isRemotePipeline = true
+		}
+	}
+
+	if p := pro.Status.Provenance; p != nil && p.RefSource != nil && isRemotePipeline {
+		ref := ""
+		for alg, hex := range p.RefSource.Digest {
+			ref = fmt.Sprintf("%s:%s", alg, hex)
+			break
+		}
+		buildConfigSource := map[string]string{
+			"ref":        ref,
+			"repository": p.RefSource.URI,
+			"path":       p.RefSource.EntryPoint,
+		}
+		externalParams["buildConfigSource"] = buildConfigSource
+	}
 	externalParams["runSpec"] = pro.Spec
 	return externalParams
 }
--- a/pkg/chains/formats/slsa/v2alpha2/internal/taskrun/taskrun.go
+++ b/pkg/chains/formats/slsa/v2alpha2/internal/taskrun/taskrun.go
@@ -91,6 +91,27 @@ func internalParameters(tro *objects.TaskRunObject) map[string]any {
 // externalParameters adds the task run spec
 func externalParameters(tro *objects.TaskRunObject) map[string]any {
 	externalParams := make(map[string]any)
+	// add origin of the top level task config
+	// isRemoteTask checks if the task was fetched using a remote resolver
+	isRemoteTask := false
+	if tro.Spec.TaskRef != nil {
+		if tro.Spec.TaskRef.Resolver != "" && tro.Spec.TaskRef.Resolver != "Cluster" {
+			isRemoteTask = true
+		}
+	}
+	if t := tro.Status.Provenance; t != nil && t.RefSource != nil && isRemoteTask {
+		ref := ""
+		for alg, hex := range t.RefSource.Digest {
+			ref = fmt.Sprintf("%s:%s", alg, hex)
+			break
+		}
+		buildConfigSource := map[string]string{
+			"ref":        ref,
+			"repository": t.RefSource.URI,
+			"path":       t.RefSource.EntryPoint,
+		}
+		externalParams["buildConfigSource"] = buildConfigSource
+	}
 	externalParams["runSpec"] = tro.Spec
 	return externalParams
 }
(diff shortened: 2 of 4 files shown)


pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
```diff
@@ -94,6 +94,15 @@ func internalParameters(pro *objects.PipelineRunObject) map[string]any {
 // externalParameters adds the pipeline run spec
 func externalParameters(pro *objects.PipelineRunObject) map[string]any {
 	externalParams := make(map[string]any)
+	// add top level pipeline config
```
chuangw6, line 97:
```suggestion
	// add the origin of top level pipeline config
```

pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
```diff
@@ -94,6 +94,15 @@ func internalParameters(pro *objects.PipelineRunObject) map[string]any {
...
+	if p := pro.Status.Provenance; p != nil && p.RefSource != nil && pro.Spec.PipelineRef != nil {
```
chuangw6, line 98: It seems to me that `pro.Spec.PipelineRef != nil` condition is not needed here. Only relying on the Provenance.RefSource field to report the origin should be sufficient here.

pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
```diff
@@ -94,6 +94,15 @@ func internalParameters(pro *objects.PipelineRunObject) map[string]any {
...
+		buildConfigSource := map[string]any{
+			"ref":        p.RefSource.Digest,
```
chuangw6, line 100: p.RefSource.Digest is a map field, but I suppose the intention for the `ref` field should be `string` according to the discussion in https://github.com/tektoncd/chains/issues/846#issuecomment-1611835123 and also slsa recommendation for GHA https://github.com/slsa-framework/github-actions-buildtypes/tree/main/workflow/v1#external-parameters.

Maybe we need to grab the specific digest from the map (also there should be always only one pair there).

pkg/chains/formats/slsa/v2alpha2/internal/taskrun/taskrun.go
```diff
@@ -91,6 +91,15 @@ func internalParameters(tro *objects.TaskRunObject) map[string]any {
 // externalParameters adds the task run spec
 func externalParameters(tro *objects.TaskRunObject) map[string]any {
 	externalParams := make(map[string]any)
+	// add top level task config
+	if t := tro.Status.Provenance; t != nil && t.RefSource != nil && tro.Spec.TaskRef != nil {
+		buildConfigSource := map[string]any{
+			"ref":        t.RefSource.Digest,
```
chuangw6, line 97: Same comments for pipeline level apply here.

chuangw6 (approved):
Thank you @chitrangpatel !

pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
chitrangpatel: Actually, we need to check if the resolver is not the "ClusterResolver" as well. The reason is that we plan to provide a digest (`checksum`) of in-cluster pipeline and task too. Here we only want to capture `remote` reference right?

pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
chitrangpatel: should I always set it to a string as `sha1:abc123` or `sha256:abc123`?

pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
chitrangpatel: I'm thinking it should be this:
```
if p := pro.Status.Provenance; p != nil && p.RefSource != nil &&  pro.Spec.PipelineRef != nil && pro.Spec.PipelineRef.Resolver != nil && pro.Spec.PipelineRef.ClusterResolver == nil
```
WDYT?

pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
chuangw6: SGTM!

pkg/chains/formats/slsa/v2alpha2/internal/taskrun/taskrun.go
chitrangpatel: done.

pkg/chains/formats/slsa/v2alpha2/internal/pipelinerun/pipelinerun.go
chitrangpatel: done.

chitrangpatel:
Happy to put this on hold if issue https://github.com/tektoncd/chains/issues/846 needs more discussion @wlynch @lcarva

lcarva (approved):
/lgtm

Makes sense based on [this](https://github.com/tektoncd/chains/issues/846#issuecomment-1624190731) explanation.

Issue #850, opened 2023-07-06
Enhancement on Pipeline level provenance


chuangw6:
### Feature request

1. For a PipelineRun, Chains is able to dive into individual taskruns' type hinting results without pipeline author having to surface task results to pipeline results. 
*(Current Chains: only recognize pipeline-level type hinting results for a PipelineRun.)*

1. Within a pipelinerun, Chains should generate a provenance as soon as it sees an artifact is generated by a task even though some tasks post artifact generation fail, which as a result fails the whole pipelinerun.
*(Current Chains: will not generate a provenance until individual taskruns and the pipelinerun succeed.)*

1. Per-artifact Provenance instead of Per-run Provenance
*(Current Chains: generate multiple provenances with **the exact same content**, and those provenances are associated with different artifacts being generated from that Run)* 
Examples:
    - If a pipeline consists of tasks like `clone -> build-push -> slack-notify`, the provenance associated with the artifact should only contain the info of the tasks that contribute to the build process of the artifact, which would be `clone` and `build-push` in this case. In the 2nd feature request mentioned above, when `slack-notify` fails, the provenance should still be available for that artifact. 
    -  If a pipeline consists of 2 sets of tasks for producing 2 different artifacts, `clone1 -> build-push-go` (producing go artifact) and `clone 2 -> build-push-oci` (producing oci artifact), the provenance associated with "go artifact" should only contain information of tasks `clone1` and `build-push-go`, whereas the provenance associated with "oci artifact" should only contain information of tasks `clone2` and `build-push-oci`

### Use case

- Feature 1: better user experience and avoid errors that could happen when author propagates task results to pipeline results
- Feature 2: the provenance associated with the artifact being generated in a pipeline is still available even though some post-artifact generation tasks i.e. slack notify or deploy task, or the tasks that have nothing related to building that artifact i.e. tasks for building other language artifacts failed. In particular, if we make provenance available before a deploy task, it would be helpful for some deployment gate mechanism which can do provenance verification and decide whether to proceed deployment.
- Feature 3: "focused" provenance that only contain information of tasks that are related to the artifact generation is the real "per artifact" provenance.

chuangw6:
Looking for feedback and thoughts! Thanks everyone.

chuangw6:
cc @chitrangpatel @lcarva @wlynch

lcarva:
Thanks for filing this feature request! I do disagree with it for various reasons that I have outlined below.

[...]

PR #866, opened 2023-07-18
Smarter Chains: check taskrun level results for Subjects

chuangw6:
# Changes

Step 1/2 of https://github.com/tektoncd/chains/issues/850

Prior, Chains only looks for pipeline results to understand what
artifacts were generated in a pipeline. That means pipeline authors need
to propagate child TaskRun results to pipeline level and name the pipeline
results in type hinting way even though the pulled tasks already produce
type hinting results.

Now, we introduced a new configmap field `artifacts.pipelinerun.enable-deep-inspection`
to allow Chains to inspect both pipeline results and child task results
to understand what artifacts were generated throughout a pipeline.

This way, pipeline authors no longer need to worry about the rules when
writting a pipeline as long as they pull in right tasks that produce type hinting results.
That said, users still have ability to propagate task results to pipeline
level if the tasks they referenced do not produce type hinting results.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Enable Chains to inspect both pipelinerun results and its child taskrun results.

Introduce a new configmap field `artifacts.pipelinerun.enable-deep-inspection` to allow users to select whether Chains should dive into child taskruns for capturing results.
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -64,9 +64,11 @@ Supported keys include:
 | `artifacts.pipelinerun.format` | The format to store `PipelineRun` payloads in. | `in-toto`, `slsa/v1`| `in-toto` |
 | `artifacts.pipelinerun.storage` | The storage backend to store `PipelineRun` signatures in. Multiple backends can be specified with comma-separated list ("tekton,oci"). To disable the `PipelineRun` artifact input an empty string ("").  | `tekton`, `oci`, `gcs`, `docdb`, `grafeas` | `tekton` |
 | `artifacts.pipelinerun.signer` | The signature backend to sign `PipelineRun` payloads with. | `x509`, `kms` | `x509` |
+| `artifacts.pipelinerun.enable-deep-inspection` | This boolean option will configure whether Chains should inspect child taskruns in order to capture inputs/outputs within a pipelinerun. `"false"` means that Chains only checks pipeline level results, whereas `"true"` means Chains inspects both pipeline level and task level results. | `"true"`, `"false"` | `"false"` | 
 
-> NOTE: For grafeas storage backend, currently we only support Container Analysis. We will make grafeas server address configurabe within a short time.
-> NOTE: `slsa/v1` is an alias of `in-toto` for backwards compatibility.
+> NOTE: 
+> - For grafeas storage backend, currently we only support Container Analysis. We will make grafeas server address configurabe within a short time.
+> - `slsa/v1` is an alias of `in-toto` for backwards compatibility.
 
 ### OCI Configuration
 
--- a/pkg/chains/formats/slsa/extract/extract.go
+++ b/pkg/chains/formats/slsa/extract/extract.go
@@ -27,6 +27,7 @@ import (
 	"github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/common"
 	"github.com/tektoncd/chains/internal/backport"
 	"github.com/tektoncd/chains/pkg/artifacts"
+	"github.com/tektoncd/chains/pkg/chains/formats/slsa/internal/slsaconfig"
 	"github.com/tektoncd/chains/pkg/chains/objects"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"knative.dev/pkg/logging"
@@ -39,7 +40,102 @@ import (
 //   - have suffix `IMAGE_URL` & `IMAGE_DIGEST` or `ARTIFACT_URI` & `ARTIFACT_DIGEST` pair.
 //   - the `*_DIGEST` field must be in the format of "<algorithm>:<actual-sha>" where the algorithm must be "sha256" and actual sha must be valid per https://github.com/opencontainers/image-spec/blob/main/descriptor.md#sha-256.
 //   - the `*_URL` or `*_URI` fields cannot be empty.
-func SubjectDigests(ctx context.Context, obj objects.TektonObject) []intoto.Subject {
+//
+//nolint:all
+func SubjectDigests(ctx context.Context, obj objects.TektonObject, slsaconfig *slsaconfig.SlsaConfig) []intoto.Subject {
+	var subjects []intoto.Subject
+
+	switch obj.GetObject().(type) {
+	case *v1beta1.PipelineRun:
+		subjects = subjectsFromPipelineRun(ctx, obj, slsaconfig)
+	case *v1beta1.TaskRun:
+		subjects = subjectsFromTektonObject(ctx, obj)
+	}
+
+	sort.Slice(subjects, func(i, j int) bool {
+		return subjects[i].Name <= subjects[j].Name
+	})
+
+	return subjects
+}
+
+func subjectsFromPipelineRun(ctx context.Context, obj objects.TektonObject, slsaconfig *slsaconfig.SlsaConfig) []intoto.Subject {
+	prSubjects := subjectsFromTektonObject(ctx, obj)
+
+	// If deep inspection is not enabled, just return subjects observed on the pipelinerun level
+	if !slsaconfig.DeepInspectionEnabled {
+		return prSubjects
+	}
+
+	logger := logging.FromContext(ctx)
+	// If deep inspection is enabled, collect subjects from child taskruns
+	var result []intoto.Subject
+
+	pro := obj.(*objects.PipelineRunObject)
+
+	pSpec := pro.Status.PipelineSpec
+	if pSpec != nil {
+		pipelineTasks := append(pSpec.Tasks, pSpec.Finally...)
+		for _, t := range pipelineTasks {
+			tr := pro.GetTaskRunFromTask(t.Name)
+			// Ignore Tasks that did not execute during the PipelineRun.
+			if tr == nil || tr.Status.CompletionTime == nil {
+				logger.Infof("taskrun status not found for task %s", t.Name)
+				continue
+			}
+
+			trSubjects := subjectsFromTektonObject(ctx, objects.NewTaskRunObject(tr))
+			for _, s := range trSubjects {
+				result = addSubject(result, s)
+			}
+		}
+	}
+
+	// also add subjects observed from pipelinerun level with duplication removed
+	for _, s := range prSubjects {
+		result = addSubject(result, s)
+	}
+
+	return result
+}
+
+// addSubject adds a new subject item to the original slice.
+func addSubject(original []intoto.Subject, item intoto.Subject) []intoto.Subject {
+
+	for i, s := range original {
+		// if there is an equivalent entry in the original slice, merge item's DigestSet
+		// into the existing entry's DigestSet.
+		if subjectEqual(s, item) {
+			mergeMaps(original[i].Digest, item.Digest)
+			return original
+		}
+	}
+
+	original = append(original, item)
+	return original
+}
+
+// two subjects are equal if and only if they have same name and have at least
+// one common algorithm and hex value.
+func subjectEqual(x, y intoto.Subject) bool {
+	if x.Name != y.Name {
+		return false
+	}
+	for algo, hex := range x.Digest {
+		if y.Digest[algo] == hex {
+			return true
+		}
+	}
+	return false
+}
+
+func mergeMaps(m1 map[string]string, m2 map[string]string) {
+	for k, v := range m2 {
+		m1[k] = v
+	}
+}
+
+func subjectsFromTektonObject(ctx context.Context, obj objects.TektonObject) []intoto.Subject {
 	logger := logging.FromContext(ctx)
 	var subjects []intoto.Subject
 
@@ -121,9 +217,7 @@ func SubjectDigests(ctx context.Context, obj objects.TektonObject) []intoto.Subj
 			})
 		}
 	}
-	sort.Slice(subjects, func(i, j int) bool {
-		return subjects[i].Name <= subjects[j].Name
-	})
+
 	return subjects
 }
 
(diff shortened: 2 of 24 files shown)


pkg/chains/formats/slsa/extract/extract.go
```diff
@@ -34,12 +34,67 @@ import (
...
+	}
+
+	sort.Slice(subjects, func(i, j int) bool {
+		return subjects[i].Name <= subjects[j].Name
+	})
+
+	return subjects
+}
+
+func subjectsFromPipelineRun(ctx context.Context, obj objects.TektonObject) []intoto.Subject {
```
lcarva, line 59: The PipelineRun results should be taken into account as well.

pkg/chains/formats/slsa/extract/extract.go
```diff
@@ -34,12 +34,67 @@ import (
...
+// The type hinting fields expected in TaskRun results help identify the generated software artifacts in a TaskRun/PipelineRun.
 // Valid type hinting fields must:
 //   - have suffix `IMAGE_URL` & `IMAGE_DIGEST` or `ARTIFACT_URI` & `ARTIFACT_DIGEST` pair.
 //   - the `*_DIGEST` field must be in the format of "<algorithm>:<actual-sha>" where the algorithm must be "sha256" and actual sha must be valid per https://github.com/opencontainers/image-spec/blob/main/descriptor.md#sha-256.
 //   - the `*_URL` or `*_URI` fields cannot be empty.
 func SubjectDigests(ctx context.Context, obj objects.TektonObject) []intoto.Subject {
+	var subjects []intoto.Subject
+
+	switch obj.GetObject().(type) {
+	case *v1beta1.PipelineRun:
```
lcarva, line 46: Consider adding new methods to the TektonObject interface to avoid having to treat each different implementation differently. For example, you may want to add a `TaskRuns()` method that returns a list of the TaskRuns with the object.

pkg/chains/formats/slsa/extract/extract_test.go
```diff
@@ -144,23 +144,22 @@ func createTaskRunObjectWithResults(results map[string]string) objects.TektonObj
...
+						{Name: fmt.Sprintf("%v_IMAGE_URL", prefix), Value: *v1beta1.NewStructuredValues(url)},
+					},
+				},
+			},
+		}
+		pro.AppendTaskRun(tr)
 		prefix++
 	}
 
-	return objects.NewPipelineRunObject(
```
lcarva, line 157: Let's please not drop support for PipelineRun results.

pkg/chains/objects/objects.go
```diff
@@ -135,8 +135,8 @@ func (tro *TaskRunObject) GetPullSecrets() []string {
 type PipelineRunObject struct {
 	// The base PipelineRun
 	*v1beta1.PipelineRun
-	// taskRuns that were apart of this PipelineRun
-	taskRuns []*v1beta1.TaskRun
+	// TaskRuns that were apart of this PipelineRun
+	TaskRuns []*v1beta1.TaskRun
```
lcarva, line 139: I think we should avoid this. If we want `TaskRuns` to be accessible, let's make that a method on the interface.

lcarva (changes requested):

pkg/chains/formats/slsa/extract/extract.go
```diff
@@ -34,12 +34,67 @@ import (
...
+
+	pr := obj.(*objects.PipelineRunObject)
+
+	for _, tr := range pr.TaskRuns {
+		// collect subjects from individual TaskRuns
+		trSubjects := subjectsFromTaskrun(ctx, objects.NewTaskRunObject(tr))
+
+		for _, s := range trSubjects {
+			if v, ok := subjects[s.Name]; ok {
+				mergeMaps(subjects[s.Name], v)
```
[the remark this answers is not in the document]
lcarva: Same URI and same digest can be safely combined into a single entry. Same URI with different digests should be separate entries.

pkg/chains/formats/slsa/extract/extract.go
chuangw6: Do we want to support both pipelineRun results and taskrun results by default, OR do we want to make it a configmap field to let users select how Chains should work?

The latter sounds a bit better to me b/c I feel like the former might be a bit messy if users pull in catalog tasks that produce type hinting results, and users also define some overlapping type hinting results on the pipeline level.

pkg/chains/objects/objects.go
chuangw6: SGTM!

pkg/chains/formats/slsa/extract/extract.go
chuangw6: SGTM!

chuangw6: Let me open a separate PR to address this. Since this would help us clean up some other parts where we need to access child taskruns ([example](https://github.com/tektoncd/chains/blob/5230749f22aaa1a81305a2bf8dd7616614609d66/pkg/chains/formats/slsa/v2alpha2/internal/resolved_dependencies/resolved_dependencies.go#L181-L186)).

pkg/chains/formats/slsa/extract/extract.go
lcarva: There are benefits in using the results from the PipelineRun. For example, one can use a Task that doesn't produce the expected Chains results (many of the Tasks in the tekton catalog don't).

We could make it configurable as long as the default includes the PipelineRun results so we don't break users. I don't really think making it configurable is necessary, but I don't have a strong opinion on this. More configuration options could make things more complicated for users though.

chuangw6: Thanks @lcarva for sharing the thoughts!

> For example, one can use a Task that doesn't produce the expected Chains results (many of the Tasks in the tekton catalog don't).

That's a good point. I have thought about same thing in some catalog tasks. For example, current [git clone task](https://github.com/tektoncd/catalog/tree/main/task/git-clone/0.9#results) do output the repo uri and commit sha results, but it's not in type hinting format, whereas [kaniko task](https://github.com/tektoncd/catalog/tree/main/task/kaniko/0.6#results) fully supports this. 

In those cases, I think we should make sure at least some catalog tasks produce type hinting results for consistency and better user experience, instead of letting users to figure out what to do if some catalog tasks do not produce type hinting.

That said, I am not opposing the idea of keeping support of pipeline results and I am happy to keep supporting it. But at least for some catalog tasks, we should be consistent and make sure they produce results in type hinting way. 



> We could make it configurable as long as the default includes the PipelineRun results so we don't break users. I don't really think making it configurable is necessary, but I don't have a strong opinion on this. More configuration options could make things more complicated for users though.

Agreed. ConfigMap option seems to expose too much implementation details to users. But the risk is the messy state if type hinting exist on both taskrun and pipelinerun level. (i.e. author surfaces taskruns' type hinting results to pipelines' type hinting results).

Open to other thoughts and ideas here.

lcarva: Update from today's meeting: Let's make it configurable for now with the goal of making it not configurable in the future. This will allow for a safer feature rollout and give a chance for users to adapt. Since it's temporary, it should not increase maintenance long-term.

pkg/chains/objects/objects.go
chuangw6: Done! PTAL https://github.com/tektoncd/chains/pull/879

pkg/chains/formats/slsa/extract/extract.go
chuangw6: I'll close this comment and Please review https://github.com/tektoncd/chains/pull/879 instead. Thanks

pkg/chains/formats/slsa/extract/extract.go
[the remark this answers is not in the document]
chuangw6: @lcarva That makes sense. Done. Please take a look!

pkg/chains/formats/slsa/extract/extract.go
chuangw6: @lcarva Done. Please take a look!

pkg/chains/formats/slsa/extract/extract_test.go
chuangw6: @lcarva Done!

pkg/chains/formats/slsa/extract/extract.go
```diff
@@ -39,7 +40,92 @@ import (
...
 //   - the `*_URL` or `*_URI` fields cannot be empty.
-func SubjectDigests(ctx context.Context, obj objects.TektonObject) []intoto.Subject {
+func SubjectDigests(ctx context.Context, obj objects.TektonObject, slsaconfig *slsaconfig.SlsaConfig) []intoto.Subject {
+	var subjects []intoto.Subject
+
+	switch obj.GetObject().(type) {
+	case *v1beta1.PipelineRun:
+		subjects = subjectsFromPipelineRun(ctx, obj, slsaconfig)
+	case *v1beta1.TaskRun:
+		subjects = subjectsFromTektonObject(ctx, obj)
```
chitrangpatel, line 50: Why not call it `subjectsFromTaskRun`? `Tekton Object` refers to both task and pipeline runs.

chuangw6: subjectsFromPipelineRun also passes PipelineRunObject to subjectsFromTektonObject if `pr` mode is configured.

pkg/chains/formats/slsa/extract/extract.go
```diff
@@ -39,7 +40,94 @@ import (
...
+
+	return subjects
+}
+
+func subjectsFromPipelineRun(ctx context.Context, obj objects.TektonObject, slsaconfig *slsaconfig.SlsaConfig) []intoto.Subject {
+	logger := logging.FromContext(ctx)
+	// If the configured input/output observation mode is pipeline level, then
+	// call the generic function to parse the subject.
+	if slsaconfig.PrObserveMode == "pr" {
+		return subjectsFromTektonObject(ctx, obj)
```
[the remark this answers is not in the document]
chuangw6: Done Thanks!

pkg/chains/formats/slsa/extract/extract.go
```diff
@@ -39,7 +40,94 @@ import (
...
+}
+
+// two subjects are equal if and only if they have same name and have at least
+// one common algorithm and hex value.
+func subjectEqual(x, y intoto.Subject) bool {
+	if x.Name != y.Name {
+		return false
+	}
+	for algo, hex := range x.Digest {
+		if v, ok := y.Digest[algo]; ok && v == hex {
```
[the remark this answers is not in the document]
chuangw6: Done Thanks!

docs/config.md
```diff
@@ -64,9 +64,11 @@ Supported keys include:
 | `artifacts.pipelinerun.format` | The format to store `PipelineRun` payloads in. | `in-toto`, `slsa/v1`| `in-toto` |
 | `artifacts.pipelinerun.storage` | The storage backend to store `PipelineRun` signatures in. Multiple backends can be specified with comma-separated list ("tekton,oci"). To disable the `PipelineRun` artifact input an empty string ("").  | `tekton`, `oci`, `gcs`, `docdb`, `grafeas` | `tekton` |
 | `artifacts.pipelinerun.signer` | The signature backend to sign `PipelineRun` payloads with. | `x509`, `kms` | `x509` |
+| `artifacts.pipelinerun.observe-mode` | The way that Chains observes inputs & outputs of a PipelineRun. The default option `pr` configures Chains to only inspect Pipeline level params/results, whereas the option `tr` configures Chains to dive into child TaskRuns. | `pr`, `tr` | `pr` |
```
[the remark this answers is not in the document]
chuangw6: SGTM. Changed as suggested. PTAL! Thanks!

docs/config.md
```diff
@@ -64,9 +64,11 @@ Supported keys include:
 | `artifacts.pipelinerun.format` | The format to store `PipelineRun` payloads in. | `in-toto`, `slsa/v1`| `in-toto` |
 | `artifacts.pipelinerun.storage` | The storage backend to store `PipelineRun` signatures in. Multiple backends can be specified with comma-separated list ("tekton,oci"). To disable the `PipelineRun` artifact input an empty string ("").  | `tekton`, `oci`, `gcs`, `docdb`, `grafeas` | `tekton` |
 | `artifacts.pipelinerun.signer` | The signature backend to sign `PipelineRun` payloads with. | `x509`, `kms` | `x509` |
+| `artifacts.pipelinerun.enable-deep-inspection` | This boolean option will configure whether Chains should inspect child taskruns for captureing inputs/outputs within a pipelinerun. `false` means that Chains only checks pipeline level results, whereas `true` means Chains inspects both pipeline level and task level results. | `true`, `false` | `false` |
```
lcarva, line 67: `s/captureing/capturing`

lcarva, line 67: Although it's a boolean, it's actually a string in the ConfigMap, right? Maybe we can say `"false"` and `"true"` instead of `false` and `true`?

lcarva (approved):

docs/config.md
chuangw6: Oops. Thanks for catching it!

chuangw6:
Sorry made a [minor change](https://github.com/tektoncd/chains/compare/e690eace07d33ac1d06bc7fcf7aa901089dadc3e..4b8acd4ca942557864b54470e624f8157e03a7b1) (unit test name)

@lcarva @chitrangpatel

PR #879, opened 2023-07-21
Make child TaskRuns accessible

chuangw6:
# Changes

/kind cleanup

Prior, the child TaskRuns of a PipelineRunObject are not accessible directly. Only way to access it is to loop over task name from pipeline spec and use GetTaskRunFromTask method to get the taskrun object.

Now, we introduce a method in the Object interface to expose the slice of child taskruns directly so that we can easily access all child taskruns, though the old GetTaskRunFromTask is still useful in one case i.e. 0.2 buildConfig for which we do need to loop over PipelineTask type and call it to get taskrun.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/chains/formats/slsa/internal/compare/compareSLSA.go
+++ b/pkg/chains/formats/slsa/internal/compare/compareSLSA.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2023 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package compare
+
+import (
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/in-toto/in-toto-golang/in_toto"
+	"github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/common"
+	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1"
+)
+
+// SLSAV1CompareOptions returns the comparison options for sorting some slice fields in
+// SLSA v1 statement including ResourceDescriptor and Subject.
+func SLSAV1CompareOptions() []cmp.Option {
+	// checking content + uri + digest should be sufficient here based on the fact that
+	// a ResourceDescriptor MUST specify one of uri, digest or content at a minimum.
+	// Source: https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md#fields
+	resourceDescriptorSort := func(x, y slsa.ResourceDescriptor) bool {
+		if string(x.Content) != string(y.Content) {
+			return string(x.Content) < string(y.Content)
+		}
+		if x.URI != y.URI {
+			return x.URI < y.URI
+		}
+		return lessDigestSet(x.Digest, y.Digest)
+	}
+
+	subjectSort := func(x, y in_toto.Subject) bool {
+		if x.Name != y.Name {
+			return x.Name < y.Name
+		}
+		return lessDigestSet(x.Digest, y.Digest)
+	}
+
+	return []cmp.Option{
+		cmpopts.SortSlices(resourceDescriptorSort),
+		cmpopts.SortSlices(subjectSort),
+	}
+}
+
+// MaterialsCompareOption returns the comparison option to sort and compare a
+// list of Materials.
+func MaterialsCompareOption() cmp.Option {
+	materialsSort := func(x, y common.ProvenanceMaterial) bool {
+		if x.URI != y.URI {
+			return x.URI < y.URI
+		}
+		return lessDigestSet(x.Digest, y.Digest)
+	}
+	return cmpopts.SortSlices(materialsSort)
+}
+
+func lessDigestSet(x, y common.DigestSet) bool {
+	for algo, digestX := range x {
+		digestY, ok := y[algo]
+		if !ok {
+			// Algorithm not present in y, x is considered greater.
+			return false
+		}
+		// Compare the digests lexicographically.
+		if digestX != digestY {
+			return digestX < digestY
+		}
+		// The digests are equal, check the next algorithm.
+	}
+
+	// All algorithms in x have corresponding entries in y, so check if y has more algorithms.
+	return len(x) < len(y)
+}
--- a/pkg/chains/formats/slsa/internal/material/material.go
+++ b/pkg/chains/formats/slsa/internal/material/material.go
@@ -28,7 +28,6 @@ import (
 	"github.com/tektoncd/chains/pkg/chains/formats/slsa/attest"
 	"github.com/tektoncd/chains/pkg/chains/objects"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
-	"knative.dev/pkg/logging"
 )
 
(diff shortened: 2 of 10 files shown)


chuangw6:
cc @lcarva @chitrangpatel 
PTAL! Thanks!

pkg/chains/formats/slsa/internal/material/material.go
```diff
@@ -77,38 +75,28 @@ func PipelineMaterials(ctx context.Context, pro *objects.PipelineRunObject) ([]c
...
-	pSpec := pro.Status.PipelineSpec
-	if pSpec != nil {
-		pipelineTasks := append(pSpec.Tasks, pSpec.Finally...)
-		for _, t := range pipelineTasks {
-			tr := pro.GetTaskRunFromTask(t.Name)
-			// Ignore Tasks that did not execute during the PipelineRun.
-			if tr == nil || tr.Status.CompletionTime == nil {
-				logger.Infof("taskrun status not found for task %s", t.Name)
-				continue
-			}
```
wlynch, line 89: I think the previous way is safer to do because it lets you catch cases where the Tasks failed to load - the PipelineRun remains the source of truth of what TaskRuns that should be present.

We defer loading Tasks so we don't frontload fetches if there's no work to do, but this also means that a PipelineRunObject may not have a fully produced result for GetTaskRuns() depending on how you call it, and a caller wouldn't know whether this is complete or not.

e.g. how would you detect something like:

```go
o := NewPipelineRunObject(pr)
o.GetTaskRuns() // This is empty
```

We may want to keep things as they were before - IIUC, this already has the same data?

wlynch (changes requested):

pkg/chains/formats/slsa/internal/material/material.go
chuangw6: I think the old code has same results as the new change, but the new code has cleaner logic.

The formatting code will not be invoked until [controller logic](https://github.com/tektoncd/chains/blob/ddf41fbba572a313f9f21c3549c0fa0e04f91195/pkg/reconciler/pipelinerun/pipelinerun.go#L95-L119) observes all runs finished successfully, and fetches child taskruns and registers it under the pipelinerun.

chuangw6: friendly ping for any more thoughts? @wlynch Thanks!

pkg/chains/formats/slsa/internal/material/material.go
```diff
@@ -77,38 +75,28 @@ func PipelineMaterials(ctx context.Context, pro *objects.PipelineRunObject) ([]c
...
 
-			stepMaterials, err := FromStepImages(tr.Status.Steps)
-			if err != nil {
-				return mats, err
-			}
-			mats = append(mats, stepMaterials...)
+	for _, tr := range pro.GetTaskRuns() {
```
chitrangpatel, line 79: May be we do that here?

chitrangpatel:
```go
if tr == nil || tr.Status.CompletionTime == nil {
        logger.Infof("taskrun status not found for task %s", t.Name)
        continue
}
```

chitrangpatel: or may be move this logic in `pro.GetTaskRuns` and only return the tasks that are `"good"`?

chitrangpatel: and call it `GetCompletedTaskRuns()`?

pkg/chains/formats/slsa/internal/material/material.go
chuangw6: [Controller logic](https://github.com/tektoncd/chains/blob/ddf41fbba572a313f9f21c3549c0fa0e04f91195/pkg/reconciler/pipelinerun/pipelinerun.go#L106-L119) already make sure that tr and completion time are not nil before registering the trs to the pipeline.

chitrangpatel: Ah ok! I missed it. So we had this redundant check that should have been removed in the first place?

chitrangpatel: In that case, feel free to resolve my other comments without any action.

pkg/chains/formats/slsa/internal/compare/compareSLSA.go
```diff
@@ -0,0 +1,80 @@
...
+		return lessDigestSet(x.Digest, y.Digest)
+	}
+
+	return []cmp.Option{
+		cmpopts.SortSlices(resourceDescriptorSort),
+		cmpopts.SortSlices(subjectSort),
+	}
+}
+
+func MaterialsCompareOptions() cmp.Option {
```
chitrangpatel, line 54: May be add some documentation since its an exported function?

chuangw6: Done Thanks!

pkg/chains/formats/slsa/internal/material/material.go
chuangw6: > Ah ok! I missed it. So we had this redundant check that should have been removed in the first place?

Yeah, exactly!

pkg/chains/formats/slsa/internal/material/material.go
chuangw6: closed as it's resolved in https://github.com/tektoncd/chains/pull/879#discussion_r1271132211

pkg/chains/objects/objects.go
```diff
@@ -197,6 +197,11 @@ func (pro *PipelineRunObject) AppendTaskRun(tr *v1beta1.TaskRun) {
 	pro.taskRuns = append(pro.taskRuns, tr)
 }
 
+// Get all child TaskRuns spawned from the PipelineRun
+func (pro *PipelineRunObject) GetTaskRuns() []*v1beta1.TaskRun {
+	return pro.taskRuns
```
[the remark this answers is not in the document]
chuangw6: closed as it's resolved in https://github.com/tektoncd/chains/pull/879#discussion_r1271132211

pkg/chains/formats/slsa/internal/material/material.go
wlynch: While the code results in the same behavior for Chains overall, it's brittle in the sense that it requires callers to need to pre-populate the TaskRuns, or else they can get wrong data.

My concern is this isn't obvious from an outside point-of-view, and it makes it easy to mistakenly fetch partial/no data without realizing - I'd want something like this to error out if the Object wasn't in the correct state.

I think there's a few options to address this:

1. Revert to the old behavior, where we use the PR status as the source of truth (because then it's obvious there's something wrong if there's a mismatch).
2. Wrap Task fetching in a [sync.Once](https://pkg.go.dev/sync#Once) and lazily fetch it whenever a client tries to read the data, which will guarantee that the data is always fetched (or else error).

chuangw6: I feel like there might be some misalignment on the old behaviour.

The old behaviour doesn't not fetch TaskRuns as needed using PR status as source of truth. Instead it also assumes the caller pre-populates the TaskRuns, which is reflected on line 84 in which GetTaskRunFromTask just searches the TaskRun object based on the name from the pre-populated `PipelineRunObject.taskruns` list.


```go
   tr := pro.GetTaskRunFromTask(t.Name)
```

chuangw6: Spoke with @wlynch offline. TL;DR: we will keep the existing behaviour and close this PR.

The good thing about the old behaviour is that it can cross check what's reported in the PR status and what's populated to the `PipelineRunObject.taskruns` list, and reports/logs if TaskRuns are not populated, which is important when others use Chains library to reproduce the provenance generation.

chuangw6:
/close

https://github.com/tektoncd/chains/pull/879#discussion_r1275502875

chuangw6:
extract the compare options to the new PR https://github.com/tektoncd/chains/pull/887.

PR #885, opened 2023-07-25
Pass slsa-related config via explicit struct instead of separate args

chuangw6:
# Changes

/cleanup

Prior, we received slsa-related configs from configmap and passed them
via separate arguments down to GenerateAttestation.

Now, we pass them via a struct as a whole instead of separate args. This
will enable slsa formatters to be extensible to receive more slsa-related
configs without having to change downstream functions' interface i.e.
config option [`ObserveMode`](https://github.com/tektoncd/chains/pull/866#discussion_r1269715883)
which customizes how Chains observes inputs/outputs.

@chuangw6

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/chains/formats/slsa/internal/slsaconfig/slsaconfig.go
+++ b/pkg/chains/formats/slsa/internal/slsaconfig/slsaconfig.go
@@ -1,0 +1,21 @@
+/*
+Copyright 2023 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package slsaconfig
+
+// SlsaConfig carries common information that is needed across different SLSA formatters.
+type SlsaConfig struct {
+	// BuilderID is the URI of the trusted build platform.
+	BuilderID string
+}
--- a/pkg/chains/formats/slsa/v1/intotoite6.go
+++ b/pkg/chains/formats/slsa/v1/intotoite6.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/tektoncd/chains/pkg/chains/formats"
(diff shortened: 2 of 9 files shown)


chuangw6:
Note: this change currently only applies to slsa v1 provenance. Will make changes to other slsa formatters after collecting some feedback on this approach.

chuangw6:
Looking for feedback and inputs. Thanks


cc @chitrangpatel @lcarva @wlynch

chitrangpatel:
This is great! I think we should also implement this for other formats `v1 & v2alpha1`.

wlynch (dismissed):
I would recommend against doing this. Context values are tricky because it becomes less-obvious that certain values need to be set.

My general rule of thumb is that callers should be able to operate normally if context values are not present, but here I think BuilderID is critical.

I'd recommend taking a look at the v1 interface proposal, particularly [Payloader](https://github.com/tektoncd/chains/blob/main/docs/v1-proposal.md#payloaders). I think this will do what you want, but with an explicit struct field instead.

chuangw6:
Thank you for the feedback on the context! Concur that context is a less good solution.

> > be extensible to receive more slsa-related configs without having to change downstream functions' interface i.e. config option https://github.com/tektoncd/chains/pull/866#discussion_r1269715883 which customizes how Chains observes inputs/outputs.
>
> An alternative to achieve that benefit is to make GenerateAttestation functions take a simple struct, something like a SLSAConfig.

Yeah, this is the alternative that I thought before other than context -  passing the explicit struct. (worth noting that unlike context, if we need to pass down the configs beyond GenerateAttestation i.e. to subjects / materials, we need to change their interface to accept the struct as well. But I think it's fine given that context is not best solution.) 

Just changed the code to the struct solution. PTAL.


> I'd recommend taking a look at the v1 interface proposal, particularly [Payloader](https://github.com/tektoncd/chains/blob/main/docs/v1-proposal.md#payloaders). I think this will do what you want, but with an explicit struct field instead.

@wlynch is the solution mentioned above what you mean here?

chuangw6:
The new change is ready for review. PTAL @chitrangpatel @wlynch.

Thanks!

chitrangpatel:
/approve

I'll let @wlynch lgtm since he had comments regarding the original PR.

wlynch:
> @wlynch is the solution mentioned above what you mean here?

No, but it's probably going to change soon so 🤷

chuangw6:
Looks like the PR merge is blocked by the requested changes. Maybe need approval as well? @wlynch

PR #896, opened 2023-08-08
Increase direct usage of Tekton Object Interface

chitrangpatel:
This PR further increases the usage of the Tekton object interface in our codebase. It adds functions to get images from steps and sidecars and also extracts the `RefSource` from `Status.Provenance`.
 

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Internally increase direct usage of Tekton Object Interface 
```
/kind cleanup


pkg/chains/formats/slsa/internal/material/material_test.go
```diff
@@ -400,6 +400,7 @@ func TestStructuredResultPipelineMaterials(t *testing.T) {
 	}
 }
 
+/*
```
lcarva, line 403: Let's either update the tests or remove them.

lcarva (commented):
Nice!

pkg/chains/formats/slsa/internal/material/material_test.go
chitrangpatel: Done

pkg/chains/objects/objects.go
```diff
@@ -40,12 +41,39 @@ type Object interface {
 	runtime.Object
 }
 
+type ParamValue struct {
```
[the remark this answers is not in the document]
chitrangpatel: Thanks! I will look into this.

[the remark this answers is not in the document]
chitrangpatel: I removed this out of this PR. I think I will make it a part of the PR that switches usage from v1beta1 to v1.

pkg/chains/formats/slsa/attest/attest.go
```diff
@@ -58,10 +59,11 @@ func Step(step *v1beta1.Step, stepState *v1beta1.StepState) StepAttestation {
 	return attestation
 }
 
-func Invocation(source *v1beta1.RefSource, params []v1beta1.Param, paramSpecs []v1beta1.ParamSpec, meta metav1.Object) slsa.ProvenanceInvocation {
+func Invocation(obj objects.TektonObject, params []v1beta1.Param, paramSpecs []v1beta1.ParamSpec, meta metav1.Object) slsa.ProvenanceInvocation {
```
chuangw6, line 62: Seems like there are some improvement area for this function interface though I understand that we do need to separate out params and paramSpec to accept both task level and pipeline level. 

Maybe remove `meta metav1.Object` b/c `obj` already contains meta data right?

chitrangpatel: Yes. Fixed. Thanks!

pkg/chains/formats/slsa/attest/attest.go
```diff
@@ -77,10 +79,18 @@ func Invocation(source *v1beta1.RefSource, params []v1beta1.Param, paramSpecs []
...
 	annotations := map[string]string{}
+	var meta metav1.Object
+	switch o := obj.GetObject().(type) {
+	// nolint:staticcheck
+	case *v1beta1.PipelineRun:
+		meta = o.GetObjectMeta()
+	// nolint:staticcheck
+	case *v1beta1.TaskRun:
+		meta = o.GetObjectMeta()
+	}
```
chuangw6, line 93: I think this is not needed. As it's interface, we can just `obj.GetAnnotations()` and `obj.GetLabels()`.

chitrangpatel: Good catch! I fixed it. Thanks

pkg/chains/objects/objects.go
```diff
@@ -58,6 +64,7 @@ type TektonObject interface {
 	GetLatestAnnotations(ctx context.Context, clientSet versioned.Interface) (map[string]string, error)
 	Patch(ctx context.Context, clientSet versioned.Interface, patchBytes []byte) error
 	GetResults() []Result
+	GetRefSource() *RefSource
```
wlynch, line 67: Prefer limiting things in this interface to the common denominator - in this case, the provenance field would be more fitting.

Both TaskRuns/PipelineRuns use the same type (and the same underlying RefSource type) so I don't think we'd need to define our own fork type here.

wlynch (changes requested):
Almost there! PR comment probably needs an update. 🙏

pkg/chains/objects/objects.go
chitrangpatel: fixed!

pkg/chains/objects/objects.go
```diff
@@ -58,6 +58,7 @@ type TektonObject interface {
 	GetLatestAnnotations(ctx context.Context, clientSet versioned.Interface) (map[string]string, error)
 	Patch(ctx context.Context, clientSet versioned.Interface, patchBytes []byte) error
 	GetResults() []Result
+	GetRefSource() *v1beta1.RefSource
```
wlynch, line 61:
```suggestion
	GetProvenance() *v1beta1.Provenance
```

pkg/chains/objects/objects.go
wlynch: I think we can go further - provenance feels like a common enough thing that we'll use for both pipelineruns and taskruns that we should return the type itself rather than make individual methods for its sub-fields.

Added suggestions!

wlynch (changes requested):
Almost there!

pkg/chains/objects/objects.go
chitrangpatel: Done!

wlynch (commented):
/lgtm

wlynch (approved):

PR #911, opened 2023-08-30
 Make e2e tests faster 

chitrangpatel:
Instead of waiting for the secrets and configmap to propagate to the pod which could take [~2 mins](https://ahmet.im/blog/kubernetes-secret-volumes-delay/), we restart the pod and force this update since the pod restart takes only a few seconds. This in turn, reduces the time per test from ~1.5 mins down to a few seconds.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```
/kind cleanup

### DIFF
--- a/test/test_utils.go
+++ b/test/test_utils.go
@@ -37,8 +37,11 @@ import (
 	"github.com/tektoncd/chains/pkg/test/tekton"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	pipelineclientset "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 )
 
 func getTr(ctx context.Context, t *testing.T, c pipelineclientset.Interface, name, ns string) (tr *v1beta1.TaskRun) {
@@ -191,7 +194,10 @@ func setConfigMap(ctx context.Context, t *testing.T, c *clients, data map[string
 	if err != nil {
 		t.Fatal(err)
 	}
-	time.Sleep(30 * time.Second) // https://github.com/tektoncd/chains/issues/664
+	err = restartChainsControllerPod(ctx, c.KubeClient, 300*time.Second)
+	if err != nil {
+		t.Fatalf("Failed to restart the pod: %v", err)
+	}
 
 	return func() {
 		for k := range data {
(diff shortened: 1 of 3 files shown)


test/test_utils.go
```diff
@@ -281,3 +287,30 @@ func verifySignature(ctx context.Context, t *testing.T, c *clients, obj objects.
 		}
 	}
 }
+
+// restartChainsControllerPod restarts the pod running Chains
+// it then waits for a given timeout for the pod to resume running state
+func restartChainsControllerPod(c kubernetes.Interface, timeout time.Duration) error {
```
lcarva, line 293: Should this take a context? It looks like the caller, setConfigMap, already has one to pass it in.

test/test_utils.go
```diff
@@ -281,3 +287,30 @@ func verifySignature(ctx context.Context, t *testing.T, c *clients, obj objects.
...
+		return err
+	}
+
+	return wait.PollUntilContextTimeout(context.Background(), 2*time.Second, timeout, true, func(context.Context) (done bool, err error) {
+		pods, err := c.CoreV1().Pods("tekton-chains").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			return false, err
+		}
+		pod := pods.Items[0]
+		if pod.UID != podUid {
```
wlynch, line 311: I don't think this will ever return true - a newly created pod should always have a new UID.

In fact, you may want to check the opposite - if it is the same UID then it means that we're still on the existing pod and we're not done waiting yet.

test/test_utils.go
```diff
@@ -281,3 +287,30 @@ func verifySignature(ctx context.Context, t *testing.T, c *clients, obj objects.
...
+// restartChainsControllerPod restarts the pod running Chains
+// it then waits for a given timeout for the pod to resume running state
+func restartChainsControllerPod(c kubernetes.Interface, timeout time.Duration) error {
+	pods, err := c.CoreV1().Pods("tekton-chains").List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	pod := pods.Items[0]
+	podUid := pod.UID
+	err = c.CoreV1().Pods(pod.Namespace).Delete(context.Background(), pod.Name, metav1.DeleteOptions{})
```
wlynch, line 300: May want to set `GracePeriodSeconds: 0` to speed up the delete.

test/test_utils.go
```diff
@@ -281,3 +287,30 @@ func verifySignature(ctx context.Context, t *testing.T, c *clients, obj objects.
...
 }
+
+// restartChainsControllerPod restarts the pod running Chains
+// it then waits for a given timeout for the pod to resume running state
+func restartChainsControllerPod(c kubernetes.Interface, timeout time.Duration) error {
+	pods, err := c.CoreV1().Pods("tekton-chains").List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	pod := pods.Items[0]
```
wlynch, line 298: This is brittle - there's no guarantee the controller is the first pod.

You'll want to verify the name / use a label selector (e.g. `app.kubernetes.io/component=controller`)

wlynch (changes requested):

test/test_utils.go
chitrangpatel: Fixed

test/test_utils.go
chitrangpatel: Fixed

test/test_utils.go
chitrangpatel: Done!

test/test_utils.go
chitrangpatel: Good point! Done.

wlynch (commented):
/lgtm

test/test_utils.go
```diff
@@ -281,3 +287,33 @@ func verifySignature(ctx context.Context, t *testing.T, c *clients, obj objects.
...
+// restartChainsControllerPod restarts the pod running Chains
+// it then waits for a given timeout for the pod to resume running state
+func restartChainsControllerPod(ctx context.Context, c kubernetes.Interface, timeout time.Duration) error {
+	pods, err := c.CoreV1().Pods("tekton-chains").List(ctx, metav1.ListOptions{LabelSelector: "app.kubernetes.io/component=controller"})
+	if err != nil {
+		return err
+	}
+	pod := pods.Items[0]
+	podUid := pod.UID
+	gradePeriodSeconds := int64(0)
```
PuneetPunamiya, line 300:
```suggestion
	gracePeriodSeconds := int64(0)
```

Not sure if this is a nit change 😄

chitrangpatel: Fixed 😄

lcarva (approved):
/lgtm

wlynch (approved):

PR #933, opened 2023-09-22
Fixes Hashivault configuration with Chains

PuneetPunamiya:
* Initially when kms was misconfigured, then the chains controller pod was getting panic and ended up getting crashed

* Hence, this patch checks it the vault address is valid and if it is not valid, then returns the error which handles the pod from getting panic and crashed

* Example of Logs of pod getting crashed

```bash=
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1b43574]

goroutine 173 [running]:
github.com/tektoncd/chains/vendor/github.com/jellydator/ttlcache/v3.(*Item[...]).Value(0xc000513a00?)
	github.com/tektoncd/chains/vendor/github.com/jellydator/ttlcache/v3/item.go:110 +0x34
github.com/tektoncd/chains/vendor/github.com/sigstore/sigstore/pkg/signature/kms/hashivault.(*hashivaultClient).public(0xc0003be480)
	github.com/tektoncd/chains/vendor/github.com/sigstore/sigstore/pkg/signature/kms/hashivault/client.go:251 +0xe9
github.com/tektoncd/chains/vendor/github.com/sigstore/sigstore/pkg/signature/kms/hashivault.SignerVerifier.PublicKey(...)
	github.com/tektoncd/chains/vendor/github.com/sigstore/sigstore/pkg/signature/kms/hashivault/signer.go:140
github.com/tektoncd/chains/pkg/chains/signing.Wrap({0x2af1440?, 0xc000578db0?}, {0x37567c8, 0xc001110b90})
	github.com/tektoncd/chains/pkg/chains/signing/wrap.go:32 +0x46
github.com/tektoncd/chains/pkg/chains.(*ObjectSigner).Sign(0xc000d35110, {0x3750aa0, 0xc001032f90}, {0x377ac40, 0xc0010984f8})
	github.com/tektoncd/chains/pkg/chains/signing.go:153 +0x916
github.com/tektoncd/chains/pkg/reconciler/taskrun.(*Reconciler).FinalizeKind(0xc000611f40, {0x3750aa0, 0xc001032f90}, 0xc000af0f00)
	github.com/tektoncd/chains/pkg/reconciler/taskrun/taskrun.go:67 +0x1e2
github.com/tektoncd/chains/pkg/reconciler/taskrun.(*Reconciler).ReconcileKind(0xc000f7ef80?, {0x3750aa0?, 0xc001032f90?}, 0x11?)
[...]

PR #936, opened 2023-10-09
Fixes hashivault connection check

PuneetPunamiya:
* Initially in the https://github.com/tektoncd/chains/pull/933 it was checking if the vault address is valid or not, but if the url did not have port then it used to return an error and could not proceed further for signing

* Hence this patch fixes this by checking if the url has port or if the url is a http or https url and then validates if the url is valid so that it can proceed for signing

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/chains/signing/kms/kms.go
+++ b/pkg/chains/signing/kms/kms.go
@@ -17,6 +17,7 @@ package kms
 import (
 	"context"
 	"crypto"
+	"fmt"
 	"net"
 	"net/url"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/chains/signing/kms/kms.go
```diff
@@ -51,11 +52,35 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
...
+				Host:   vaultAddress.Host + ":443",
+			}
+		case vaultAddress.Scheme == "":
+			vaultUrl = &url.URL{
+				Scheme: "http",
+				Host:   cfg.Auth.Address + ":80",
+			}
+		}
+
+		if vaultUrl != nil {
```
lcarva, line 76: Shouldn't we return an error if it is `nil`?

PuneetPunamiya: as the error will be nil right ?

lcarva: Right. You want to create a new error in this case.

pkg/chains/signing/kms/kms.go
```diff
@@ -51,11 +52,35 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
 			return nil, err
 		}
 
-		conn, err := net.DialTimeout("tcp", vaultAddress.Host, 5*time.Second)
-		if err != nil {
-			return nil, err
+		var vaultUrl *url.URL
+		switch {
```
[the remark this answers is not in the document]
vdemeester: Can't we rely on a default scheme ? (`https://` or something ?)

vdemeester (approved):

pkg/chains/signing/kms/kms.go
```diff
@@ -51,11 +52,45 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
...
+			}
+			if vaultUrl.Port() == "" {
+				vaultUrl.Host = cfg.Auth.Address + ":80"
+			}
+		}
+
+		if vaultUrl != nil {
+			conn, err := net.DialTimeout("tcp", vaultUrl.Host, 5*time.Second)
+			if err != nil {
+				fmt.Printf("Error connecting to URL %s: %v\n", vaultUrl, err)
```
lcarva, line 87: Let's not print here.

pkg/chains/signing/kms/kms.go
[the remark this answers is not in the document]
PuneetPunamiya: Won't this case cover this scenario https://github.com/tektoncd/chains/pull/936/files#diff-63cec1a1260227d38ef0e60ce232822a3f65f580869912386619e2c8d66ce191R74-R82 🤔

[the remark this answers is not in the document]
lcarva: Yup. Missed that.

PR #971, opened 2023-10-26
feat: add PipelineRun support to storage/gcs

### LINKED ISSUE #852: storage/gcs: PipelineRun support
### Feature request

When working on https://github.com/tektoncd/chains/pull/851 I noticed that GCS client doesn't support PipelineRuns. It should be fairly easy to add, but filing an issue to track.

### Use case

Parity with other storage interfaces.

aaron-prindle:
fixes #852 

Prior to this commit the storage/gcs client didn't support PipelineRuns, only TaskRuns which is not in parity with our other storage interfaces.  This PR adds PipelineRuns support to storage/gcs.

`//nolint:staticcheck` used similar to elsewhere in code as otherwise golangci-lint reports usage of deprecated v1beta1 APIs, snippet below:
```
  Error: SA1019: v1beta1.TaskRun is deprecated: Please use v1.TaskRun instead.  (staticcheck)
  Error: SA1019: v1beta1.PipelineRun is deprecated: Please use v1.PipelineRun instead.  (staticcheck)
  ...
  ```

NOTE: This is my first tektoncd/chains PR.  If I misunderstood anything about the issue, changes necessary, best practices, etc. I'm happy to iterate as needed.  Thanks!

### DIFF
--- a/pkg/chains/storage/gcs/gcs.go
+++ b/pkg/chains/storage/gcs/gcs.go
@@ -32,8 +32,11 @@ import (
 const (
 	StorageBackendGCS = "gcs"
 	// taskrun-$namespace-$name/$key.<type>
-	SignatureNameFormat = "taskrun-%s-%s/%s.signature"
-	PayloadNameFormat   = "taskrun-%s-%s/%s.payload"
+	SignatureNameFormatTaskRun = "taskrun-%s-%s/%s.signature"
+	PayloadNameFormatTaskRun   = "taskrun-%s-%s/%s.payload"
+	// pipelinerun-$namespace-$name/$key.<type>
+	SignatureNameFormatPipelineRun = "pipelinerun-%s-%s/%s.signature"
+	PayloadNameFormatPipelineRun   = "pipelinerun-%s-%s/%s.payload"
 )
 
 // Backend is a storage backend that stores signed payloads in the TaskRun metadata as an annotation.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/chains/storage/gcs/gcs.go [hunk 1]
```diff
@@ -163,20 +206,32 @@ func (b *Backend) retrieveObject(ctx context.Context, object string) (string, er
...
-func payloadName(tr *v1beta1.TaskRun, opts config.StorageOpts) string {
-	return fmt.Sprintf(PayloadNameFormat, tr.Namespace, tr.Name, opts.ShortKey)
+//nolint:staticcheck
+func pipelineRunPayloadName(pr *v1beta1.PipelineRun, opts config.StorageOpts) string {
+	return fmt.Sprintf(PayloadNameFormatPipelineRun, pr.Namespace, pr.Name, opts.ShortKey)
 }
 
+//nolint:staticcheck
 var (
 	_ api.Storer[*v1beta1.TaskRun, *in_toto.Statement] = &TaskRunStorer{}
```
lcarva, line 231: Let's do a similar check for PipelineRunStorer to ensure the interface is properly implemented.

pkg/chains/storage/gcs/gcs.go
```diff
@@ -185,41 +240,68 @@ type TaskRunStorer struct {
...
+//
+//nolint:staticcheck
+func (s *PipelineRunStorer) Store(ctx context.Context, req *api.StoreRequest[*v1beta1.PipelineRun, *in_toto.Statement]) (*api.StoreResponse, error) {
+	pr := req.Artifact
+	return store(ctx, s.writer, "pipelinerun", s.key,
+		pr.GetNamespace(), pr.GetName(), string(pr.GetUID()),
+		req.Bundle.Signature, req.Bundle.Content, req.Bundle.Cert, req.Bundle.Chain)
+}
+
+func store(ctx context.Context, writer gcsWriter, kind, key, namespace, name, uid string,
```
lcarva, line 272: This is quite a bit of params, and of the same type - easy to make a mistake.

Would it make sense to combine `kind, key, namespace, name, uid` into a single `prefix` parameter?

lcarva (commented):
Just some minor comments, but overall this looks good!

pkg/chains/storage/gcs/gcs.go
aaron-prindle: Ah thanks for catching this, updated now

pkg/chains/storage/gcs/gcs.go
aaron-prindle: Done, changed the params to use a `prefix` param directly instead of the additional 4 params mentioned - `kind, key, namespace, name, uid

aaron-prindle:
Thanks for the review @lcarva!  I believe I have addressed all of the feeback mentioned, lmk if there is anything additional to do here.

lcarva (approved):
/lgtm

PR #974, opened 2023-10-27
Fixes value for storage.oci.repository

PuneetPunamiya:
* Previously while providing repo url value for storage oci repository, chains controller was giving an error as
`a digest must contain exactly one '@' separator (e.g. registry/repository@digest)`
because, it was not able to add digest and repo name

* Hence this patch fixes it, by formatting the value provided by the user and thus storing the
attestations/signatures in the provide location

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/chains/storage/oci/legacy.go
+++ b/pkg/chains/storage/oci/legacy.go
@@ -54,7 +54,8 @@ type Backend struct {
 // NewStorageBackend returns a new OCI StorageBackend that stores signatures in an OCI registry
 func NewStorageBackend(ctx context.Context, client kubernetes.Interface, cfg config.Config) *Backend {
 	return &Backend{
-		cfg:    cfg,
+		cfg: cfg,
+
 		client: client,
 		getAuthenticator: func(ctx context.Context, obj objects.TektonObject, client kubernetes.Interface) (remote.Option, error) {
 			kc, err := k8schain.New(ctx, client,
@@ -119,12 +120,17 @@ func (b *Backend) uploadSignature(ctx context.Context, format simple.SimpleConta
 	imageName := format.ImageName()
 	logger.Infof("Uploading %s signature", imageName)
 
-	ref, err := newDigest(b.cfg, imageName)
+	ref, err := name.NewDigest(imageName)
 	if err != nil {
 		return errors.Wrap(err, "getting digest")
 	}
 
-	store, err := NewSimpleStorerFromConfig(WithTargetRepository(ref.Repository))
+	repo, err := newRepo(b.cfg, ref)
+	if err != nil {
+		return errors.Wrapf(err, "getting storage repo for sub %s", imageName)
+	}
+
+	store, err := NewSimpleStorerFromConfig(WithTargetRepository(repo))
 	if err != nil {
 		return err
 	}
@@ -154,12 +160,17 @@ func (b *Backend) uploadAttestation(ctx context.Context, attestation in_toto.Sta
 		imageName := fmt.Sprintf("%s@sha256:%s", subj.Name, subj.Digest["sha256"])
 		logger.Infof("Starting attestation upload to OCI for %s...", imageName)
 
-		ref, err := newDigest(b.cfg, imageName)
+		ref, err := name.NewDigest(imageName)
 		if err != nil {
 			return errors.Wrapf(err, "getting digest for subj %s", imageName)
 		}
 
-		store, err := NewAttestationStorer(WithTargetRepository(ref.Repository))
+		repo, err := newRepo(b.cfg, ref)
+		if err != nil {
+			return errors.Wrapf(err, "getting storage repo for sub %s", imageName)
+		}
+
+		store, err := NewAttestationStorer(WithTargetRepository(repo))
 		if err != nil {
 			return err
 		}
@@ -278,16 +289,14 @@ func (b *Backend) RetrieveArtifact(ctx context.Context, obj objects.TektonObject
 	return m, nil
 }
 
-func newDigest(cfg config.Config, imageName string) (name.Digest, error) {
-	// Override image name from config if set.
-	if r := cfg.Storage.OCI.Repository; r != "" {
-		imageName = r
-	}
-
+func newRepo(cfg config.Config, imageName name.Digest) (name.Repository, error) {
 	var opts []name.Option
 	if cfg.Storage.OCI.Insecure {
 		opts = append(opts, name.Insecure)
 	}
 
-	return name.NewDigest(imageName, opts...)
+	if storageOCIRepository := cfg.Storage.OCI.Repository; storageOCIRepository != "" {
+		return name.NewRepository(storageOCIRepository, opts...)
+	}
+	return name.NewRepository(imageName.Repository.Name(), opts...)
 }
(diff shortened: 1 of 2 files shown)


pkg/chains/storage/oci/legacy.go
```diff
@@ -280,8 +281,11 @@ func (b *Backend) RetrieveArtifact(ctx context.Context, obj objects.TektonObject
 
 func newDigest(cfg config.Config, imageName string) (name.Digest, error) {
 	// Override image name from config if set.
-	if r := cfg.Storage.OCI.Repository; r != "" {
-		imageName = r
+	if storageOCIRepository := cfg.Storage.OCI.Repository; storageOCIRepository != "" {
+		parts := strings.SplitN(imageName, "/", 3)
```
lcarva, line 285: I'm not sure if this covers the different possible use cases. `registry.com/spam/spam/spam/spam/spam/spam` is a perfectly valid repo name in most registries.

pkg/chains/storage/oci/legacy.go
```diff
@@ -280,8 +281,11 @@ func (b *Backend) RetrieveArtifact(ctx context.Context, obj objects.TektonObject
...
+		if len(parts) == 3 && (strings.ContainsRune(parts[0], '.') || strings.ContainsRune(parts[0], ':')) {
+			imageName = fmt.Sprintf("%s/%s", cfg.Storage.OCI.Repository, parts[2])
```
lcarva, line 287: This doesn't seem right. My understanding is that if `cfg.Storage.OCI.Repository` is set, then that's the full path to the repo where signatures and attestations will be pushed to. It is not meant to be a prefix to which a repo can be appended to.

lcarva (commented):
I'm not sure how this fixes the issue from the description. A unit test might be a good idea to express it.

pkg/chains/storage/oci/legacy.go
PuneetPunamiya: Yeah exactly I thought the same that it should be the full path of the repo where signatures and attestations should be pushed, but there are couple of things which are needed to push the signatures and attestations 

Consider you have given the value for `storage.oci.repository` as `docker.io/<username>/kaniko-chains` in the chains-config configMap

* First thing when it calls the [newDigest](https://github.com/tektoncd/chains/blob/main/pkg/chains/storage/oci/legacy.go#L281) function it returns an error which is from the [vendor](https://github.com/google/go-containerregistry/blob/main/pkg/name/digest.go#L67)  and hence the signing fails

* Now as you said 
> if cfg.Storage.OCI.Repository is set, then that's the full path to the repo where signatures and attestations will be pushed to

This is not true because consider you have set the value for `storage.oci.repository` as `docker.io/<username>/kaniko-chains` and artifacts are signed and pushed, but now you create a new taskrun which is different then it should not be pushed to the same artifact right 

Hence the value which should be set should be just the registry name and the repo name (considering if user doesn't provides registry name and just the repo name then as of today `docker.io` is used as by default)  

For example : `docker.io/<username>`

pkg/chains/storage/oci/legacy.go
PuneetPunamiya: Yes I agree with this, probably I should find a better way to make it more generalize

PuneetPunamiya: Yes I agree with this, probably I should find a better way to make it more generalize

PuneetPunamiya:
> I'm not sure how this fixes the issue from the description. A unit test might be a good idea to express it.

Yeah will be adding it 👍🏻

pkg/chains/storage/oci/legacy.go
lcarva: I'm not sure what you mean. In the example above, `registry.com` is the registry, and `spam/spam/spam/spam/spam/spam` is the repo name. Some registries may treat the first `spam` as a username, but that's not globally true.

PuneetPunamiya: Yes, you are right, there was a gap in my understanding, I was not aware that user is part of repo
I have updated the code accordingly

pkg/chains/storage/oci/legacy.go
```diff
@@ -279,15 +280,29 @@ func (b *Backend) RetrieveArtifact(ctx context.Context, obj objects.TektonObject
 }
 
 func newDigest(cfg config.Config, imageName string) (name.Digest, error) {
```
lcarva, line 282: I understand the problem now. So I think this needs a bit more work. The result of `newDigest` is used in two different places. Once to configure the storage, and once to configure the artifact being signed.

We want to take into account `storage.oci.repository` only for the storage piece of it. In other words, we don't want to create a signature/attestation with `storage.oci.repository` as its subject.

This appears to have broken in https://github.com/tektoncd/chains/pull/842.

I suggest the following:

1. Introduce a new helper function, e.g. newRepo, which returns the repo which is fed into `WithTargetRepository`. This should use `storage.oci.repository` if set. Otherwise, just use the image's repo.
2. Change newDigest to just parse the given image ref without making any modifications. Its return values should be used as the `Artifact` attribute in the `StoreRequest`.

PuneetPunamiya: Yeah, make sense, I have updated the pr accordingly

lcarva (changes requested):

pkg/chains/storage/oci/legacy.go
```diff
@@ -119,12 +119,22 @@ func (b *Backend) uploadSignature(ctx context.Context, format simple.SimpleConta
 	imageName := format.ImageName()
 	logger.Infof("Uploading %s signature", imageName)
 
-	ref, err := newDigest(b.cfg, imageName)
+	var opts []name.Option
+	if b.cfg.Storage.OCI.Insecure {
+		opts = append(opts, name.Insecure)
+	}
```
PuneetPunamiya, line 125: @lcarva  Not sure if this would be the right method to make sure the insecure option is applied even if cfg.Storage.OCI.Repository == "".

lcarva: In chat, I had suggested something like this:

```go
func newRepo(cfg config.Config, imageName name.Digest) (name.Repository, error) {
 	var opts []name.Option
 	if cfg.Storage.OCI.Insecure {
 		opts = append(opts, name.Insecure)
 	}
 	if storageOCIRepository := cfg.Storage.OCI.Repository; storageOCIRepository != "" {
		return name.NewRepository(storageOCIRepository, opts...)
 	}
	return name.NewRepository(imageName.RepositoryStr(), opts...)
 }
```

That should do the job.

pkg/chains/storage/oci/legacy.go
```diff
@@ -119,12 +119,22 @@ func (b *Backend) uploadSignature(ctx context.Context, format simple.SimpleConta
...
+
+	ref, err := name.NewDigest(imageName, opts...)
```
lcarva, line 127: I don't think we need the insecure option when calling `NewDigest`. That is just doing a simple string parsing and not actually hitting the registry.

lcarva (approved):
/lgtm

PR #998, opened 2023-11-23
Adds initial Attestor implementation.

wlynch:
# Changes

This is the initial implementation of Attestors, which uses generics to link chains components together with strict typing.

To start, this adds Attestor implementations of OCI signing and v1 SLSA attestations. These Attestors are NOT wired up to the controller yet, since they don't yet support the full range of config options (and there's likely a few tweaks we need to make in order to help reuse components like signers between Attestors).

`attestors.go` is the file to pay most attention to in this PR.

Part of #780 

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/chains/formats/simple/simple.go
+++ b/pkg/chains/formats/simple/simple.go
@@ -15,6 +15,7 @@ package simple
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/sigstore/sigstore/pkg/signature/payload"
@@ -66,6 +67,20 @@ func (i SimpleContainerImage) ImageName() string {
 	return fmt.Sprintf("%s@%s", i.Critical.Identity.DockerReference, i.Critical.Image.DockerManifestDigest)
 }
 
+func (i SimpleContainerImage) MarshalBinary() ([]byte, error) {
+	return json.Marshal(i)
+}
+
 func (i *SimpleSigning) Type() config.PayloadType {
 	return formats.PayloadTypeSimpleSigning
+}
+
+var (
+	_ formats.Formatter[name.Digest, SimpleContainerImage] = &SimpleSigningPayloader{}
+)
+
+type SimpleSigningPayloader SimpleSigning
+
+func (SimpleSigningPayloader) FormatPayload(_ context.Context, v name.Digest) (SimpleContainerImage, error) {
+	return NewSimpleStruct(v), nil
 }
--- a/pkg/chains/formats/slsa/v1/intotoite6.go
+++ b/pkg/chains/formats/slsa/v1/intotoite6.go
@@ -18,8 +18,10 @@ package v1
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
+	"github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/tektoncd/chains/pkg/chains/formats"
 	"github.com/tektoncd/chains/pkg/chains/formats/slsa/internal/slsaconfig"
 	"github.com/tektoncd/chains/pkg/chains/formats/slsa/v1/pipelinerun"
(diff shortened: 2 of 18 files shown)


pkg/chains/formats/simple/simple.go
```diff
@@ -66,6 +67,20 @@ func (i SimpleContainerImage) ImageName() string {
...
+
 func (i *SimpleSigning) Type() config.PayloadType {
 	return formats.PayloadTypeSimpleSigning
 }
+
+var (
+	_ formats.Formatter[name.Digest, SimpleContainerImage] = &SimpleSigningPayloader{}
+)
+
+type SimpleSigningPayloader SimpleSigning
```
lcarva, line 82: Nit, but this should probably be called something like `SimpleSigningFormatter`.

pkg/chains/formats/slsa/v1/intotoite6.go
```diff
@@ -34,21 +36,57 @@ const (
...
+	formats.RegisterPayloader(PayloadTypeInTotoIte6, NewPayloader)
+	formats.RegisterPayloader(PayloadTypeSlsav1, NewPayloader)
 }
 
 type InTotoIte6 struct {
 	slsaConfig *slsaconfig.SlsaConfig
 }
 
-func NewFormatter(cfg config.Config) (formats.Payloader, error) {
+func NewPayloader(cfg config.Config) (formats.Payloader, error) {
```
lcarva, line 47: Should this be marked as deprecated (also `NewPayloaderFromConfig`)?

lcarva: Actually... I don't think I fully understand Payloader vs Formatter :thinking:

pkg/chains/internal/attestors/attestors.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func NewContainerSigner(ctx context.Context, cfg config.Config) (*Attestor[name.Digest, simple.SimpleContainerImage], error) {
+	signer, err := chains.NewSignerFromConfig(ctx, "", cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	var opts []oci.Option
+	if repo := cfg.Storage.OCI.Repository; repo != "" {
```
lcarva, line 68: Let's pay special attention here. We recently had to fix a [bug](https://github.com/tektoncd/chains/pull/974) in our handling of this config value.

We should also make sure `cfg.Storage.OCI.Insecure` is handled properly.

pkg/chains/internal/attestors/attestors.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	storer, err := oci.NewAttestationStorer[*v1.ProvenanceStatement](opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Attestor[objects.TektonObject, *v1.ProvenanceStatement]{
+		payloader: v1.NewPayloaderFromConfig(cfg),
+		signer:    wrapped,
+		// TODO: add support for other storage options.
```
lcarva, line 114: Yeah, I'd be curious to see what this function would look like once we have support for all. It could end up quite large. Or maybe, `storer` is determined by another function that takes into account the current configuration, and returns the right thing.

pkg/chains/storage/oci/attestation.go
```diff
@@ -73,16 +69,47 @@ func (s *AttestationStorer) Store(ctx context.Context, req *api.StoreRequest[nam
...
+		return nil, merr
 	}
-	logger.Infof("Successfully uploaded attestation for %s", req.Artifact.String())
 
 	return &api.StoreResponse{}, nil
 }
+
+func (s *AttestationStorer[T]) storeImage(ctx context.Context, img name.Digest, att oci.Signature) error {
+	logger := logging.FromContext(ctx)
+	repo := img.Repository
```
lcarva, line 95: We need to handle `cfg.Storage.OCI.Insecure` here somehow.

lcarva (commented):
Overall, it makes sense to me. Need to have a more thorough look.

lcarva:
@wlynch, is this something you're still actively pursuing?

anithapriyanatarajan:
Closing this PR, since there is no activity. The underlying issue is open for pick up

PR #999, opened 2023-11-24
Use cluster issuer as default builder id

### LINKED ISSUE #640: The default builderID needs to be changed.
# Background
One important field in SLSA provenance is [predicate.builder.builderID](https://slsa.dev/provenance/v0.2#builder). 

Chains does support setting this value via the configmap field named [builder.id](https://github.com/tektoncd/chains/blob/main/docs/config.md#in-toto-configuration), which is great and we'd like to keep this. But the default builderID `https://tekton.dev/chains/v2` set [here](https://github.com/tektoncd/chains/blob/6d80daed68aa9b08670b6213df02f763d077d31d/pkg/config/config.go#L240-L242) sounds not right.

# Expected Behavior
 
Chains has correct default value for the builder id.

# Actual Behavior

The default value used by Chains is `https://tekton.dev/chains/v2`. It doesn't make sense because Chains apparently is not the builder and doesn't contribute to the artifact build.

wlynch:
# Changes

This is probably the most "correct" value for this field - it should uniquely identify the cluster, and will match other signature data included in Fulcio certs, etc.

This is technically a breaking change, but likely one worth making. Users can still override this behavior with the config map as before. If omitted, this field is not populated as an indication that we don't know how to accurately identify this cluster.

Part of #640 - this doesn't resolve the SLSA Build Level aspects of the Builder ID (which I wish it didn't have to be encoded in the Builder ID, but that ship has sailed for SLSA v1 ¯\_(ツ)_/¯)

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
BREAKING CHANGE: The default Builder ID for SLSA provenance now defaults to the cluster's OIDC token issuer instead of a static string.
```


anithapriyanatarajan:
Closing this PR, since there is no activity.

Issue #858, opened 2023-07-07
Chains adds "signed: true" annotation even if no keys supplied


PuneetPunamiya:
* When chains is installed, it adds the `signed: true` annotation to the taskruns even though the controller is filled with logs around not being able to configure a signer and not having any configured keys.

For example :

I create this simple taskrun which is mentioned in the getting-started tutorial 
```yaml
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  name: build-push-run-output-image-test
spec:
  serviceAccountName: ""
  taskSpec:
    results:
    - name: IMAGE_URL
      type: string
    - name: IMAGE_DIGEST
      type: string
    steps:
    - name: create-image
      image: busybox
      script: |-
        #!/usr/bin/env sh
        echo 'gcr.io/foo/bar' | tee $(results.IMAGE_URL.path)
        echo 'sha256:05f95b26ed10668b7183c1e2da98610e91372fa9f510046d4ce5812addad86b5' | tee $(results.IMAGE_DIGEST.path)
```

```bash
kubectl create -f https://raw.githubusercontent.com/tektoncd/chains/main/examples/taskruns/task-output-image.yaml
```

Now even if no field is provided in the the `chain-config` configMap it adds annotation `chains.tekton.dev/signed=true` in the taskrun 

```bash
╰─ tkn tr describe --last 
Name:              build-push-run-output-image-test
Namespace:         default
Service Account:   default
Timeout:           1h0m0s
Labels:
 app.kubernetes.io/managed-by=tekton-pipelines
Annotations:
 chains.tekton.dev/signed=true
 pipeline.tekton.dev/release=c802069

🌡️  Status

STARTED         DURATION    STATUS
2 minutes ago   28s         Succeeded

📝 Results

 NAME             VALUE
 ∙ IMAGE_DIGEST   sha256:05f95b26ed10668b7183c1e2da98610e91372fa9f510046d4ce5812addad86b5
 ∙ IMAGE_URL      gcr.io/foo/bar

🦶 Steps

 NAME             STATUS
 ∙ create-image   Completed

```

* Chains Controller Logs
 ```bash
INFO  16:30:15 Unable to read vcs.revision from binary
INFO  16:30:15 Profiling enabled: false
INFO  16:30:15 Running with Standard leader election
INFO  16:30:15 Starting configuration manager...
INFO  16:30:16 Starting informers...
INFO  16:30:16 Starting controllers...
[...]

PR #1003, opened 2023-11-30
Skips the signing of artifacts when valid configuration is not done

PuneetPunamiya:
Before this patch, if proper configurations where not done, chains used to sign the tekton resources with the default configurations which in turn used to create a impression that tekton resources are signed as it used to add the `chains.tekton.dev/signed=true` annotation

Hence, with this patch, if proper configuration is not done (like for e.g. private keys are not stored using cosign or x509) then in that case tekton resources will be skipped from signing, indicating the user to do proper configurations to sign the resources

Fixes: https://github.com/tektoncd/chains/issues/858

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -46,7 +46,7 @@ type ObjectSigner struct {
 	Pipelineclientset versioned.Interface
 }
 
-func allSigners(ctx context.Context, sp string, cfg config.Config) map[string]signing.Signer {
+func allSigners(ctx context.Context, sp string, cfg config.Config) (map[string]signing.Signer, error) {
 	l := logging.FromContext(ctx)
 	all := map[string]signing.Signer{}
 	neededSigners := map[string]struct{}{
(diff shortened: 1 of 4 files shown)


pkg/chains/signing.go
```diff
@@ -64,22 +64,23 @@ func allSigners(ctx context.Context, sp string, cfg config.Config) map[string]si
...
+				return nil, err
 			}
 			all[s] = signer
 		default:
 			// This should never happen, so panic
 			l.Panicf("unsupported signer: %s", s)
 		}
 	}
-	return all
+	fmt.Println(all)
```
lcarva, line 82: Let's remove this :)

pkg/chains/signing_test.go
```diff
@@ -184,6 +185,56 @@ func TestSigner_Sign(t *testing.T) {
 	}
 }
 
+func TestSigningObjectsSkipped(t *testing.T) {
+	ctx, _ := rtesting.SetupFakeContext(t)
+	ps := fakepipelineclient.Get(ctx)
+
+	tro := objects.NewTaskRunObject(&v1beta1.TaskRun{
```
chitrangpatel, line 192: agreed with lint. Why are we using `v1beta1` here? I think it should be `v1`?

lcarva:
This [comment](https://github.com/tektoncd/chains/issues/858#issuecomment-1845796823) indicates we do not want to pursue the `skipped` annotation approach. Let's take the discussion back to that issue.

lcarva:
Based on my previous comment, let's close out this PR.

PR #1025, opened 2024-01-03
Change Leader Election Config Map for Chains Controller

### LINKED ISSUE #1024: Change Leader Election Configmap for Controllers
# Expected Behavior
We should have a separate configmap for leader election.

# Actual Behavior
Currently, it assumes default configmap config-leader-election which is used by all knative controllers.

This is more of a problem when we try to install chains in the same namespace as the Pipelines controller.  And, we want to fine-tune every controller.

1.
2.
3.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  (paste your output here)
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

khrm:
Currently, the chains controller assumes `config-leader-election` configmap as the name for the leader election configmap. This is the default name for the Knative controllers' leader election configmap. If some other controllers also exist, it's a good practice to have a separate configmap for every controller.
Fixes #1024 

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
BREAKING CHANGE:
The name of the configmap for leader election of chains controller has been changed to
`tekton-chains-config-leader-election` from default `config-leader-election`.
Any existing users of this configmap are requested to override the name to back to `config-leader-election`,
or migrate their existing config to `tekton-chains-config-leader-election`.
```

### DIFF
--- a/config/100-deployment.yaml
+++ b/config/100-deployment.yaml
@@ -103,6 +103,9 @@ spec:
               value: tekton.dev/chains
             - name: CONFIG_OBSERVABILITY_NAME
               value: tekton-chains-config-observability
+            - name: CONFIG_LEADERELECTION_NAME
+              value: tekton-chains-config-leader-election
+
           ports:
             - name: metrics
               containerPort: 9090
--- a/config/config-leader-election.yaml
+++ b/config/config-leader-election.yaml
@@ -1,0 +1,52 @@
+# Copyright 2023 Tekton Authors LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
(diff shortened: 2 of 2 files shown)


khrm (commented):
/kind bug

wlynch (dismissed):
Lgtm, but this should be called out as a potentially breaking change in the release notes, since this will affect anyone relying on the previous behavior.

vdemeester:
@wlynch given we did not ship the configmap at all in the chains release, do we think a lot of people relied on it ?
Anyway, I agree, we need to highlight that in the release changelog yes.

wlynch:
> @wlynch given we did not ship the configmap at all in the chains release, do we think a lot of people relied on it ? Anyway, I agree, we need to highlight that in the release changelog yes.

Technically the feature has always existed because of our usage of the knative controller framework, but we never really advertised it. This is changing the default configmap that chains looks for, which would technically break anyone currently setting the existing default knative leader election configmap (which I assume is how @khrm discovered this).

I agree its unlikely that many people were aware that this feature existed, which is why I'm fine with making the change even though it's potentially breaking. But we should update the PR description release notes with the change + instructions for how to revert back to the old behavior if needed.

khrm:
@wlynch I have added the Breaking change note. And also updated the PR description.

khrm (commented):
/remove-kind bug

lcarva:
@wlynch, I think your comment was addressed, but before I add the lgtm label, I want to make sure you have a chance to take another look.

wlynch:
Made some tweaks to the release note, but otherwise lgtm!

/lgtm

PR #1048, opened 2024-02-08
Adds custom namespace in e2e tests

PuneetPunamiya:
- Previously all namespace was hard coded in e2e tests and also when chains is installed in operator it is not installed in `tekton-chains` ns by default, it is installed in `tekton-pipelines`

- Hence this patch by default uses `tekton-chains` ns in e2e, but if chains is installed in other ns, then we can still run the tests in that particular ns

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/test/clients.go
+++ b/test/clients.go
@@ -35,6 +35,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"testing"
 
@@ -202,7 +203,7 @@ type secret struct {
 
 func setupSecret(ctx context.Context, t *testing.T, c kubernetes.Interface, opts setupOpts) secret {
 	// Only overwrite the secret data if it isn't set.
-	namespace := "tekton-chains"
+	namespace := os.Getenv("namespace")
 	s := corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "signing-secrets",
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -17,6 +17,9 @@
 # This script calls out to scripts in tektoncd/plumbing to setup a cluster
 # and deploy Tekton Pipelines to it for running integration tests.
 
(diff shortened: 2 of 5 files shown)


test/e2e_test.go
```diff
@@ -53,11 +53,17 @@ import (
 	logtesting "knative.dev/pkg/logging/testing"
 )
 
+var namespace string
+
+func init() {
+	namespace = os.Getenv("namespace")
```
lcarva, line 59: Does this need a default value?

test/e2e-tests.sh
```diff
@@ -41,6 +41,14 @@ chains_patch_spire
 
 failed=0
 
+namespace="tekton-chains"
```
lcarva, line 44: You can do something like this to provide a default value:

```bash
namespace="${NAMESPACE:-tekton-chains}"
```

test/clients.go
```diff
@@ -202,7 +203,7 @@ type secret struct {
 
 func setupSecret(ctx context.Context, t *testing.T, c kubernetes.Interface, opts setupOpts) secret {
 	// Only overwrite the secret data if it isn't set.
-	namespace := "tekton-chains"
+	namespace := os.Getenv("namespace")
```
lcarva, line 206: Set default value?

test/e2e-tests.sh
```diff
@@ -41,6 +41,14 @@ chains_patch_spire
...
+
+# Check if the NAMESPACE environment variable is set
+if [[ ! -z "$NAMESPACE" ]]; then
+  namespace="$NAMESPACE"
+fi
+echo "Using namespace: $namespace"
```
lcarva, line 50: I think this is ok, but it has the potential of being out of sync with the logic in the tests.

test/e2e_test.go
PuneetPunamiya: Good catch, adding default value as `tekton-chains` if env is empty

lcarva (approved):
/lgtm

Issue #1128, opened 2024-06-03
in-toto Statement not compatible with previous versions


renzodavid9:
# Expected Behavior
When getting the statement, the resulting json should have same field names as before so it doesn't break any integration.

# Actual Behavior
The generated statement is changing the following field names:
* From `_type` to `type`
* From `predicateType` to `predicate_type`

E.g:
```json
{
    "type": "https://in-toto.io/Statement/v0.1",
    "predicate_type": "https://slsa.dev/provenance/v0.2",
    "subject": [
        {
            "name": "gcr.io/foo/bar",
            "digest": {
                "sha256": "05f95b26ed10668b7183c1e2da98610e91372fa9f510046d4ce5812addad86b5"
            }
        }
    ],
...
}
```

# Steps to Reproduce the Problem

Running any TaskRun/PipelineRun, after the statement is generated it looks like the snippet above

# Additional Info
* Chains version: v0.21.0
* More info in: https://github.com/in-toto/attestation/issues/363

PR #1129, opened 2024-06-03
Fix field names in generated json statement

renzodavid9:
Fixes: #1128

To keep backwards compatibility with previous struct versions (e.g, github.com/in-toto/in-toto-golang/in_toto) we have to use protojson.Marshal function to transform the struct to its json representation so the tags defined in the intoto.Statement are follow and applied to the resulting string.

# Changes

When getting the payload object we check if it is an instance of proto.Message to use the proper Marshall function.

# Submitter Checklist

  functionality, content, code)

### DIFF
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/go-multierror"
+	intoto "github.com/in-toto/attestation/go/v1"
 	"github.com/tektoncd/chains/pkg/artifacts"
 	"github.com/tektoncd/chains/pkg/chains/formats"
 	"github.com/tektoncd/chains/pkg/chains/objects"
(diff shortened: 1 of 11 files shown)


pkg/chains/signing.go
```diff
@@ -248,3 +250,23 @@ func HandleRetry(ctx context.Context, obj objects.TektonObject, ps versioned.Int
 	}
 	return MarkFailed(ctx, obj, ps, annotations)
 }
+
+// getRawPayload returns the payload as a json string. If the given payload is a proto.Message type, protojson.Marshal
+// is used to get the proper labels/field names in the resulting json.
+func getRawPayload(payload interface{}) ([]byte, error) {
```
chitrangpatel, line 256: Can we test this function by any chance?

renzodavid9: I saw some other places where private functions are unit-tested, so I follow the same pattern and added a unit-test for this function, let me know if that's what you had in mind. Thanks!

chitrangpatel: LGTM!

chitrangpatel (approved):

PR #1148, opened 2024-07-01
Support multinamespace informer filtering

jkhelil:
# Changes
Support of specifying a list of namespaces to restrict the informer. You can provide a list of namespaces as an argument to the controller, and it will only watch objects (PipelineRuns and TaskRuns) within these specified namespaces. If no list is provided, the controller will default to watching all namespaces.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
The Chains controller now supports specifying a list of namespaces to restrict the informer. 
You can provide a list of namespaces as an argument to the controller, and it will only watch objects (PipelineRuns and TaskRuns) within these specified namespaces. 
If no list is provided, the controller will default to watching all namespaces.

```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"flag"
+	"strings"
 
 	"github.com/tektoncd/chains/pkg/reconciler/pipelinerun"
 	"github.com/tektoncd/chains/pkg/reconciler/taskrun"
@@ -24,6 +25,7 @@ import (
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/injection"
 	"knative.dev/pkg/injection/sharedmain"
+	"knative.dev/pkg/logging"
 	"knative.dev/pkg/signals"
 
 	// Run with all of the upstream providers.
@@ -41,10 +43,20 @@ import (
 
 func main() {
 	flag.IntVar(&controller.DefaultThreadsPerController, "threads-per-controller", controller.DefaultThreadsPerController, "Threads (goroutines) to create per controller")
-	namespace := flag.String("namespace", "", "Namespace to restrict informer to. Optional, defaults to all namespaces.")
+	namespaceList := flag.String("namespace", "", "Comma-separated list of namespaces to restrict informer to. Optional, if empty defaults to all namespaces.")
 
 	// This also calls flag.Parse().
 	cfg := injection.ParseAndGetRESTConfigOrDie()
+
+	ctx := signals.NewContext()
+	logger := logging.FromContext(ctx)
+
+	var namespaces []string
+	if *namespaceList != "" {
+		// Remove any whitespace from the namespaces string and split it
+		namespaces = strings.Split(strings.ReplaceAll(*namespaceList, " ", ""), ",")
+		logger.Infof("controller is scoped to the following namespaces: %s\n", namespaces)
+	}
 
 	if cfg.QPS == 0 {
 		cfg.QPS = 2 * rest.DefaultQPS
@@ -57,8 +69,5 @@ func main() {
 	cfg.QPS = 2 * cfg.QPS
 	cfg.Burst = 2 * cfg.Burst
 
-	flag.Parse()
-	ctx := injection.WithNamespaceScope(signals.NewContext(), *namespace)
-
-	sharedmain.MainWithConfig(ctx, "watcher", cfg, taskrun.NewController, pipelinerun.NewController)
+	sharedmain.MainWithConfig(ctx, "watcher", cfg, taskrun.NewNamespacesScopedController(namespaces), pipelinerun.NewNamespacesScopedController(namespaces))
 }
--- a/docs/config.md
+++ b/docs/config.md
@@ -4,9 +4,7 @@ linkTitle: "Chains Configuration"
 weight: 20
 ---
 -->
-
 # Chains Configuration
-
 `Chains` works by observing `TaskRun` and `PipelineRun` executions, capturing relevant information, and storing it in a cryptographically-signed format.
 
 `TaskRuns` and `PipelineRuns` can indicate inputs and outputs which are then captured and surfaced in the `Chains` payload formats, where relevant.
@@ -159,3 +157,16 @@ chains.tekton.dev/transparency-upload: "true"
 
 > [!IMPORTANT]
 > To project the latest token values without needing to recreate the pod, avoid using `subPath` in volume mount.
+
+## Namespaces Restrictions in Chains Controller
+This feature allows you to specify a list of namespaces for the controller to monitor, providing granular control over its operation. If no namespaces are specified, the controller defaults to monitoring all namespaces.
+
+### Usage
+To restrict the Chains Controller to specific namespaces, pass a comma-separated list of namespaces as an argument to the controller using the --namespace flag.
+
+### Example
+To restrict the controller to the dev and test namespaces, you would start the controller with the following argument:
+```shell
+--namespace=dev,test
+```
+In this example, the controller will only monitor resources (pipelinesruns and taskruns) within the dev and test namespaces.
--- a/pkg/reconciler/filter.go
+++ b/pkg/reconciler/filter.go
@@ -1,0 +1,80 @@
+/*
+Copyright 2024 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconciler
+
+import (
+	v1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	"knative.dev/pkg/controller"
+)
+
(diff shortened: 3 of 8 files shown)


cmd/controller/main.go
```diff
@@ -41,7 +44,17 @@ import (
...
+	namespaceList := flag.String("namespaces", "", "Comma-separated list of namespaces to restrict informer to. Optional, if empty defaults to all namespaces.")
+	flag.Parse()
+
+	var namespaces []string
+	if namespaceList != nil && *namespaceList != "" {
+		space := regexp.MustCompile(`\s+`)
+		namespaces = strings.Split(space.ReplaceAllString(*namespaceList, ""), ",")
+		fmt.Printf("controller is scopped to the following namespaces: %s\n", namespaces)
+	} else {
+		namespaces = nil // Default to all namespaces if the list is empty
```
PuneetPunamiya, line 56: I think this will fail and won't set to all namespaces as `""` won't be set i.e. it will not default to all namespaces and it will fail here https://github.com/tektoncd/chains/pull/1148/files#diff-4ce7e773e47a412c0fcdc0db37ed953113e7744b300b2db1f485a20ca6b7ed0bR32 as it will have an empty array

pkg/reconciler/filter.go
```diff
@@ -0,0 +1,67 @@
+package reconciler
```
PuneetPunamiya, line 1: Can you please add license here

PuneetPunamiya, line 1: Can you please add unit test for this

cmd/controller/main.go
jkhelil: @PuneetPunamiya Why would it fail ? len(nil slice) = 0, and namespaces will not be an empty error because we force to nil fi the namespaces flag is not set or empty

pkg/reconciler/filter.go
jkhelil: done

jkhelil: done

pkg/reconciler/taskrun/controller.go
```diff
@@ -31,50 +33,57 @@ import (
...
+			Pipelineclientset: pipelineClient,
+			Recorder:          taskrunmetrics.Get(ctx),
+		}
 
-	c := &Reconciler{
-		TaskRunSigner:     tsSigner,
-		Pipelineclientset: pipelineClient,
-	}
-	impl := taskrunreconciler.NewImpl(ctx, c, func(impl *controller.Impl) controller.Options {
-		cfgStore := config.NewConfigStore(logger, func(name string, value interface{}) {
```
PuneetPunamiya, line 52: I don't understand how lint was not failing before here 🤔

cmd/controller/main.go
```diff
@@ -57,8 +71,7 @@ func main() {
 	cfg.QPS = 2 * cfg.QPS
 	cfg.Burst = 2 * cfg.Burst
 
-	flag.Parse()
-	ctx := injection.WithNamespaceScope(signals.NewContext(), *namespace)
+	ctx := injection.WithNamespaceScope(signals.NewContext(), "")
```
wlynch, line 74: We can omit this (if the context value isn't set, this returns "" anyway).

https://github.com/knative/pkg/blob/7ecd5485cbc68c6cac94f626d3c7b11ff3b1275a/injection/context.go#L45-L51

pkg/reconciler/pipelinerun/controller.go
```diff
@@ -34,61 +34,68 @@ import (
 	_ "github.com/tektoncd/chains/pkg/chains/formats/all"
 )
 
-func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
-	logger := logging.FromContext(ctx)
-	pipelineRunInformer := pipelineruninformer.Get(ctx)
-	taskRunInformer := taskruninformer.Get(ctx)
+// NewNamespacesScoppedController returns a new controller implementation where informer is filtered
+// given a list of namespaces
+func NewNamespacesScoppedController(namespaces []string) func(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
```
wlynch, line 39:
```suggestion
func NewNamespacesScopedController(namespaces []string) func(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
```

pkg/reconciler/taskrun/controller.go
```diff
@@ -31,50 +33,57 @@ import (
 	_ "github.com/tektoncd/chains/pkg/chains/formats/all"
 )
 
-func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
-	logger := logging.FromContext(ctx)
-	taskRunInformer := taskruninformer.Get(ctx)
+// NewNamespacesScoppedController returns a new controller implementation where informer is filtered
+// given a list of namespaces
+func NewNamespacesScoppedController(namespaces []string) func(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
```
wlynch, line 38:
```suggestion
func NewNamespacesScopedController(namespaces []string) func(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
```

pkg/reconciler/taskrun/controller.go
wlynch: Lint is configured to only look at new changes, so anything that was failing lint before it was enabled doesn't get flagged until it's modified.

wlynch (changes requested):

cmd/controller/main.go
```diff
@@ -41,7 +44,20 @@ import (
...
 	flag.IntVar(&controller.DefaultThreadsPerController, "threads-per-controller", controller.DefaultThreadsPerController, "Threads (goroutines) to create per controller")
-	namespace := flag.String("namespace", "", "Namespace to restrict informer to. Optional, defaults to all namespaces.")
+	namespaceList := flag.String("namespace", "", "Comma-separated list of namespaces to restrict informer to. Optional, if empty defaults to all namespaces.")
+	flag.Parse()
+
+	var namespaces []string
+	if *namespaceList != "" {
+		// Remove any whitespace from the namespaces string
+		space := regexp.MustCompile(`\s+`)
+		namespaces = strings.Split(space.ReplaceAllString(*namespaceList, ""), ",")
```
lcarva, line 54: nipick: `strings.ReplaceAll` is probably a better approach here.

cmd/controller/main.go
```diff
@@ -41,7 +44,20 @@ import (
...
+		fmt.Printf("controller is scopped to the following namespaces: %s\n", namespaces)
```
lcarva, line 55: Let's use a logger here?

Also, `s/scopped/scoped`

cmd/controller/main.go
lcarva: This seems unnecessary though. If it got here, it's already `nil`.

cmd/controller/main.go
```diff
@@ -41,7 +44,20 @@ import (
 
 func main() {
 	flag.IntVar(&controller.DefaultThreadsPerController, "threads-per-controller", controller.DefaultThreadsPerController, "Threads (goroutines) to create per controller")
-	namespace := flag.String("namespace", "", "Namespace to restrict informer to. Optional, defaults to all namespaces.")
+	namespaceList := flag.String("namespace", "", "Comma-separated list of namespaces to restrict informer to. Optional, if empty defaults to all namespaces.")
+	flag.Parse()
```
lcarva, line 48: Is moving this call here going to cause issues when calling `cfg := injection.ParseAndGetRESTConfigOrDie()` later on? I don't _think_ it will, but we should verify.

docs/config.md
```diff
@@ -140,3 +138,12 @@ chains.tekton.dev/transparency-upload: "true"
 | `signers.kms.auth.oidc.role`      | Role used for OIDC authentication                                                           |                  |
 | `signers.kms.auth.spire.sock`     | URI of the Spire socket used for KMS token (e.g. `unix:///tmp/spire-agent/public/api.sock`) |                  |
 | `signers.kms.auth.spire.audience` | Audience for requesting a SVID from Spire                                                   |                  |
+
+## Supporting Namespace Restrictions in Chains Controller
```
lcarva, line 142: Consider re-writing this section to only describe the feature once, then explain how to use it, and provide an example:

> ## Namespace Restrictions
> 
> By default, Chains will process TaskRun and PipelineRun resources in all namespaces. This may not be ideal for certain use cases. To restrict the namespaces processed by Chains use the `--namespace` parameter in the Chains controller. The value for this parameter is a comma separated list of namespaces, e.g. `--namespace default,my-app`.

docs/config.md
```diff
@@ -158,4 +156,18 @@ chains.tekton.dev/transparency-upload: "true"
...
 
 > [!IMPORTANT]
-> To project the latest token values without needing to recreate the pod, avoid using `subPath` in volume mount.
\ No newline at end of file
+> To project the latest token values without needing to recreate the pod, avoid using `subPath` in volume mount.
+
+## Namespaces Restrictions in Chains Controller
+This feature allows you to specify a list of namespaces for the controller to monitor, providing granular control over its operation. If no namespaces are specified, the controller defaults to monitoring all namespaces.
+
+Usage
```
lcarva, line 164: This line and the next are rendered as one line. Did you intend something like this instead:

```md
### Usage

To restrict...
```

That will render as:

> ### Usage
> 
> To restrict...

Or maybe just add an extra line break:

```md
Usage

To restrict...
```

which should render as:

> Usage
>
> To restrict...

Same for the other header-like line in this section.

lcarva:
@wlynch, I believe your comments have been addressed. Could you have another look?

wlynch (approved):

PR #1158, opened 2024-07-16
Read KMS Auth token from filepath

anithapriyanatarajan:
# Changes

Support for retrieving KMS provider Auth token from mount path configured with the new key `signers.kms.auth.token-dir` in configMap chains-config.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Chains now supports extracting the KMS provider auth token (e.g., `VAULT_TOKEN`) from a specified filepath. 
Instead of using the configMap key `signers.kms.auth.token`, you can now use `signers.kms.auth.token-dir` pointing to a valid filepath within the container (e.g., `/etc/kms-secrets`).

### Steps to Utilize This Feature:
- Create a secret with the key name `KMS_AUTH_TOKEN`.
- Update the Chains deployment specification to volume mount this secret. 

**Note:** To project the latest token values without needing to recreate the pod, avoid using `subPath`.
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -147,7 +147,15 @@ chains.tekton.dev/transparency-upload: "true"
 | :-------------------------------- | :------------------------------------------------------------------------------------------ | :--------------- | :------ |
 | `signers.kms.auth.address`        | URI of KMS server (e.g. the value of `VAULT_ADDR`)                                          |                  |
 | `signers.kms.auth.token`          | Auth token KMS server (e.g. the value of `VAULT_TOKEN`)                                     |                  |
+| `signers.kms.auth.token-dir`      | Path to store KMS server Auth token (e.g. `/etc/kms-secrets`)                               |                  |
 | `signers.kms.auth.oidc.path`      | Path used for OIDC authentication (e.g. `jwt` for Vault)                                    |                  |
 | `signers.kms.auth.oidc.role`      | Role used for OIDC authentication                                                           |                  |
 | `signers.kms.auth.spire.sock`     | URI of the Spire socket used for KMS token (e.g. `unix:///tmp/spire-agent/public/api.sock`) |                  |
 | `signers.kms.auth.spire.audience` | Audience for requesting a SVID from Spire                                                   |                  |
+> NOTE:
+>
+> If `signers.kms.auth.token-dir` is set, create a secret with the key `KMS_AUTH_TOKEN` and ensure the Chains deployment mounts this secret to 
+> the path specified by `signers.kms.auth.token-dir`. 
+
+> [!IMPORTANT]
+> To project the latest token values without needing to recreate the pod, avoid using `subPath` in volume mount.
--- a/pkg/chains/signing/kms/kms.go
+++ b/pkg/chains/signing/kms/kms.go
@@ -20,6 +20,9 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/sigstore/sigstore/pkg/signature"
@@ -95,12 +98,26 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
 	// pass through configuration options to RPCAuth used by KMS in sigstore
 	rpcAuth := options.RPCAuth{
 		Address: cfg.Auth.Address,
-		Token:   cfg.Auth.Token,
 		OIDC: options.RPCAuthOIDC{
 			Role: cfg.Auth.OIDC.Role,
 			Path: cfg.Auth.OIDC.Path,
 		},
 	}
+
+	// get token from file KMS_AUTH_TOKEN, a mounted secret at signers.kms.auth.token-dir or
+	// as direct value set from signers.kms.auth.token.
+	// If both values are set, priority will be given to token-dir.
+
+	if cfg.Auth.TokenDir != "" {
+		rpcAuthToken, err := getKMSAuthToken(cfg.Auth.TokenDir)
+		if err != nil {
+			return nil, err
+		}
+		rpcAuth.Token = rpcAuthToken
+	} else {
+		rpcAuth.Token = cfg.Auth.Token
+	}
+
 	// get token from spire
 	if cfg.Auth.Spire.Sock != "" {
 		token, err := newSpireToken(ctx, cfg)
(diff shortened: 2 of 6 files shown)


pkg/chains/signing/kms/kms.go
```diff
@@ -95,12 +98,27 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
 	// pass through configuration options to RPCAuth used by KMS in sigstore
 	rpcAuth := options.RPCAuth{
 		Address: cfg.Auth.Address,
-		Token:   cfg.Auth.Token,
+		//Token:   cfg.Auth.Token,
```
PuneetPunamiya, line 101: Can we remove this instead of adding a comment ?

pkg/chains/signing/kms/kms.go
```diff
@@ -95,12 +98,27 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
...
 			Role: cfg.Auth.OIDC.Role,
 			Path: cfg.Auth.OIDC.Path,
 		},
 	}
+
+	// get token from file KMS_AUTH_TOKEN, a mounted secret at signers.kms.auth.token-dir or
+	// as direct value set from signers.kms.auth.token.
+	// If both values are set, priority will be given to token-dir.
+
+	if cfg.Auth.TokenDir != "" {
```
PuneetPunamiya, line 112: The only thing I'm thinking this is should we support for all kms providers or just vault ?

pkg/chains/signing/kms/kms.go
```diff
@@ -95,12 +98,27 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
...
+		rpcAuthToken, err := getRPCAuthToken(cfg.Auth.TokenDir)
```
PuneetPunamiya, line 113: If we are adding for all kms providers 

```suggestion
		rpcAuthToken, err := getKMSAuthToken(cfg.Auth.TokenDir)
```

docs/config.md
```diff
@@ -147,6 +147,7 @@ chains.tekton.dev/transparency-upload: "true"
 | :-------------------------------- | :------------------------------------------------------------------------------------------ | :--------------- | :------ |
 | `signers.kms.auth.address`        | URI of KMS server (e.g. the value of `VAULT_ADDR`)                                          |                  |
 | `signers.kms.auth.token`          | Auth token KMS server (e.g. the value of `VAULT_TOKEN`)                                     |                  |
+| `signers.kms.auth.token-dir`      | Path to store KMS server Auth token (e.g. `/etc/kms-secrets`)                               |                  |
```
PuneetPunamiya, line 150: Also we should add a note in the docs that a secret should be created with the key and should be mounted at a certain path to chains controller pod

pkg/chains/signing/kms/kms.go
anithapriyanatarajan: removed.

pkg/chains/signing/kms/kms.go
anithapriyanatarajan: There is a possibility for other providers as well to allow token based auth to the keys. The dependent package sigstore signature options as well has it generic.(https://github.com/sigstore/sigstore/blob/main/pkg/signature/options/rpcauth.go#L25) . so I believe it is fine to have it generic.

pkg/chains/signing/kms/kms.go
anithapriyanatarajan: changed

docs/config.md
anithapriyanatarajan: Added required notes.

pkg/chains/signing/kms/kms.go
```diff
@@ -120,6 +137,28 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
 	}, nil
 }
 
+// getKMSAuthToken retreives token from the given mount path
+func getKMSAuthToken(dir string) (string, error) {
+	tokenEnv := "KMS_AUTH_TOKEN"
+	stat, err := os.Stat(dir)
```
lcarva, line 143: I wonder if it's necessary to check that the path exists and it is a directory. We could simply do

```go
filePath := filepath.Join(dir, tokenEnv)
fileData, err := os.ReadFile(filePath)
```

which should catch those cases and more.

pkg/chains/signing/kms/kms.go
```diff
@@ -120,6 +137,28 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
...
+	}
+	// If the path exists but is not a directory, then throw an error
+	if !stat.IsDir() {
+		return "", fmt.Errorf("path specified %s is not a directory", dir)
+	}
+
+	filePath := filepath.Join(dir, tokenEnv)
+	fileData, err := os.ReadFile(filePath)
+	if err != nil {
+		return "", err
```
lcarva, line 155: Maybe wrap this error to make it clear which file Chains is trying to load:

```go
return "", fmt.Errorf("reading file %q in directory %q: %w", tokenEnv, dir, err)
```

pkg/chains/signing/kms/kms.go
```diff
@@ -95,12 +98,26 @@ func NewSigner(ctx context.Context, cfg config.KMSSigner) (*Signer, error) {
...
-		Token:   cfg.Auth.Token,
 		OIDC: options.RPCAuthOIDC{
 			Role: cfg.Auth.OIDC.Role,
 			Path: cfg.Auth.OIDC.Path,
 		},
 	}
+
+	// get token from file KMS_AUTH_TOKEN, a mounted secret at signers.kms.auth.token-dir or
+	// as direct value set from signers.kms.auth.token.
+	// If both values are set, priority will be given to token-dir.
```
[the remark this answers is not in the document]
anithapriyanatarajan: Added notes to docs/signing.md https://github.com/tektoncd/chains/pull/1158/files#diff-fce50e287cd36c8a0cf91255755170b8cd8441d0460666abee8378ec918ed3acR77

pkg/chains/signing/kms/kms.go
anithapriyanatarajan: Thanks for this comment. removed check for directory with os.Stat and tested. I may be helpful to user to discover that the mountpath is incorrect  but with a detailed error message the user should be able to figure out the issue. Hence removed as suggested.

pkg/chains/signing/kms/kms.go
anithapriyanatarajan: Incorporated

lcarva (approved):

lcarva (approved):
/lgtm

PR #1167, opened 2024-07-29
Make kms path generic to read from a file path instead of `KMS_AUTH_TOKEN`

PuneetPunamiya:
* Previously when `signers.kms.auth.token-dir` was set, then the token was read from the path specified in the field plus `KMS_AUTH_TOKEN` i.e. (for e.g. /tmp/kms-secrets/KMS_AUTH_TOKEN) so user had to create the secret with the key `KMS_AUTH_TOKEN`

* Hence, with this patch now user can specify the path now and token will be read from the path specified

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -145,15 +145,16 @@ chains.tekton.dev/transparency-upload: "true"
 | :-------------------------------- | :------------------------------------------------------------------------------------------ | :--------------- | :------ |
 | `signers.kms.auth.address`        | URI of KMS server (e.g. the value of `VAULT_ADDR`)                                          |                  |
 | `signers.kms.auth.token`          | Auth token KMS server (e.g. the value of `VAULT_TOKEN`)                                     |                  |
-| `signers.kms.auth.token-dir`      | Path to store KMS server Auth token (e.g. `/etc/kms-secrets`)                               |                  |
+| `signers.kms.auth.token-path`     | Path to store KMS server Auth token (e.g. `/etc/kms-secrets`)                               |                  |
 | `signers.kms.auth.oidc.path`      | Path used for OIDC authentication (e.g. `jwt` for Vault)                                    |                  |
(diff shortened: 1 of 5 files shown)


docs/config.md
anithapriyanatarajan, line 1: Please include the changes to https://github.com/tektoncd/chains/blob/main/docs/signing.md#authentication section as well.

chitrangpatel (approved):
Just a catch that this means that this path can come from other volumes too, not just secrets. Is there a way to detect that?

PuneetPunamiya:
> Just a catch that this means that this path can come from other volumes too, not just secrets. Is there a way to detect that?

Good question! I wander how do we it for `signing-secrets` https://github.com/tektoncd/chains/blob/main/config/100-deployment.yaml#L93

lcarva (approved):

lcarva:
> > Just a catch that this means that this path can come from other volumes too, not just secrets. Is there a way to detect that?
> 
> Good question! I wander how do we it for `signing-secrets` https://github.com/tektoncd/chains/blob/main/config/100-deployment.yaml#L93

Maybe Chains doesn't have to be aware of this. It just reads the file from disk however it got there.

lcarva:
@chitrangpatel, since you have an open question. I'll won't add an `/lgtm` comment here.

PuneetPunamiya:
> Just a catch that this means that this path can come from other volumes too, not just secrets. Is there a way to detect that?

I was thinking why do we need to detect that ?

chitrangpatel:
> > Just a catch that this means that this path can come from other volumes too, not just secrets. Is there a way to detect that?
> 
> I was thinking why do we need to detect that ?

Just in case we wanted to ensure that the path was created using a secret instead of some volume. That can be resolved later though. Approving for now.

PR #1170, opened 2024-07-31
Initial proposal to create new counters/timeseries to account for tasknames and namespaces

nistar:
# Changes
Created new const values for a possible future functionality

We're trying to implement the following solution:

Having some metric statistics on container signing, meaning "this many containers were generated, this many were signed, this many weren't". Having a namespace categorization would be helpful.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


PuneetPunamiya:
@nistar Can you please sign the SLA

nistar:
> /ok-to-test



> @nistar Can you please sign the SLA

Done. Please, let me know if I need to provide more info

pkg/chains/constants.go
```diff
@@ -14,24 +14,32 @@ limitations under the License.
...
+	MarkedAsSignedCount          = "mrcount"
+	PipelineRunSignedName        = "pipelinerun_sign_created_total"
+	PipelineRunSignedDesc        = "Total number of signed messages for pipelineruns"
+	PipelineRunUploadedName      = "pipelinerun_payload_uploaded_total"
+	PipelineRunUploadedDesc      = "Total number of uploaded payloads for pipelineruns"
+	PipelineRunStoredName        = "pipelinerun_payload_stored_total"
+	PipelineRunStoredDesc        = "Total number of stored payloads for pipelineruns"
+	PipelineRunMarkedName        = "pipelinerun_marked_signed_total"
+	PipelineRunMarkedDesc        = "Total number of objects marked as signed for pipelineruns"
+	PipelineRunSignedMsg         = "pipelinerun_signed_messages"
```
lcarva, line 29: How's this different than `PipelineRunSignedName` ?

nistar: The ones currently used produce totals across the entire cluster. We need a more detailed information, such as a namespace, taskname, success/failure and other kinds of labels attached to the counters. So my proposal is to leave the old ones alone and create a new time series that can be utilized to display info related to individual projects in Grafana or some other tool.

lcarva: That makes sense. Can we make the name and description indicative of this?

nistar: Do we have a naming convention?

lcarva: Not an official one. I would just like to make it obvious that certain metric is namespace based. Does that make sense?

nistar: Absolutely

nistar: Does it look ok?

nistar: Hi @lcarva do the names look ok to you?

nistar: Yes, just wanted to make small and manageable changes at a time.

nistar:
It looks like we don't honor storage.oci.repository.insecure flag on every execution path and it results in https/http protocol errors when trying to communicate with insecure registries like kind. I've added a few checks to make sure it's implemented. Welcome any comments/suggestions

nistar:
@lcarva @PuneetPunamiya I've modified a couple of files to honor the storage.oci.repository.insecure flag to be able to test with kind registry. What are your thoughts on this?

lcarva:
> @lcarva @PuneetPunamiya I've modified a couple of files to honor the storage.oci.repository.insecure flag to be able to test with kind registry. What are your thoughts on this?

@nistar, I don' think `name.Digest` queries the registry. Have you actually seen this cause an error?

nistar:
> > @lcarva @PuneetPunamiya I've modified a couple of files to honor the storage.oci.repository.insecure flag to be able to test with kind registry. What are your thoughts on this?
> 
> @nistar, I don' think `name.Digest` queries the registry. Have you actually seen this cause an error?

@lcarva The code eventually ends up calling Ping:

image

resulting in "using https with http protocol" error message

lcarva:
@nistar ah! `name.NewDigest` creates a new object that stores the insecure registry option. This is later used in queries to the registry. Good find!

Could you open a separate PR for that? :pray:

anithapriyanatarajan:
Closing this PR, since there is no activity.

PR #1213, opened 2024-09-24
Support mongo server url path

PuneetPunamiya:
* Previously when `storage.docdb.mongo-server-url-dir` was set, then value was read from the path specified in the field plus `MONGO_SERVER_URL` i.e. (for e.g. /mnt/mongo-creds-secret/MONGO_SERVER_URL) so user had to create the secret with the key `MONGO_SERVER_URL`

* Hence, with this patch user can specify the path now and the tokwn will be read from the path specified

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


pkg/chains/storage/docdb/docdb.go
```diff
@@ -262,6 +266,18 @@ func populateMongoServerURL(ctx context.Context, cfg config.Config) error {
 	logger := logging.FromContext(ctx)
 	mongoEnv := "MONGO_SERVER_URL"
```
jkhelil, line 267: looks like we need MONGO_SERVER_URL to put as constant, because it is repeated  in some places

pkg/chains/storage/docdb/docdb.go
```diff
@@ -145,13 +129,23 @@ func WatchBackend(ctx context.Context, cfg config.Config, watcherStop chan bool)
...
+						backendChan <- nil
+					}
+				} else if cfg.Storage.DocDB.MongoServerURLDir != "" {
+					updatedEnv, err = getMongoServerURLFromDir(cfg.Storage.DocDB.MongoServerURLDir)
+					if err != nil {
+						logger.Error(err)
+						backendChan <- nil
+					}
 				}
+
```
jkhelil, line 146: we need MONGO_SERVER_URL as constant, because it is repeated  in some places

pkg/chains/storage/docdb/docdb.go
```diff
@@ -327,3 +343,58 @@ func getMongoServerURLFromDir(dir string) (string, error) {
...
+			// capture via the fsnotify event watcher
+			// filepath.Join(cfg.Storage.DocDB.MongoServerURLPath, "..data"),
+			filepath.Join(dirPath, "..data"),
+		}
+	} else if cfg.Storage.DocDB.MongoServerURLDir != "" {
+		logger.Infof("setting up fsnotify watcher for directory: %s", cfg.Storage.DocDB.MongoServerURLDir)
+		pathsToWatch = []string{
+			// mongo-server-url-dir/MONGO_SERVER_URL is where the MONGO_SERVER_URL environment
+			// variable is expected to be mounted, either manually or via a Kubernetes secret, etc.
+			filepath.Join(cfg.Storage.DocDB.MongoServerURLDir, "MONGO_SERVER_URL"),
```
jkhelil, line 385: same for MONGO_SERVER_URL

docs/config.md
```diff
@@ -93,6 +94,10 @@ This can be achieved in a few ways:
   - This field takes precedence over `storage.docdb.mongo-server-url` and `MONGO_SERVER_URL` env var.
   - The value should point to a directory that has a file named `MONGO_SERVER_URL` that contains the env var. Each time the file is updated, the new value will be read.
   - One common use case is to store the value of `MONGO_SERVER_URL` in a secret with the key `MONGO_SERVER_URL` and mount the secret at the path specified in this field. When the secret is updated, the new value will be fetched by Tekton Chains.
+- Setting the value of `storage.docdb.mongo-server-url-path` field
```
jkhelil, line 97: I am sorry but i find the documentation not clear on how to configure a mondgo databse as dbstore, why not add examples
- configuration using storage.docdb.mongo-server-url-dir
storage.docdb.mongo-server-url-dir: /mnt/mongo-creds-secret
How to mount the secret to have the mondb file configuraiton to match the directory
- configuration using url
storage.docdb.mongo-server-url-path: /mnt/mongo-creds-secret/mongo-server-url

what needs to be done here ? a secret mounted there ?

then we can add priorities
Also why should we have two methods there? shouldnt be just  a configuration using url

docs/config.md
```diff
@@ -93,6 +94,10 @@ This can be achieved in a few ways:
...
+  - This field takes precedence over `storage.docdb.mongo-server-url-dir`, `storage.docdb.mongo-server-url` and `MONGO_SERVER_URL` env var.
+  - This field should point to the file path that contains the mongo server url. Each time the file is updated, the new value will be read
+  - One common use case is to store the value of mongo server url in a secret with any key and mount the secret at the path specified. When the secret is updated, the new value will be fetched by Tekton Chains
```
savitaashture, line 101: It would be helpful if you create a table which shows precedence of using `storage.docdb.mongo-server-url-dir`, `storage.docdb.mongo-server-url` and `MONGO_SERVER_URL`

pkg/chains/storage/docdb/docdb.go
```diff
@@ -262,6 +266,18 @@ func populateMongoServerURL(ctx context.Context, cfg config.Config) error {
 	logger := logging.FromContext(ctx)
 	mongoEnv := "MONGO_SERVER_URL"
 
+	if cfg.Storage.DocDB.MongoServerURLPath != "" {
+		logger.Infof("setting %s from storage.docdb.mongo-server-url-path: %s", mongoEnv, cfg.Storage.DocDB.MongoServerURLPath)
```
savitaashture, line 270: It would be good to use constant `docDBMongoServerURLPathKey` instead of key directly

pkg/chains/storage/docdb/docdb.go
```diff
@@ -262,6 +266,18 @@ func populateMongoServerURL(ctx context.Context, cfg config.Config) error {
...
+			return err
+		}
+		if err := os.Setenv(mongoEnv, mongoServerURL); err != nil {
+			return err
+		}
+		return nil
+	}
+
 	if cfg.Storage.DocDB.MongoServerURLDir != "" {
 		logger.Infof("setting %s from storage.docdb.mongo-server-url-dir: %s", mongoEnv, cfg.Storage.DocDB.MongoServerURLDir)
```
savitaashture, line 282: It would be good to use constant `docDBMongoServerURLDirKey` instead of key directly

pkg/chains/storage/docdb/docdb.go
```diff
@@ -327,3 +343,58 @@ func getMongoServerURLFromDir(dir string) (string, error) {
...
+		logger.Infof("setting up fsnotify watcher for mongo server url path: %s", cfg.Storage.DocDB.MongoServerURLPath)
+		dirPath := filepath.Dir(cfg.Storage.DocDB.MongoServerURLPath)
+		pathsToWatch = []string{
+			// mongo-server-url-path/<path> is where the mongo server url token
+			// When a Kubernetes secret is mounted on a path, the `data` in that secret is mounted
+			// under path/..data that is then `symlink`ed to the key of the data. In this instance,
+			// the mounted path is going to look like:
+			// file 1 - ..2024_05_03_11_23_23.1253599725
+			// file 2 - ..data -> ..2024_05_03_11_23_23.1253599725
+			// file 3 - MONGO_SERVER_URL -> ..data/<path>
```
savitaashture, line 373: Is it still MONGO_SERVER_URL or its a `<path>`  :thinking:

docs/config.md
PuneetPunamiya: Yeah make sense on adding examples 👍🏻 
Will update the docs 

> Also why should we have two methods there? shouldnt be just a configuration using url

Because we cannot remove this field directly to maintain backward compatibility

docs/config.md
```diff
@@ -62,16 +62,17 @@ Supported keys include:
...
+| `storage.docdb.mongo-server-url-path` (optional) | The path of the file that contains that stores the value of mongo server url                                                                                                                                                                                                                                  | If the file `/mnt/mongo-creds-secret/mongo-server-url` has the value, then set `storage.docdb.mongo-server-url-path: /mnt/mongo-creds-secret/mongo-server-url`                                                                                                                                                                                                                                                                                                      |         |
```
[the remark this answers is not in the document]
PuneetPunamiya: User can have different path right, if we make that generic in that case it adds limits for the user to specify the same path which he mentioned for the directory

pkg/chains/storage/docdb/docdb.go
```diff
@@ -145,13 +129,23 @@ func WatchBackend(ctx context.Context, cfg config.Config, watcherStop chan bool)
...
+					updatedEnv, err = getMongoServerURLFromDir(cfg.Storage.DocDB.MongoServerURLDir)
+					if err != nil {
+						logger.Error(err)
+						backendChan <- nil
+					}
 				}
+
 				if updatedEnv != os.Getenv("MONGO_SERVER_URL") {
-					logger.Infof("directory %s has been updated, reconfiguring backend...", cfg.Storage.DocDB.MongoServerURLDir)
+					logger.Infof("directory %s has been updated, reconfiguring backend...", cfg.Storage.DocDB.MongoServerURLPath)
```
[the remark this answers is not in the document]
PuneetPunamiya: It should not be file neither directory, will update the log

docs/config.md
```diff
@@ -62,16 +62,17 @@ Supported keys include:
...
+| `storage.docdb.mongo-server-url` (optional)      | The value of MONGO_SERVER_URL env var with the MongoDB connection URI                                                                                                                                                                                                                                               | Example: `mongodb://[USER]:[PASSWORD]@[HOST]:[PORT]/[DATABASE]`                                                                                                                                                                                                                                                                                                                                                                                                     |         |
```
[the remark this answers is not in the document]
jkhelil: @PuneetPunamiya do we need both ? why not just have mongo-server-url-path, by it self it means the path for the url of mongo, i would even call this mongo-server-connection-path because i guess it has user/password and url, It is acceptable to keep the name but i find having both mongo-server-url-dir and mongo-server-url-path is a bit confusing, and it does really bring a real value to have mongo-server-url-dir

docs/config.md
jkhelil: @PuneetPunamiya I still find this text very confusing
can we break down it like this way or similar
 You can provide MongoDB connection through different options 
- Using MONGO_SERVER_URL Environment Variable
User can set the MongoDB connection URL in the MONGO_SERVER_URL env var in the Chains deployment.
- Using storage.docdb.mongo-server-url field in the Chains configmap
You can set the connection URL using the storage.docdb.mongo-server-url field in the Chains configmap
This field overrides the MONGO_SERVER_URL env var.
- Using storage.docdb.mongo-server-url-dir Field
Another option is to set storage.docdb.mongo-server-url-dir, which points to a directory containing a file named MONGO_SERVER_URL. 
The directory path setting takes precedence over both storage.docdb.mongo-server-url and the MONGO_SERVER_URL env var.
For instance, if /mnt/mongo-creds-secret/MONGO_SERVER_URL contains the MongoDB URL, set storage.docdb.mongo-server-url-dir: /mnt/mongo-creds-secret.
- Using storage.docdb.mongo-server-url-path Field
You can use. storage.docdb.mongo-server-url-path field in Chains configmap to directly reference the file containing the MongoDB URL.
This field overrides all others (mongo-server-url-dir, mongo-server-url, and MONGO_SERVER_URL env var).

PuneetPunamiya: Updated, can you please take a look, thanks!

jkhelil (approved):

docs/config.md
lcarva: And also we should make a recommendation to users on which approach is recommended. For example, `storage.docdb.mongo-server-url` should be avoided since credentials are stored in a ConfigMap instead of a Secret.

pkg/chains/storage/docdb/docdb.go
```diff
@@ -145,13 +128,23 @@ func WatchBackend(ctx context.Context, cfg config.Config, watcherStop chan bool)
...
+					updatedEnv, err = getMongoServerURLFromDir(cfg.Storage.DocDB.MongoServerURLDir)
+					if err != nil {
+						logger.Error(err)
+						backendChan <- nil
+					}
 				}
+
 				if updatedEnv != os.Getenv("MONGO_SERVER_URL") {
-					logger.Infof("directory %s has been updated, reconfiguring backend...", cfg.Storage.DocDB.MongoServerURLDir)
+					logger.Infof("Mongo server url %s has been updated, reconfiguring backend...", cfg.Storage.DocDB.MongoServerURLPath)
```
lcarva, line 147: We don't know at this point if it was updated because of a change in `cfg.Storage.DocDB.MongoServerURLDir` or `cfg.Storage.DocDB.MongoServerURLPath`. Maybe just don't include the path at all in the log message?

pkg/chains/storage/docdb/docdb.go
```diff
@@ -179,11 +172,21 @@ func WatchBackend(ctx context.Context, cfg config.Config, watcherStop chan bool)
...
-	if err != nil {
-		return nil, err
+	if cfg.Storage.DocDB.MongoServerURLPath != "" {
+		dirPath := filepath.Dir(cfg.Storage.DocDB.MongoServerURLPath)
+		// Add a path.
+		err = watcher.Add(dirPath)
+		if err != nil {
+			return nil, err
+		}
+	} else if cfg.Storage.DocDB.MongoServerURLDir != "" {
```
lcarva, line 182: I'm wondering if this should just be an `if-block` instead of a `else-if-block`. I'm thinking of the case where the config changes. It's probably safer to watch all the paths that are not an empty string.

pkg/chains/storage/docdb/docdb.go
```diff
@@ -260,7 +263,18 @@ func populateMongoServerURL(ctx context.Context, cfg config.Config) error {
 	// If that doesn't exist, then we move on to `storage.docdb.mongo-server-url`.
```
lcarva, line 263: These docs are out of date.

pkg/chains/storage/docdb/docdb.go
```diff
@@ -260,7 +263,18 @@ func populateMongoServerURL(ctx context.Context, cfg config.Config) error {
...
+	if cfg.Storage.DocDB.MongoServerURLPath != "" {
+		logger.Infof("setting %s from storage.docdb.mongo-server-url-path: %s", mongoEnv, cfg.Storage.DocDB.MongoServerURLPath)
+		mongoServerURL, err := getMongoServerURLFromPath(cfg.Storage.DocDB.MongoServerURLPath)
+		if err != nil {
+			return err
+		}
+		if err := os.Setenv(mongoEnv, mongoServerURL); err != nil {
+			return err
+		}
+		return nil
```
lcarva, line 276: For consistency, I think we should either `return nil` in the other if-block in this function:

```go
if cfg.Storage.DocDB.MongoServerURLPath != "" {
  ...
  return nil
}

if cfg.Storage.DocDB.MongoServerURLDir != "" {
  ...
  return nil
}

if cfg.Storage.DocDB.MongoServerURL != "" {
  ...
  return nil
}
```

or chain them with if-else-blocks:

```go
if cfg.Storage.DocDB.MongoServerURLPath != "" {
  ...
} else if cfg.Storage.DocDB.MongoServerURLDir != "" {
  ...
} else if cfg.Storage.DocDB.MongoServerURL != "" {
  ...
}
```

pkg/chains/storage/docdb/docdb.go
```diff
@@ -288,15 +302,15 @@ func setMongoServerURLFromDir(dir string) error {
...
+	if err = os.Setenv(mongoEnv, fileDataNormalized); err != nil {
 		return err
 	}
 
 	return nil
 }
 
 func getMongoServerURLFromDir(dir string) (string, error) {
-	mongoEnv := "MONGO_SERVER_URL"
+	// mongoEnv := "MONGO_SERVER_URL"
```
lcarva, line 313: Let's remove it.

pkg/chains/storage/docdb/docdb.go
```diff
@@ -327,3 +341,58 @@ func getMongoServerURLFromDir(dir string) (string, error) {
...
+
+// getMongoServerUrlFromPath retreives token from the given mount path
+func getMongoServerURLFromPath(path string) (string, error) {
+	fileData, err := os.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("reading file in %q: %w", path, err)
+	}
+
+	// A trailing newline is fairly common in mounted files, so remove it.
+	fileDataNormalized := strings.TrimSuffix(string(fileData), "\n")
```
lcarva, line 353: Good idea! Consider even using [string.TrimSpace](https://pkg.go.dev/strings#TrimSpace).

If so, let's do that in `getMongoServerURLFromDir` as well.

(Ok to ignore this comment since it's an already existing pattern.)

pkg/chains/storage/docdb/docdb.go
```diff
@@ -327,3 +341,58 @@ func getMongoServerURLFromDir(dir string) (string, error) {
...
+			// file 1 - ..2024_05_03_11_23_23.1253599725
+			// file 2 - ..data -> ..2024_05_03_11_23_23.1253599725
+			// file 3 - ..data/<path>
+			// So each time the secret is updated, the file is not updated,
+			// instead the underlying symlink at `..data` is updated and that's what we want to
+			// capture via the fsnotify event watcher
+			// filepath.Join(cfg.Storage.DocDB.MongoServerURLPath, "..data"),
+			filepath.Join(dirPath, "..data"),
+		}
+	} else if cfg.Storage.DocDB.MongoServerURLDir != "" {
```
lcarva, line 378: Similar comment as a previous one. Let's make this an `if-block` instead of an `else-if-block`.

pkg/chains/storage/docdb/docdb.go
```diff
@@ -179,11 +172,21 @@ func WatchBackend(ctx context.Context, cfg config.Config, watcherStop chan bool)
 		}
 	}()
 
-	// Add a path.
-	err = watcher.Add(cfg.Storage.DocDB.MongoServerURLDir)
-	if err != nil {
-		return nil, err
+	if cfg.Storage.DocDB.MongoServerURLPath != "" {
```
lcarva, line 175: Can we leverage `pathsToWatch` here instead?

pkg/chains/storage/docdb/docdb.go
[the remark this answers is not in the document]
savitaashture: Thanks for updating
earlier it was not that clear

pkg/chains/storage/docdb/docdb.go
savitaashture: I see its resolved but i don't see changes :thinking:

pkg/chains/storage/docdb/docdb.go
savitaashture: Same here

pkg/chains/storage/docdb/docdb.go
PuneetPunamiya: updated!

pkg/chains/storage/docdb/docdb.go
PuneetPunamiya: updated!

docs/config.md
PuneetPunamiya: updated!

pkg/chains/storage/docdb/docdb.go
PuneetPunamiya: Good catch, thanks 👍🏻

docs/config.md
```diff
@@ -83,16 +84,28 @@ You can read about the go-cloud docstore URI format [here](https://gocloud.dev/h
...
+  * For instance, if `/mnt/mongo-creds-secret/MONGO_SERVER_URL` contains the MongoDB URL, set `storage.docdb.mongo-server-url-dir`: `/mnt/mongo-creds-secret`
+
+* Using `storage.docdb.mongo-server-url-path` field
+  * You can use `storage.docdb.mongo-server-url-path` field in chains-config configmap to directly reference the file containing the MongoDB URL
+  * This field overrides all others (`mongo-server-url-dir, mongo-server-url, and MONGO_SERVER_URL env var`)
+  * For instance, if `/mnt/mongo-creds-secret/mongo-server-url`  contains the MongoDB URL, then set `storage.docdb.mongo-server-url-path`: `/mnt/mongo-creds-secret/mongo-server-url`
+
+**NOTE** :- 
+* When using `storage.docdb.mongo-server-url-dir` or `storage.docdb.mongo-server-url-path` field, store the value of mongo server url in a secret and mount the secret. When the secret is updated, the new value will be fetched by Tekton Chains controller
+* Also using `storage.docdb.mongo-server-url-dir` or `storage.docdb.mongo-server-url-path` field are recommended, using `storage.docdb.mongo-server-url-dir` should be avoided since credentials are stored in a ConfigMap instead of a secret
```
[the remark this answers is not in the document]
PuneetPunamiya: Good catch, thanks 👍🏻

lcarva (approved):
/lgtm

PR #1316, opened 2025-02-27
Add Archivista Storage Backend

mikhailswift:
# Changes

Adds the ability to store signed TaskRun and PipelineRun results in Archivista. Archivista currently only supports payloads that are wrapped in a DSSE envelope, so any signatures that are not DSSE envelopes will not be stored.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Signed attestations are now able to be stored in Archivista
```


mikhailswift:
Tagging @colek42 for visibility

colek42:
@PuneetPunamiya do you see any issues with this PR?  Let us know if you would like to discuss at a community meeting.  FYSA Archivista is an intoto project which is CNCF graduated.

pkg/chains/storage/archivista/archivista.go
```diff
@@ -0,0 +1,134 @@
...
+		return err
+	}
+	logger.Infof("Successfully uploaded DSSE envelope to Archivista, response: %+v", uploadResp)
+
+	// Update the Tekton object with Archivista annotations.
+	annotations := map[string]string{
+		"chains.tekton.dev/archivista-gitoid": uploadResp.Gitoid,
+		"chains.tekton.dev/archivista-url":    a.url,
+	}
+	obj.SetAnnotations(annotations)
```
anithapriyanatarajan, line 107: @lcarva  - right now we do not set any annotation on tekton objects post storage with other backend types. Is there any concern having the info captured ?

lcarva: I haven't reviewed the PR yet, but this is what comes to mind:

Chains will add annotations once it is done processing a Tekton object (TaskRun or PipelineRun). That happens right at the *end* of reconciliation. This update triggers another reconciliation but that's generally ok because Chains quickly detects the object has already been processed (based on it annotations) and skips that iteration of the reconciliation loop.

If Chains updates the Tekton object before those final annotations are added, the reconciliation that is triggered as a side-effect will cause signing and attesting to happen again. This may happen multiple times.

I would advise against changing the object during reconciliation. If needed, try tying into the already existing mechanism Chains has for adding the final annotations, see [here](https://github.com/tektoncd/chains/blob/7824cdc5b20e1bffb05453e7360fbfbc5ea45640/pkg/chains/signing.go#L241).

NOTE: The `tekton` storage does update the object during reconciliation. We should probably fix that, but that's less of a concern given that that storage type is meant for test/develop.

bradbeck: @lcarva Are you suggesting that `tekton` storage should not be used in production? Isn't it included in some of the default configuration?

colek42: It might be best for us to remove this for now and follow up with the annotation in a separate PR.  cc @mikhailswift

lcarva: > Are you suggesting that tekton storage should not be used in production?

Yes, 100%. This storage stores payloads as annotations which can get quite large if you have a pipeline with many tasks, for example. You'll face etcd issues eventually.

> Isn't it included in some of the default configuration?

Yes, it's not ideal. We could pursue changing this default to OCI which is a more suitable storage type that requires no configuration if you are already building container images.

mikhailswift: The annotations have been removed from the Archivista storage backend

pkg/chains/storage/archivista/archivista.go
```diff
@@ -0,0 +1,84 @@
+package archivista
```
PuneetPunamiya, line 1: Can you please add license here

pkg/chains/storage/archivista/archivista.go
```diff
@@ -0,0 +1,84 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	archivistaClient "github.com/in-toto/archivista/pkg/http-client"
+	"github.com/in-toto/go-witness/dsse"
+	"github.com/tektoncd/chains/pkg/chains/objects"
+	"github.com/tektoncd/chains/pkg/config" // if needed
```
PuneetPunamiya, line 14: Do we need the comment added here ?

pkg/chains/storage/archivista/archivista.go
```diff
@@ -0,0 +1,84 @@
...
+	return &Backend{
+		client: client,
+		url:    archCfg.URL,
+		cfg:    archCfg,
+	}, nil
+}
+
+// StorePayload attempts to parse `signature` as a DSSE envelope, and if successful
+// sends it to an Archivista server for storage.
+func (a *Backend) StorePayload(ctx context.Context, obj objects.TektonObject, rawPayload []byte, signature string, opts config.StorageOpts) error {
```
PuneetPunamiya, line 50:
```suggestion
func (b *Backend) StorePayload(ctx context.Context, obj objects.TektonObject, rawPayload []byte, signature string, opts config.StorageOpts) error {
```

pkg/chains/storage/archivista/archivista.go
```diff
@@ -0,0 +1,84 @@
...
+	uploadResp, err := a.client.Store(ctx, env)
+	if err != nil {
+		logger.Errorw("Failed to upload DSSE envelope to Archivista", "error", err)
+		return err
+	}
+	logger.Infof("Successfully uploaded DSSE envelope to Archivista, response: %+v", uploadResp)
+	return nil
+}
+
+// RetrievePayload is not implemented for Archivista.
```
PuneetPunamiya, line 67: I think it would be nice to create an issue in Archivista project where it can have a function to retrieve the data 

cc @lcarva @wlynch

pkg/chains/storage/archivista/archivista_test.go
```diff
@@ -0,0 +1,129 @@
+package archivista
```
PuneetPunamiya, line 1: Can you please add license here ?

pkg/config/config.go
```diff
@@ -155,6 +156,12 @@ type TransparencyConfig struct {
 	URL              string
 }
 
+// ArchivistaStorageConfig holds configuration for the Archivista storage backend.
+type ArchivistaStorageConfig struct {
+	// URL is the endpoint for the Archivista service.
+	URL string `json:"url,omitempty"`
```
PuneetPunamiya, line 162: I wander if if need `omitempty` json tags ?

PuneetPunamiya:
One thing I would encourage is if you can submit a follow up pr to add a small tutorial for using the storage backend you have added for example - https://github.com/tektoncd/chains/tree/main/docs/tutorials

pkg/chains/storage/archivista/archivista.go
mikhailswift: I think the main thing we need here is details of what Tekton uses to retrieve the data. I looked into it a bit and I am a bit unclear on what the `ShortKey` is that is being used to find the payload.

Archivista can search for payloads with it's GraphQL API, and can download the DSSE envelope. From Archivista's side of things we need to make sure we are indexing the ShortKey or some other identifier that Tekton will use to retrieve the payload.

mikhailswift:
I addressed the few comments here -- If I can get clarification on what `ShortKey` is in regards to the retrieval functions I can add the indexes to Archivista to support retrieval.

I can add that and a tutorial in a followup PR.

mikhailswift:
Re-pushed to address the linting job errors

pkg/chains/storage/archivista/archivista.go
anithapriyanatarajan: @mikhailswift - Please lookup -  https://github.com/tektoncd/chains/blob/main/pkg/config/options.go#L26C1-L36C1.  Does this help?


	// ShortKey is the short version of an artifact identifier. This is useful for annotation based storage
	// because annotation key has limitations (https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set).
	// - For OCI artifact, it is first 12 chars of the image digest.
	// - For TaskRun/PipelineRun artifact, it is `<KIND>-<UID>`.
	ShortKey string

mikhailswift:
Hi @PuneetPunamiya , just wanted to check if there's anything I need to do yet for this PR. As discussed I will follow up with a PR that includes a tutorial.

PuneetPunamiya:
> Hi @PuneetPunamiya , just wanted to check if there's anything I need to do yet for this PR. As discussed I will follow up with a PR that includes a tutorial.

Sorry for the late response
The PR looks good to me 
/approve

cc @lcarva @chitrangpatel

afrittoli:
@tektoncd/chains-maintainers this is ready for a second reviewer

pkg/chains/storage/archivista/archivista.go
lcarva: FYI, the `Retrieve*` functions are only used as part of the tkn chain CLI tool. They're not that useful and I'd be shocked if anyone actually uses it. So I think it's fine to leave these as not implemented.

lcarva (approved):

docs/config.md
```diff
@@ -73,6 +73,7 @@ Supported keys include:
...
+| `storage.archivista.url`         | The URL endpoint for the Archivista service.                                                                                                                                                                                                                                                                     | A valid HTTPS URL pointing to your Archivista instance (e.g. `https://archivista.testifysec.io`).                                                                                                                                                                                                                                                                                                                      | None |
```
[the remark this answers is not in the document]
lcarva: ack. that explains it ty

PR #1391, opened 2025-06-30
Add path to taskrun finalizer name

enarha:
Fixes the following warning:
"API Warning: metadata.finalizers: \"chains.tekton.dev\": prefer a domain-qualified finalizer name including a path (/) to avoid accidental conflicts with other finalizer writers".

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Finalizer name for TaskRuns now contains a path to avoid API warnings, "chains.tekton.dev/taskrun".
```

/kind misc

### DIFF
--- a/pkg/reconciler/taskrun/controller.go
+++ b/pkg/reconciler/taskrun/controller.go
@@ -86,7 +86,7 @@ func NewNamespacesScopedController(namespaces []string) func(ctx context.Context
 				// The chains reconciler shouldn't mutate the taskrun's status.
 				SkipStatusUpdates: true,
 				ConfigStore:       cfgStore,
-				FinalizerName:     "chains.tekton.dev",
+				FinalizerName:     "chains.tekton.dev/taskrun",
 			}
 		})
 


khrm:
@lcarva @vdemeester Isn't this a breaking change?

enarha:
> @lcarva @vdemeester Isn't this a breaking change?

If you believe that there are other controllers or clients managing or using that finalizer, that will be a breaking change for them. I find this quite unlikely though. The finalizer name is also not part of the CR specification, so another reason to not consider that a breaking change. That's IMO.

khrm:
If we deploy the latest chains controller, it could leave the older finalizer. Won't that cause an issue? User would need to manually delete the old finalizers. @enarha

enarha:
> > > If we deploy the latest chains controller, it could leave the older finalizer. Won't that cause an issue? User would need to manually delete the old finalizers. @enarha
> > 
> > 
> > Yes, you are right. During upgrade, existing finalizers won't be removed by the new controller instance. The user will have to remove them manually. Another option is we add some extra code, which only removes the old finalizer if found.
> 
> I created #1394 to address this issue. I need some more testing and I'll remove the WIP once done.

#1394 is ready for review. @khrm @vdemeester @lcarva PTAL.

PR #1395, opened 2025-07-02
Handle signing and attesting blobs

lcarva:
# Changes
Chains fetches existing information about the container images it signs and attests. This works well when the image reference refers to an Image Index or an Image Manifest since those are served from the same endpoint in a container registry. However, when fetching information about a blob (layer) a different endpoint is needed.

Cosign handles this behavior by making this step optional: https://github.com/sigstore/cosign/blob/c86498055d0c4ea2f39076064aa094db12f85f6a/cmd/cosign/cli/sign/sign.go#L181-L186

Thus it is possible to sign/attest a blob with the cosign CLI. This commit implements the same logic to Chains so it can also sign/attest blobs.
# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Signing and attesting blobs is now supported.
```

### DIFF
--- a/pkg/chains/storage/oci/attestation.go
+++ b/pkg/chains/storage/oci/attestation.go
@@ -61,7 +61,10 @@ func (s *AttestationStorer) Store(ctx context.Context, req *api.StoreRequest[nam
 		repo = *s.repo
 	}
 	se, err := ociremote.SignedEntity(req.Artifact, ociremote.WithRemoteOptions(s.remoteOpts...))
-	if err != nil {
+	var entityNotFoundError *ociremote.EntityNotFoundError
+	if errors.As(err, &entityNotFoundError) {
+		se = ociremote.SignedUnknown(req.Artifact)
+	} else if err != nil {
 		return nil, errors.Wrap(err, "getting signed image")
 	}
 
--- a/pkg/chains/storage/oci/attestation_test.go
+++ b/pkg/chains/storage/oci/attestation_test.go
@@ -1,0 +1,111 @@
+// Copyright 2025 The Tekton Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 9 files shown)


pkg/chains/storage/oci/attestation.go
```diff
@@ -61,7 +61,9 @@ func (s *AttestationStorer) Store(ctx context.Context, req *api.StoreRequest[nam
 		repo = *s.repo
 	}
 	se, err := ociremote.SignedEntity(req.Artifact, ociremote.WithRemoteOptions(s.remoteOpts...))
-	if err != nil {
+	if _, isEntityNotFoundErr := err.(*ociremote.EntityNotFoundError); isEntityNotFoundErr {
```
anithapriyanatarajan, line 64: Whenever there is an entity not found we are default considering it as a blob, what if the digest was incorrect or the blob is not there? Is there a possibility to check if such a blob exists before attesting?

lcarva: The code is not necessarily handling a blob but it's handling the case where the image reference does not reference an existing Image Manifest or Image Index. It just so happens that this is what makes the blob signing/attesting work. I linked in the description how cosign itself implements this logic. I think it's safe to copy what they're doing there.

lcarva:
> @lcarva Could you help clarify the scenario in which Tekton PRs or TRs would result in standalone blobs being pushed to container registries (i.e., not part of an image or manifest)?
> 
> From my understanding, typical Tekton outputs are full OCI images. Are there cases where individual blobs are pushed independently and need signing or attestation support?
> 
> This context would help in understanding the need for handling blob references explicitly in this logic.

It's really up to a Task. With something like [oras](https://oras.land/), I can push individual blobs to the registry and emit those references in my Task result.

The use case I'm exploring that highlighted this was that I was using a Tekton Task to build a python wheel. I want to make sure the digest of the wheel (same as the digest of the blob) is in the SLSA Provenance.

pkg/chains/storage/oci/attestation_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+			ref := tt.writeToRegistry(t, registryName)
+
+			storer, err := NewAttestationStorer(WithTargetRepository(ref.Repository))
+			if err != nil {
+				t.Fatalf("failed to create storer: %v", err)
+			}
+
+			ctx := logtesting.TestContextWithLogger(t)
+			_, err = storer.Store(ctx, &api.StoreRequest[name.Digest, *intoto.Statement]{
```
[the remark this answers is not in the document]
lcarva: I don't think that's important. It looks like the Store method is half-baked and always returns `&api.StoreResponse{}`.

PR #1422, opened 2025-08-23
Refactor the annotations management

enarha:
Annotations management moved to a separate package, so it can be imported by other packages and avoid cycling import. The tekton package now imports the annotations package and uses the AddAnnotations function, while previously it duplicated the same functionality.
The only functionlity previously provided by the patch package was the CreateAnnotationsPatch function, which is now moved to the annotations package and the patch package removed.
No new functionality added or removed.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


pkg/chains/annotations/annotations.go
```diff
@@ -127,13 +139,7 @@ func AddAnnotation(ctx context.Context, obj objects.TektonObject, ps versioned.I
 		mergedAnnotations[k] = v
 	}
 
-	// Add the specific key-value pair, again it must be chains annotation
-	if !strings.HasPrefix(key, ChainsAnnotationPrefix) {
-		return fmt.Errorf("invalid annotation key %q: all annotations must have prefix %q", key, ChainsAnnotationPrefix)
```
waveywaves, line 132: I think we still keep this validation just make sure that we continue to be defensive wrt there being a prefix.

pkg/chains/annotations/annotations.go
```diff
@@ -148,3 +154,53 @@ func AddAnnotation(ctx context.Context, obj objects.TektonObject, ps versioned.I
...
+func CreateAnnotationsPatch(newAnnotations map[string]string, obj objects.TektonObject) ([]byte, error) {
+	// Get GVK using the TektonObject interface method (more reliable than runtime.Object)
+	gvkStr := obj.GetGVK()
+	if gvkStr == "" {
+		return nil, fmt.Errorf("unable to determine GroupVersionKind for object %s/%s", obj.GetNamespace(), obj.GetName())
+	}
+
+	// Parse the string format "group/version/kind"
+	parts := strings.Split(gvkStr, "/")
+	if len(parts) != 3 {
```
waveywaves, line 176:
```suggestion
	if len(parts) != 3 || parts[0] == "" || parts[1] == "" || parts[2] == "" {
```

just to make sure that we don't have silent failures because of the string concatenation done on the subsequent lines

waveywaves (changes requested):
moving annotation handling to a new package, the annotation management looks more consistent and organized now. The only concern to have here is that that patch package is removed and if htere are other dependants on it which I don't think is a big deal but just something I thought I would put out there. Just have two suggestions for now over here after reviewing.

pkg/chains/annotations/annotations.go
enarha: The `key` variable does not exist anymore, that's why the check was removed.
Initially the same function could have received annotations in two ways, a key/value pair or a list of annotations (a map). To simplify this, I removed the first options and instead added a helper function `mergeAnnotations` which deals with that scenario. And the above check still exists as part of the validation of the annotations provided as a map.

pkg/chains/annotations/annotations.go
enarha: added

PR #1464, opened 2025-11-05
fix: replace HS256 token with RS256

anithapriyanatarajan:
# Changes

This pull request updates the JWT token used in the `x509_test.go` test file to use the RS256 algorithm instead of HS256. This change ensures compatibility with cosign v2.6.0 and newer, which require RS256-signed tokens. 

# Submitter Checklist

- [X] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [X] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
  functionality, content, code)
- [X] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [X] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

``` release-note

Breaking Change: 

Due to the upgrade of Cosign from v2.5.3 to v2.6.0 (PR #1441), Tekton Chains now requires identity tokens for Fulcio signing to use the RS256 algorithm instead of HS256 or other symmetric algorithms.

Impact:

If you are using Fulcio for keyless signing with custom identity tokens (via signers.x509.identity.token.file configuration), you must ensure your tokens are signed using the RS256 (RSA with SHA-256) algorithm.
Tokens using HS256, HS384, HS512, or other non-RS256 algorithms will be rejected, causing signature operations to fail with errors like: new signer: reading id token: getting id token: open <token>: no such file or directory

Who is affected:

- Users who provide custom JWT identity tokens for Fulcio authentication
- Users who have implemented custom OIDC token providers with non-RS256 algorithms
- Default Kubernetes service account tokens and standard OIDC providers (Google, GitHub, etc.) are not affected as they already use RS256

Action required:

Update your identity token generation to use RS256 algorithm. If using custom token generation scripts, ensure they sign with RSA keys
```

### DIFF
--- a/pkg/chains/signing/x509/x509_test.go
+++ b/pkg/chains/signing/x509/x509_test.go
@@ -49,8 +49,10 @@ const ed25519Priv = `-----BEGIN PRIVATE KEY-----
 MC4CAQAwBQYDK2VwBCIEIGQn0bJwshjwuVdnd/FylMk3Gvb89aGgH49bQpgzCY0n
 -----END PRIVATE KEY-----`
 
-// npx jwtgen -a HS256 -s "my-secret" -c "iss=user123" -e 3600
-const token = `eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2Nzc1NjAyMTgsImV4cCI6MTY3NzU2MzgxOCwiaXNzIjoidXNlcjEyMyJ9.c-sDgCyuZA6VaIGl7Y3-9XxttW1PUkBeNBLE9gCKG8s`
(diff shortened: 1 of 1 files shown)


anithapriyanatarajan:
Screen shot of local unit test that change of token fixes the issue reported in while executing `go test -v -run "^TestCreateSignerFulcioEnabled$" ./pkg/chains/signing/x509/` in PR `https://github.com/tektoncd/chains/pull/1463`

Screenshot From 2025-11-05 09-25-22

Issue #1178, opened 2024-08-19
TestWatchBackend/verify_mongo-server-url-dir/MONGO_SERVER_URL_is_watched is buggy


lcarva:
The test `TestWatchBackend/verify_mongo-server-url-dir/MONGO_SERVER_URL_is_watched` sometimes fails. This is likely because the fsnotify handler is not executed within the expected time.

lcarva:
Re-opening this as I'm still seeing the issue.

PR #1505, opened 2025-12-31
Fix- Update Docdb storage logic (issue #1178)

ngelman1:
This PR fixes issue #1178 
The issue was that  the DocDB watcher in Tekton Chains did not reliably detect changes to the storage.docdb.mongo-server-url-dir path because it was matching fsnotify events only on exact file names (slices.Contains), but when a secret rotates, only the symlink changes. This prevented the backend from being reconfigured when Kubernetes rotated mounted secrets, since fsnotify emits events on the underlying symlinked directories rather than the exact watched filename.

# Motivation
Kubernetes secrets are updated by creating new timestamped directories and atomically switching a ..data symlink rather than modifying the file path in place. The old watcher logic used exact name matching, so fsnotify events for these symlinked paths were ignored, resulting in missed reconfiguration and broken behavior for the DocDB storage backend. With prefix matching, fsnotify events inside the watched directory are detected properly.

# Changes

- Updated watcher logic in WatchBackend to use prefix matching (strings.HasPrefix) instead of exact path containment for filtering fsnotify events on watched paths.

- Re-enabled the previously commented test in TestWatchBackend to verify that updates to MongoServerURLDir are watched and trigger backend reconfiguration.

- Added critical case to the `TestBackend_StorePayload` function to test the PATH scenario, to assure it gets the highest priority

# Submitter Checklist

- [X] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
  functionality, content, code)

# Release Notes

``` release-note
Fixed DocDB backend watcher to correctly detect file system changes under MongoServerURLDir, enabling reliable secret rotation support without requiring restarts.
```

### DIFF
--- a/pkg/chains/storage/docdb/docdb.go
+++ b/pkg/chains/storage/docdb/docdb.go
@@ -21,7 +21,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"slices"
 	"strings"
 
 	"github.com/fsnotify/fsnotify"
@@ -124,7 +123,15 @@ func WatchBackend(ctx context.Context, cfg config.Config, watcherStop chan bool)
 					continue
 				}
 
-				if !slices.Contains(pathsToWatch, event.Name) {
+				// Checking if the event.name matches ANY of the pathsToWatch
+				matched := false
+				for _, p := range pathsToWatch {
+					if strings.HasPrefix(event.Name, p) {
+						matched = true
+						break
+					}
+				}
+				if !matched {
 					continue
 				}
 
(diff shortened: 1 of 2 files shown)


anithapriyanatarajan:
@ngelman1 - Thanks for the PR! A few thoughts:

On HasPrefix approach: Though it might capture additional events and retrigger backend configuration updates unlike contains, I am ok to merge this PR since it's still a safe option. The key benefit is reliability over efficiency.

On test flakiness: The main concern for test failures based on fsnotify events would be timing - specifically, the delay between the secret edit and the test assertion. If we observe further flakiness in the tests, it could be due to delays in fsnotify events when a file change is identified. In that case, we should introduce a minimal number of retries before the test asserts as failed.

Let us monitor the issue https://github.com/tektoncd/chains/issues/1178 for a few more days and then close.

Would it be possible to include a MongoDB-based storage tutorial as a separate PR to the docs section. Please refer to the [gist](https://gist.github.com/anithapriyanatarajan/1f2dd3d0e2b86b18a41db35bac4f0429) for reference

PR #1578, opened 2026-03-06
Handle signing OCI artifacts in *ARTIFACT_OUTPUTS

bradbeck:
# Changes

Handle signing OCI artifacts that are hinted using *ARTIFACT_OUTPUTS

Fixes #1575

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

``` release-note
Sign OCI artifacts hinted by `*ARTIFACT_OUTPUTS` and `isbuildArtifact: true`
```

PR #1601, opened 2026-03-28
Fix duplicate .att/.sig OCI layers for same digest type hints

ab-ghosh:
# Changes

 When a Task declares the same OCI image through multiple type-hint formats, produces duplicate layers in the `.att` and/or `.sig` manifests.

 This fixes the issue at two levels:
 1. **Extraction-level dedup** — Added a `seen` map in `ExtractOCIImagesFromResults` to skip already-extracted digests across all type-hint parsing loops.
2. **Storage-level dedup** — Before appending a new layer in `AttestationStorer.Store` and `SimpleStorer.Store`, check if a layer with the same content digest already exists. This handles the case where independent signable types (`TaskRunArtifact` and `OCIArtifact`) both store to the same `.att`/`.sig` tag.

Fixes #1596

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Fix: prevent duplicate .sig layers when a Task image is reported via multiple result type-hints
```

### DIFF
--- a/pkg/artifacts/signable.go
+++ b/pkg/artifacts/signable.go
@@ -164,6 +164,7 @@ func (oa *OCIArtifact) ExtractObjects(ctx context.Context, obj objects.TektonObj
 func ExtractOCIImagesFromResults(ctx context.Context, results []objects.Result) []interface{} {
 	logger := logging.FromContext(ctx)
 	objs := []interface{}{}
+	seen := map[string]bool{}
 
 	extractor := structuredSignableExtractor{
 		uriSuffix:    OCIImageURLResultName,
@@ -176,6 +177,11 @@ func ExtractOCIImagesFromResults(ctx context.Context, results []objects.Result) 
 			logger.Errorf("error getting digest: %v", err)
 			continue
 		}
+		if seen[dgst.Name()] {
+			logger.Debugf("skipping duplicate image %s", dgst.Name())
+			continue
+		}
+		seen[dgst.Name()] = true
 		objs = append(objs, dgst)
 	}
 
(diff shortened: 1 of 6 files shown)


infernus01 (approved):
Just a small NIT.

pkg/artifacts/signable.go
```diff
@@ -176,6 +177,11 @@ func ExtractOCIImagesFromResults(ctx context.Context, results []objects.Result)
 			logger.Errorf("error getting digest: %v", err)
 			continue
 		}
+		if seen[dgst.DigestStr()] {
```
anithapriyanatarajan: @ab-ghosh  dgst.DigestStr() returns only sha256:xxx,  stripping the repository name entirely. This means two legitimately different images that happen to produce the same content digest (e.g., a mirrored image gcr.io/my/img@sha256:abc and quay.io/my/img@sha256:abc) will be silently deduplicated, dropping one of them from the signing set. so could you consider dgst.Name() that uses the full repo@SHA256:XYZ. Also please include test case addressing this.

anithapriyanatarajan (changes requested):

pkg/artifacts/signable.go
ab-ghosh: Yeah, that's right, thanks for the suggestion. I have updated it to use dgst.Name() and also added a test case to cover this scenario.

anithapriyanatarajan:
@ab-ghosh - with https://github.com/tektoncd/chains/pull/1578 merged now, could we check the feasibility to dedup any image url /digest that could be duplicated with a TR that has results:
  
  results:
  - name: IMAGE_DIGEST
  - name: IMAGE_URL
  - name: build-image-ARTIFACT_OUTPUTS
    type: object
    properties:
      uri: {}
      digest: {}
      isBuildArtifact: {}
 
Please consider this task definition as a sample - https://gist.github.com/anithapriyanatarajan/1650aeb3ebdbcea05d4a9baef649507c#file-gistfile1-txt-L123-L190

ab-ghosh:
Added dedup for the ARTIFACT_OUTPUTS. Also added unit test which covers the exact scenario from your sample task (same image in both IMAGE_URL/IMAGE_DIGEST and build-image-ARTIFACT_OUTPUTS)

anithapriyanatarajan:
/approve

Approving this PR based after feature testing on a kind cluster following steps here - https://gist.github.com/anithapriyanatarajan/81fc8c92ceccf760ce74f40c7b30a2f9

PR #1610, opened 2026-04-03
feat: embed Rekor bundle in OCI attestation layer annotations

ab-ghosh:
# Changes

This PR adds support for embedding the Rekor transparency log bundle (`dev.sigstore.cosign/bundle`) in OCI attestation layer annotations when `transparency.enabled=true`. This enables offline signature verification of attestations signed by Tekton Chains using keyless (Fulcio) signing, bringing Chains' attestation format in line with what `cosign attest` produces.

Fixes #1598

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Embed Rekor bundle in OCI attestation layer annotations to enable offline verification of keyless-signed attestations.
```

### DIFF
--- a/docs/sigstore.md
+++ b/docs/sigstore.md
@@ -35,6 +35,16 @@ Right now, Chains default to storing entries in the public Rekor instance
 kubectl patch configmap chains-config -n tekton-chains -p='{"data":{"transparency.url": "<YOUR URL>"}}'
 ```
 
+### Offline Verification with Rekor Bundle
+
+When transparency is enabled and the OCI storage backend is used, Chains automatically embeds the Rekor bundle (`dev.sigstore.cosign/bundle`) in the OCI attestation layer annotations. This enables offline verification of transparency log entries without querying the Rekor server.
+
+This applies to **attestations only** (in-toto / DSSE-wrapped SLSA provenance stored via `cosign verify-attestation`). OCI image signatures using the simplesigning format do not currently embed the bundle, so `cosign verify` will not find it.
+
+No additional configuration is required — the bundle is embedded automatically when `transparency.enabled` is set to `true` and attestations are stored in an OCI registry.
+
+> **Note:** If the Rekor upload fails (e.g. due to a transient network error), attestation storage still proceeds but the bundle annotation will be absent. Consumers that rely on the bundle for offline verification will silently receive an attestation without it.
+
 ## Keyless Signing Mode
 
 Chains also supports a keyless signing mode with
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	intoto "github.com/in-toto/attestation/go/v1"
+	cbundle "github.com/sigstore/cosign/v2/pkg/cosign/bundle"
 	"github.com/tektoncd/chains/pkg/artifacts"
 	"github.com/tektoncd/chains/pkg/chains/annotations"
 	"github.com/tektoncd/chains/pkg/chains/formats"
@@ -142,6 +143,18 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
 		// Extract all the "things" to be signed.
 		// We might have a few of each type (several binaries, or images)
 		objects := signableType.ExtractObjects(ctx, tektonObj)
+
+		// Create the Rekor client once before iterating over objects since it is
+		// stateless and config-driven.
+		var tlogClient rekorClient
+		if shouldUploadTlog(cfg, tektonObj) {
+			var err error
+			tlogClient, err = getRekor(cfg.Transparency.URL)
+			if err != nil {
+				return err
+			}
+		}
+
 		// Go through each object one at a time.
 		for _, obj := range objects {
 
@@ -186,6 +199,29 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
 			}
 			measureMetrics(ctx, metrics.SignedMessagesCount, o.Recorder)
 
+			// Upload to Rekor before storage so the bundle is available for OCI attestation annotations.
+			// On upload failure, storage proceeds but the bundle annotation will be absent —
+			// consumers that rely on the bundle for offline verification will get an attestation without it.
+			var rekorBundle *cbundle.RekorBundle
+			if tlogClient != nil {
+				entry, err := tlogClient.UploadTlog(ctx, signer, signature, rawPayload, signer.Cert(), string(payloadFormat))
+				if err != nil {
+					logger.Warnf("error uploading entry to tlog: %v", err)
+					o.recordError(ctx, signableType, metrics.TlogError)
+					merr = multierror.Append(merr, err)
+				} else {
+					logger.Infof("Uploaded entry to %s with index %d", cfg.Transparency.URL, *entry.LogIndex)
+					extraAnnotations[annotations.ChainsTransparencyAnnotation] = fmt.Sprintf("%s/api/v1/log/entries?logIndex=%d", cfg.Transparency.URL, *entry.LogIndex)
+					rekorBundle = cbundle.EntryToBundle(entry)
+					if rekorBundle != nil {
+						logger.Infof("Resolved Rekor bundle for offline verification (logIndex: %d)", rekorBundle.Payload.LogIndex)
+					} else {
+						logger.Warn("Rekor entry missing verification data, skipping bundle for offline verification")
+					}
+					measureMetrics(ctx, metrics.PayloadUploadedCount, o.Recorder)
+				}
+			}
+
 			// Now store those!
 			for _, backend := range sets.List[string](signableType.StorageBackend(cfg)) {
 				b, ok := o.Backends[backend]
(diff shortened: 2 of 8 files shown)


pkg/chains/signing/iface.go
```diff
@@ -41,4 +42,6 @@ type Bundle struct {
 	Cert []byte
 	// Cert is an optional PEM encoded x509 certificate chain, if one was used for signing.
```
infernus01: Not related to this PR but I just got eyes on this.
```suggestion
	// Chain is an optional PEM encoded x509 certificate chain, if one was used for signing.
```

infernus01 (changes requested):
@ab-ghosh Would it be good to write this feature in the docs as well?

ab-ghosh (commented):
@infernus01 Good eyes on the comment typo And yes, documenting this feature makes sense, let me add that.

pkg/chains/signing.go
```diff
@@ -186,6 +187,32 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
 			}
 			measureMetrics(ctx, metrics.SignedMessagesCount, o.Recorder)
 
+			// Upload to Rekor before storage so the bundle is available for OCI attestation annotations.
+			var rekorBundle *cbundle.RekorBundle
+			if shouldUploadTlog(cfg, tektonObj) {
+				rekorClient, err := getRekor(cfg.Transparency.URL)
+				if err != nil {
+					return err
+				}
```
infernus01: This seems to be a hard return. What if getRekor() fails (maybe due to bad rekor URL config or anything) and this might prevent all attestation storage. WDYT?

pkg/chains/storage/oci/legacy.go
infernus01, line 170: One thing i noticed that the RekorBundle gets passed through in uploadAttestation below but not in uploadSignature here. since simplesigning + x509/fulcio + rekor is a valid config, the same offline-verification gap might exist for OCI image signatures too. right?
was this intentionally out of scope for this PR, or worth handling here as well?

pkg/chains/signing.go
ab-ghosh: this hard return is pre-existing, IMO if getRekor fails, it's a config error, should not be treated as a transient failure

pkg/chains/storage/oci/legacy.go
ab-ghosh: Yeah that totally makes sense, but #1598 is more aligned with attestations. We can address this in a separate issue/PR, what say?

infernus01: Alright 👍🏻

infernus01 (commented):
/lgtm

pkg/chains/signing.go
```diff
@@ -186,6 +187,32 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
...
+
+				entry, err := rekorClient.UploadTlog(ctx, signer, signature, rawPayload, signer.Cert(), string(payloadFormat))
+				if err != nil {
+					logger.Warnf("error uploading entry to tlog: %v", err)
```
jkhelil, line 209: PLease document this behaviour, if an error occur, it doesnt block storage but bundle will be unavailable." Consumers that rely on the bundle annotation being present will silently get an attestation without a bundle.

pkg/chains/signing.go
```diff
@@ -186,6 +187,32 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
 			}
 			measureMetrics(ctx, metrics.SignedMessagesCount, o.Recorder)
 
+			// Upload to Rekor before storage so the bundle is available for OCI attestation annotations.
+			var rekorBundle *cbundle.RekorBundle
+			if shouldUploadTlog(cfg, tektonObj) {
+				rekorClient, err := getRekor(cfg.Transparency.URL)
```
jkhelil: getRekor(cfg.Transparency.URL) constructs a brand-new Rekor HTTP client on every iteration of the for _, obj := range objects loop.
The client is stateless and config-driven, so it can be safely lifted to before the objects loop.

pkg/chains/signing_test.go
```diff
@@ -573,13 +719,22 @@ type mockRekor struct {
 func (r *mockRekor) UploadTlog(ctx context.Context, signer signing.Signer, signature, rawPayload []byte, cert, payloadFormat string) (*models.LogEntryAnon, error) {
```
jkhelil, line 775: The code path at signing.go:209 ("Rekor entry missing verification data, skipping bundle") is never exercised. Add a test or a second mock variant where Verification is nil, asserting that RekorBundle is nil in StorageOpts even when tlog upload succeeds.

docs/sigstore.md
```diff
@@ -35,6 +35,12 @@ Right now, Chains default to storing entries in the public Rekor instance
 kubectl patch configmap chains-config -n tekton-chains -p='{"data":{"transparency.url": "<YOUR URL>"}}'
 ```
 
+### Offline Verification with Rekor Bundle
+
+When transparency is enabled and the OCI storage backend is used, Chains automatically embeds the Rekor bundle (`dev.sigstore.cosign/bundle`) in the OCI attestation layer annotations. This enables offline verification of transparency log entries without querying the Rekor server.
+
+No additional configuration is required — the bundle is embedded automatically when `transparency.enabled` is set to `true` and attestations are stored in an OCI registry.
```
jkhelil, line 44: Add a sentence that this applies to attestations only (in-toto / DSSE-wrapped SLSA); simplesigning (OCI image signature via simplesigning format) does not currently embed the bundle. Without this, users may set up Fulcio+Rekor+OCI storage, observe that cosign verify (not verify-attestation) doesn't find the bundle, and be confused.

ab-ghosh:
Thank you for the review, @jkhelil! I've addressed all your suggestions:
  - Documented the behavior when Rekor upload fails
  - Lifted getRekor() out of the objects loop
  - Added TestSigner_NilVerificationSkipsBundle for the nil Verification path
  - Clarified in docs that the bundle applies to attestations only
 
  Please take a look when you get a chance.

jkhelil:
@anithapriyanatarajan can you have a look please

docs/sigstore.md
anithapriyanatarajan: @ab-ghosh - is this done?

pkg/chains/signing/iface.go
```diff
@@ -39,6 +40,8 @@ type Bundle struct {
 	Signature []byte
 	// Cert is an optional PEM encoded x509 certificate, if one was used for signing.
 	Cert []byte
-	// Cert is an optional PEM encoded x509 certificate chain, if one was used for signing.
+	// Chain is an optional PEM encoded x509 certificate chain, if one was used for signing.
```
anithapriyanatarajan, line 43: ❤️

anithapriyanatarajan:
@ab-ghosh Thank you for this PR. Verified that with PR rekor bundle for reference
`{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "size": 233,
    "digest": "sha256:f772b2a04aa836eba6af79e20228f0364fb0fa0877038884c56d5524f3880846"
  },
  "layers": [
    {
      "mediaType": "application/vnd.dsse.envelope.v1+json",
      "size": 2654,
      "digest": "sha256:cab0e1e0dc66c9084883ced14fb5b5d4d62e5f922c3b9e7fcedbefde38cc36aa",
      "annotations": {
        "dev.cosignproject.cosign/signature": "",
        **"dev.sigstore.cosign/bundle": "{\"SignedEntryTimestamp\":\"MEYCIQCXuOiPPOVB6ZlVHmQc9lc2JPKSwOOjbgW+KOEm90P5FAIhALERFX+qw2ocJBR1z5ucs+9711QUCIiStdXTZD9CsM4z\",\"Payload\":{\"body\":\"eyJhcGlWZXJzaW9uIjoiMC4wLjEiLCJraW5kIjoiaW50b3RvIiwic3BlYyI6eyJjb250ZW50Ijp7Imhhc2giOnsiYWxnb3JpdGhtIjoic2hhMjU2IiwidmFsdWUiOiJjYWIwZTFlMGRjNjZjOTA4NDg4M2NlZDE0ZmI1YjVkNGQ2MmU1ZjkyMmMzYjllN2ZjZWRiZWZkZTM4Y2MzNmFhIn0sInBheWxvYWRIYXNoIjp7ImFsZ29yaXRobSI6InNoYTI1NiIsInZhbHVlIjoiY2M3MTk5MTE3NmZjZDAyNjlkMmI3OTNlNDE3MDYwNTcyNWIzMDM3MjE5ZDI1ODEzNDRjM2RlOWMzMzEzN2YyMSJ9fSwicHVibGljS2V5IjoiTFMwdExTMUNSVWRKVGlCUVZVSk1TVU1nUzBWWkxTMHRMUzBLVFVacmQwVjNXVWhMYjFwSmVtb3dRMEZSV1VsTGIxcEplbW93UkVGUlkwUlJaMEZGV1ZGd1pqUlFSRE5LWmxOTGIzRkNPV2xEUTNWSVNtcGxUMnRMVVFvMmJXWlpjelV5VEd4eGR6ZFRjMG92Y2t0UmJWTXZZVkJSV2tOVk9VVnFZMlpQUzA1dFZETXhTalkxU1VjemJGbEVhMFZZTjA5c05VVjNQVDBLTFMwdExTMUZUa1FnVUZWQ1RFbERJRXRGV1MwdExTMHRDZz09In19\",\"integratedTime\":1783071679,\"logIndex\":2058342836,\"logID\":\"c0d23d6ad406973f9559f3ba2d1ca01f84147d8ffc5b8445c224f98b9591801d\"}}",**
        "dev.sigstore.cosign/certificate": "",
        "dev.sigstore.cosign/chain": ""
      }
    }
  ]
}
`
Happy to lgtm. If possible please post a blog with cosign steps to verify. Thank you.

PR #1636, opened 2026-04-16
Update example performance doc

enarha:
Original values don't make sense from a performance point of view as they are lower than the default values. Also high number of threads combined with bottleneck on QPS has the opposite effect as objects are re-queued and wait in the workqueue with exponentially increasing timeout.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -19,9 +19,9 @@ spec:
     template:
         spec:
             containers:
-                - image: gcr.io/tekton-releases/github.com/tektoncd/chains/cmd/controller:v0.20.0
+                - image: ghcr.io/tektoncd/chains/controller-92006fd957c0afd31de6a40b3e33b39f:v0.26.2
                   args:
                     - --threads-per-controller=32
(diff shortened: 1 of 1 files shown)


docs/performance.md
```diff
@@ -22,6 +22,6 @@ spec:
                 - image: gcr.io/tekton-releases/github.com/tektoncd/chains/cmd/controller:v0.20.0
```
anithapriyanatarajan, line 22: @enarha Thankyou for this PR. while reviewing I noticed the examples of chains are still pointing to gcr.io which is obsolete. Could you consider updating all places where gcr.io/tekton-releases/..... with the following in docs samples:

ghcr.io/tektoncd/chains/controller-92006fd957c0afd31de6a40b3e33b39f:v0.26.2

anithapriyanatarajan (changes requested):

PR #1666, opened 2026-05-12
fix: aggregate per-artifact signing failures so signed annotation is not set

### LINKED ISSUE #1663: Bug: MarkSigned called despite per-artifact signing failures
## Summary

Per-artifact failures in `ObjectSigner.Sign()` (`pkg/chains/signing.go`) are logged and `continue`d without contributing to the aggregate error (`merr`). If all per-artifact operations fail silently, `merr` remains nil, and `MarkSigned` is called — permanently marking the resource as `chains.tekton.dev/signed=true` despite no artifacts being actually signed.

## Bug

Three failure paths in the signing loop use `continue` without appending to `merr`:

1. **Payload creation failure** (L148-152) — `CreatePayload` error → logged → `continue`
2. **Payload marshal failure** — `getRawPayload` error → logged → `continue`  
3. **Signing failure** — `SignMessage` error → logged → `continue`

Only **storage failures** correctly aggregate into `merr`.

## Impact

- `chains.tekton.dev/signed=true` is set when no artifacts were signed
- `signed=true` is a terminal marker — suppresses retries on subsequent reconciliation
- Errors are only logged, not surfaced to the resource

## Reproduction

Create a TaskRun with a malformed `IMAGE_DIGEST` result. Chains logs the error at `signable.go:176` ("unsupported digest algorithm") but sets `signed=true` anyway.

## Fix

Aggregate per-artifact errors into `merr`:

```go
// Example for payload creation path:
payload, err := payloader.CreatePayload(ctx, obj)
if err != nil {
    logger.Error(err)
    o.recordError(ctx, signableType, metrics.PayloadCreationError)
    merr = multierror.Append(merr, fmt.Errorf("payload creation for %s: %w", signableType.Type(), err))
    continue
}
```

Same pattern for marshal and signing paths. This ensures `MarkSigned` is only called when all artifacts are successfully processed.

/kind bug

SAY-5:
# Changes

Per-artifact `CreatePayload`, payload marshalling and `SignMessage` failures in `ObjectSigner.Sign()` were only logged and `continue`d without contributing to `merr`. When every per-artifact operation failed this way, `merr` stayed nil, so `MarkSigned` set `chains.tekton.dev/signed=true` even though nothing was signed, permanently suppressing retries. These three paths now append to `merr` exactly like the storage-failure path a few lines below already does.

No new unit test is included: the change is the same `multierror.Append(merr, ...)` pattern already used for storage failures in the same loop, and exercising the payload/marshal/sign failure branches in isolation would require stubbing out the real formatter and x509 signer used by `TestSigner_Sign`.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Fix Chains marking a TaskRun/PipelineRun as signed when all per-artifact payload creation, marshalling or signing operations failed.
```

Closes #1663

### DIFF
--- a/pkg/chains/signing.go
+++ b/pkg/chains/signing.go
@@ -149,6 +149,7 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
 			if err != nil {
 				logger.Error(err)
 				o.recordError(ctx, signableType, metrics.PayloadCreationError)
+				merr = multierror.Append(merr, fmt.Errorf("creating payload for %s: %w", signableType.Type(), err))
 				continue
 			}
 			logger.Infof("Created payload of type %s for %s %s/%s", string(payloadFormat), tektonObj.GetGVK(), tektonObj.GetNamespace(), tektonObj.GetName())
@@ -175,6 +176,7 @@ func (o *ObjectSigner) Sign(ctx context.Context, tektonObj objects.TektonObject)
 			if err != nil {
 				logger.Warnf("Unable to marshal payload for %s: %v", signerType, err)
 				o.recordError(ctx, signableType, metrics.MarshalPayloadError)
+				merr = multierror.Append(merr, fmt.Errorf("marshalling payload for %s: %w", signableType.Type(), err))
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


infernus01 (commented):
Thanks @SAY-5 for the PR. You'd need to sign CLA please check the link in the CI checks below.

waveywaves (commented):
Thanks for the fix. The code change makes sense: payload creation, marshal, and signing errors now contribute to `merr`, so `Sign` returns an error and avoids marking the object as signed after a partial failure.

Before this is ready to merge, could you please:

1. Sign the CLA; the current EasyCLA check is blocking the PR.
2. Add regression coverage in `pkg/chains/signing_test.go` for one of these failure paths, verifying that `Sign` returns an error and the object is not marked signed.

I ran the existing targeted test locally with this branch:

```sh
go test ./pkg/chains -run TestSigner_Sign -count=1
```

and it passes.

PR #1691, opened 2026-06-02
feat: add OCI 1.1 Referrers API support with configurable encoding format

anithapriyanatarajan:
**NOTE TO REVIEWERS:** Originally this PR intended to allow a user-configurable approach for both distribution and serialization. During later [review](https://github.com/tektoncd/chains/pull/1691#discussion_r3397879385) and iterative testing it became clear this adds needless complexity for users, and we additionally discovered via [sigstore/cosign#4841](https://github.com/sigstore/cosign/pull/4841) that the `sigstore-bundle` + DSSE combination is not supported in cosign. The PR therefore distills to two simple scenarios behind a single config knob, summarized below.

# Changes

Chains currently stores all OCI signatures and attestations using a tag-based scheme (`sha256-<digest>.sig` / `.att`). This creates extra tags in the registry for every signed artifact. This PR adds support for the OCI 1.1 Referrers API as an alternative storage mechanism.

### New configuration key

A single new key controls both the serialization format and the distribution mechanism. They are intentionally coupled — not separately configurable:

| Key | Values | Default |
|-----|--------|---------|
| `storage.oci.encoding-format` | `dsse`, `sigstore-bundle` | `dsse` |

Possible Scenarios:

| `encoding-format` | Encoding | Behaviour |
|-------------------|----------|-----------|
| `dsse` | DSSE envelope | Tag-based storage (`.sig` / `.att` tags) — unchanged default |
| `sigstore-bundle` | Sigstore protobuf bundle | OCI 1.1 Referrers API — no `.sig` / `.att` tags |

`dsse` + protobuf-bundle and `sigstore-bundle` + DSSE are intentionally not offered: cosign's verification for referrer-stored attestations expects the protobuf bundle, and `sigstore-bundle` exists precisely to move away from the legacy tag layout, which has downsides:

- The registry fills up with extra tags that aren't real images.
- These tags are easy to confuse with actual image tags.
- No OCI standard describes this layout, so every tool has to special-case it.

This keeps the surface to one knob and avoids a confusing matrix of combinations, some of which no tool can verify.

### Backward compatibility

All existing deployments with no OCI encoding format config set continue to behave identically (`dsse` + tag-based). No action required.

## Implementation

- **`config.go`** — `OCIStorageConfig` gains an `EncodingFormat` field; new constants (`OCIEncodingFormatDSSE`, `OCIEncodingFormatSigstoreBundle`) and a `validateOCIEncodingFormat` validator. Defaults to `dsse` when unset.
- **`options.go`** — New `WithEncodingFormat()` option function; houses the `referrersRepoOverrideIgnored` helper.
- **`attestation.go` / `simple.go`** — Dispatch keys off `encodingFormat`; `sigstore-bundle` writes attestations as a protobuf bundle via `ociremote.WriteAttestationNewBundleFormat`, while signatures use cosign's native signature referrer.
- **`legacy.go`** — Backend passes `WithEncodingFormat` when constructing storers.
- **`signing.go` / `signing/iface.go`** — `signing.Bundle` and `StorageOpts` gain a `PublicKey crypto.PublicKey` field; `PublicKey()` extraction after signing is non-fatal (logs a warning, storage continues) so KMS signers don't break the default path. Also wires up Rekor bundle population before storage so the bundle annotation is available for offline verification.

## Tests

- **`config_test.go`** — Tests the default, valid/invalid values for `encoding-format`, and the `validateOCIEncodingFormat` validator.
- **`oci_test.go`** — Tests `WithEncodingFormat`, empty defaults, the Backend config, and that `storage.oci.repository` override is ignored in `sigstore-bundle` mode.
- **`store_test.go`** — Updated `defaultStorage` to reflect the struct field rename.
- **`attestation_test.go` / `simple_test.go`** — Tests for the `sigstore-bundle` store path including dedup.
- **`referrers_api_e2e_test.go`** (new) — E2E tests for `TestOCIStorageSigstoreBundle_TaskRun` and `TestOCIStorageSigstoreBundle_PipelineRun`.

## Docs

- **`oci-artifact-distribution-format-referrers-schema.md`** (new) — Tutorial-style page explaining the Referrers schema, how cosign and Chains enable it, why referrers + protobuf bundle (not DSSE), registry compatibility, and verification examples.
- **`config.md`** — New `storage.oci.encoding-format` row in the storage configuration table.

# Submitter Checklist

- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)

# Release Notes

```release-note
feat: OCI storage now supports the OCI 1.1 Referrers API via a new config key:

Set `storage.oci.encoding-format: sigstore-bundle` in the `chains-config` ConfigMap
to store signatures and attestations as Sigstore protobuf bundles (v0.3) via the
OCI 1.1 Referrers API instead of `.sig`/`.att` tags.

- Default value `dsse` preserves existing tag-based behavior (backward compatible).
- `sigstore-bundle` mode: both signatures and attestations are pushed as OCI 1.1 referrers with `artifactType: application/vnd.dev.sigstore.bundle.v0.3+json`, discoverable via `oras discover` or `cosign download`.
- Transparency log entries (Rekor) are embedded in the bundle when
  `transparency.enabled: "true"`.
- Verification: `cosign verify` and `cosign verify-attestation` work with both modes.
- Re-signing in `sigstore-bundle` mode accumulates referrers (one per signing invocation) because ECDSA is non-deterministic; the dedup only coalesces exact crash-recovery retries.
- Invalid values for `storage.oci.encoding-format` are rejected by the controller at config-load time; the last valid config continues to be used.
```

/kind feature

Co-Authored-by: Copilot (Claude Sonnet 4.6)


anithapriyanatarajan:
@tektoncd/chains-collaborators @arewm - Please review and share your comments. Thank you

anithapriyanatarajan:
Detailed steps to test and observations are captured here for reference - https://gist.github.com/anithapriyanatarajan/b112638e7b3d78ee6638c6fec4f51bc8

Observed Gaps, which would be analyzed further with appropriate follow ups:

1. cosign verify discoverability (used cosign 3.0.3):  the signature path uses the low-level WriteReferrer. Switching to cosign-native WriteSignaturesExperimentalOCI would let cosign verify reverse-discover signatures directly. Scoped as a follow-up to keep the storage PR focused. Update NOTE: Fixed in [commit] (https://github.com/tektoncd/chains/pull/1691/commits/4d43d459ecc6590a1015febd7069040c87ba1d64) - this should no more be a problem

2. quay.io read path:  write succeeds but referrers aren't returned on read. Registry-side bug, not a Chains issue. The functionality was completely verifiable if we target ghcr.io. we may need to check the status of other registries and add to the release notes.

Registry readiness for OCI 1.1 is much critical to realize the benefit of this PR for users.

pkg/chains/storage/oci/legacy.go
```diff
@@ -208,6 +225,7 @@ func (b *Backend) uploadAttestation(ctx context.Context, attestation *intoto.Sta
 				Signature: []byte(signature),
 				Cert:      []byte(storageOpts.Cert),
 				Chain:     []byte(storageOpts.Chain),
+				PublicKey: storageOpts.PublicKey,
```
ab-ghosh: PublicKey is added here for the protobuf-bundle path, but `Content` on the line above is still nil.
`uploadSignature`  correctly passes Content: rawPayload, but uploadAttestation leaves it nil. This is fine for `storeLegacy` and `storeWithReferrersAPI` (DSSE mode) which only use req.Bundle.Signature.
But `storeWithProtobufBundle` calls `buildDSSEEnvelope(req.Bundle.Content, ...)` and `cbundle.MakeNewBundle(..., req.Bundle.Content, ...)` both receive nil.

Verified on ghcr.io: the protobuf-bundle attestation's dsseEnvelope has no "payload" key, while the DSSE-mode attestation correctly has ["payload", "payloadType", "signatures"].

pkg/chains/storage/oci/attestation.go
```diff
@@ -98,14 +122,110 @@ func (s *AttestationStorer) Store(ctx context.Context, req *api.StoreRequest[nam
...
+	}
+	att, err := static.NewAttestation(req.Bundle.Signature, attOpts...)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating attestation")
+	}
+	newImage, err := mutate.AttachAttestationToEntity(se, att)
+	if err != nil {
+		return nil, errors.Wrap(err, "attaching attestation to entity")
+	}
+	if err := ociremote.WriteAttestationsReferrer(req.Artifact, newImage, ociremote.WithRemoteOptions(s.remoteOpts...)); err != nil {
```
ab-ghosh: Just want to confirm, here we are using req.Artifact, which drops the override of repo and ignores the storage.oci.repository. Should we use the overridden repo instead?

anithapriyanatarajan:
More details on why we are doing this change -  https://github.com/sigstore/docs/pull/411 & https://github.com/kubernetes-sigs/tejolote/issues/639

pkg/config/config.go
```diff
@@ -237,6 +256,67 @@ func (artifact *Artifact) Enabled() bool {
...
+	}
+	// Only migrate when no explicit new-field values are present.
+	_, distSet := data[ociDistributionMethodKey]
+	_, serSet := data[ociSerializationFormatKey]
+	if distSet || serSet {
+		return
+	}
+	if referrersVal == "true" {
+		cfg.Storage.OCI.DistributionMethod = OCIDistributionReferrersAPI
+		cfg.Storage.OCI.SerializationFormat = OCISerializationProtobuf
```
jkhelil: shall this be 
if referrersVal == "true" {
    cfg.Storage.OCI.DistributionMethod = OCIDistributionReferrersAPI
    cfg.Storage.OCI.SerializationFormat = OCISerializationDSSE  
}
to avoid use the experimental protobuf bundle

pkg/chains/storage/oci/attestation.go
```diff
@@ -98,14 +122,110 @@ func (s *AttestationStorer) Store(ctx context.Context, req *api.StoreRequest[nam
...
 	if err := ociremote.WriteAttestations(repo, newImage, ociremote.WithRemoteOptions(s.remoteOpts...)); err != nil {
+		return nil, errors.Wrap(err, "writing attestations")
+	}
+	logger.Infof("Successfully uploaded attestation using legacy format for %s", req.Artifact.String())
+	return &api.StoreResponse{}, nil
+}
+
+// storeWithReferrersAPI uploads attestations using the OCI 1.1 Referrers API
+// (DSSE serialisation – same payload as legacy, fewer registry tags).
+func (s *AttestationStorer) storeWithReferrersAPI(ctx context.Context, req *api.StoreRequest[name.Digest, *intoto.Statement], se oci.SignedEntity) (*api.StoreResponse, error) {
```
jkhelil: storeLegacy in both storers has explicit dedup logic (checks existing attestation/signature digests before writing). The new storeWithReferrersAPI paths have no such check. In referrers mode, repeated reconciler invocations for the same artifact can push duplicate referrers manifests into the registry.

jkhelil:
@anithapriyanatarajan 
Does the PR handles automatically fall back to the legacy tag schema when a registry does not support the OCI 1.1 Referrers API, so that storage.oci.format=referrers can be safely configured without knowing the registry's OCI 1.1 support status.

jkhelil:
@anithapriyanatarajan can you add e2e tests ?

docs/config.md
```diff
@@ -69,6 +69,9 @@ Supported keys include:
...
+| `storage.oci.serialization-format`              | Controls how the cryptographic payload is encoded. `protobuf-bundle` requires `storage.oci.distribution-method: referrers-api`. See [OCI Storage Formats](oci-storage-formats.md) for details.                                                                    | `dsse`, `protobuf-bundle`                                                                                                                                                                                                                                                                                                                                                                                                                                           | `dsse`   |
```
arewm: I ended up implementing this type of separation for my PR but I feel like @lcarva and I had a conversation somewhere (somewhen) about proposing the alternative.

Cosign support has moved to protobuf + oci as the default. The previous default was tag-based + dsse. Since we should be trying to encourage people to get to the current best practice, I think that we should only have those two options. Either users get what is implemented today or they get the protobuf + bundle format.

If we "support" more variants, we will likely just be resulting in confusion.

anithapriyanatarajan: Thank you for pointing this out. Realized this while assessing the referrer+DSSE combination and located https://github.com/sigstore/cosign/pull/4841 which clearly states to not support this in cosign. verify-attestation will not work for this combo. Reworking on these lines. Will mark this PR as draft to revisit the implementation. Will request re-review once done

arewm: Sounds good. From my previous experience, I would expect that it should simplify things. I imagine, for example, that you might go from three configuration knobs to one. 😄

anithapriyanatarajan:
> @anithapriyanatarajan Does the PR handles automatically fall back to the legacy tag schema when a registry does not support the OCI 1.1 Referrers API, so that storage.oci.format=referrers can be safely configured without knowing the registry's OCI 1.1 support status.

@jkhelil - Yes, storage.oci.format=referrers-api can be safely configured without knowing the registry's OCI 1.1 support status. The fallback to the Referrers Tag Schema is handled automatically by go-containerregistry (the underlying library used by cosign and vendored in Chains). If the registry natively supports the Referrers API (e.g., Quay), it is used directly. If not, go-containerregistry automatically falls back to storing referrers as an OCI Image Index tag, per the OCI distribution spec mandate ([spec lines 777–806](https://github.com/opencontainers/distribution-spec/blob/main/spec.md?plain=1#L777-L806)). Since this fallback is transparent at the library level, Chains does not need to implement its own fallback logic, and users do not need to manually revert to the legacy storage format based on registry capability. I have verified this behavior with quay and ghcr.  For quay it takes the referrers API path while for ghcr it takes the referrers tag path. Will share evidences after fixing a few other findings. Thank you

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+
+## The problem with the old layout
+
+cosign, which Chains uses under the hood, has always stored a signature by
+pushing an extra tag next to the image. For an image at digest `sha256:abc...`,
+it creates tags like `sha256-abc....sig` and `sha256-abc....att`.
+
+This works on every registry, but it isn't ideal:
+
+- The registry fills up with extra tags that aren't real images.
```
arewm: nit: This is only an issue if you expect everything to be a pullable container image. If you expect that the registry is an OCI registry, it isn't necessarily an issue. But even with that, it muddies the purpose of a content that is stored in a registry and is referenceable by tags. The supporting metadata can get confused with the top level artifacts themselves. Maybe that is what you were saying below?

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+question: "what artifacts refer to this image?"
+
+This is the **Referrers schema**. The payoff is no extra tags, a clean registry,
+and a standard that registries and policy tools already understand.
+
+## How cosign enables it
+
+cosign (and the `go-containerregistry` library it uses) already speaks the
+Referrers schema. If the registry supports the Referrers API natively, cosign
+uses it. If it doesn't, cosign falls back to the spec's **referrers tag schema**:
```
arewm: This isn't necessarily the case. I don't think cosign would have used the referrer's API in v2 unless an additional option was passed.

But maybe you have tested things more than me, so your content might be more authoritative.

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+it keeps a single `sha256-<digest>` index tag and uses it to track referrers.
+
+Either way it is still "referrers mode" — the content is the same, no `.sig` or
+`.att` tags are created, and `cosign verify` and `oras discover` both work. This
```
arewm: Another issue is that the .att points to a manifest referencing all of the attestations. This does not have a stable digest. If some other process pushes additional attestations, I think they will all be added into this one manifest.

By moving attestations to use the referrer's API, each attestation will have its own image manifest and will therefore have a stable digest.

This was one of the primary motivators why I wanted to change.

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+> [!NOTE]
+> This flag only takes effect when the OCI storage backend is in use — that is,
+> when `artifacts.oci.storage`, `artifacts.taskrun.storage`, or
+> `artifacts.pipelinerun.storage` includes `oci`. If you store signatures and
+> attestations somewhere else (for example Tekton results, a docstore, or
+> Grafeas), `storage.oci.distribution-method` has no effect.
+
+> [!TIP]
+> The cosign bundled with Chains is new enough for everything described here, so
+> you don't need to install anything or pass extra flags.
```
arewm: This seems out of context?

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Why referrers + protobuf, and not referrers + DSSE
+
+There is one subtlety worth understanding: the *encoding* of the payload.
+
+- In **legacy** mode, the attestation is a **DSSE** envelope. This is what cosign
+  has always written and what existing tooling verifies.
+- In **referrers** mode, Chains writes the attestation as a **Sigstore protobuf
+  bundle**.
```
arewm: Maybe you want to differentiate by what is the default configuration for cosign versions?

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+You might expect a third option — referrers with DSSE — but it's a dead end in
+practice. cosign's verification for referrer-stored attestations expects the
+protobuf bundle, not a DSSE envelope. A DSSE-over-referrers attestation can be
+written, but `cosign verify-attestation` won't reliably verify it. The industry
+is also moving toward the protobuf bundle as the standard format.
+
+So Chains keeps it simple: legacy means DSSE, referrers means protobuf. The two
+are tied together on purpose — it avoids a confusing matrix of combinations,
+some of which no tool can actually verify. If a genuine need for a separate
+encoding option ever appears, it can be added later without changing this flag.
```
arewm: Can we just simplify this to state that we maintain support with cosign which doesn't support referrer's without protobuf?

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+
+Image **signatures** are not affected by this. In referrers mode they use
+cosign's native signature manifest — the exact thing `cosign verify` looks for —
+so signature verification works with no extra flags.
```
arewm: I thought the referring signatures were stored as protobuf as well.

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+1. **`storage.oci.repository` is ignored.** A referrer has to live next to the
+   image it points at, so the override that redirects storage elsewhere doesn't
+   apply. Chains logs a warning and stores the referrer next to the image. The
+   override still works in `legacy` mode.
+
+2. **cosign's `--experimental-oci11` discovery may not find the attestation.**
+   Chains stores it as a protobuf bundle; that older discovery path filters on a
+   different type. The attestation is still there — `oras discover` shows it and
+   policy engines can use it — and `cosign verify` of the signature is
+   unaffected.
```
arewm: Is there cosign documentation that we can point to? I don't think that this option exists for all versions of cosign (i.e. protobuf is the default for newer cosign versions, so I wouldn't expect it to be there).

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+
+```shell
+oras discover <image>@sha256:<digest>
+```
+
+## Things to keep in mind in referrers mode
+
+These are interoperability notes, not bugs in Chains.
+
+1. **`storage.oci.repository` is ignored.** A referrer has to live next to the
```
arewm: How is this used today?

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+   reports success but you can't read the referrer back, it isn't fully OCI 1.1
+   compliant. Switch that registry to `legacy`.
+
+4. **`oras discover` may show a different `artifactType` per registry.** This is
+   display only; the stored content and verification don't change.
+
+5. **Concurrent writes can race on fallback registries.** Without the native API,
+   the index tag is updated with read-append-write, so simultaneous writes to the
+   same image can drop an entry. A registry with the native Referrers API avoids
+   this.
```
arewm: This refers just to the legacy mode?

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Registry compatibility
+
+cosign works with a wide range of registries, including AWS ECR, GCP Artifact
+Registry, Docker Hub, Azure Container Registry, JFrog Artifactory, GitLab and
+GitHub Container Registries, Harbor, and Quay. See the
```
arewm: alphabetize?

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,179 @@
...
+[cosign registry support page](https://docs.sigstore.dev/cosign/system_config/registry_support/)
+for the current list.
+
+| Registry | `legacy` | `referrers-api` |
+|---|---|---|
+| GCR, ECR, Artifact Registry, quay.io | ✓ | ✓ (native Referrers API) |
+| GHCR (`ghcr.io`) | ✓ | ✓ (native API not exposed; cosign uses the tag-schema fallback) |
+| Any other OCI registry | ✓ | ✓ (native API if available, otherwise tag-schema fallback) |
```
arewm: This table is confusing. I was expecting that it was trying to say which ones had support for the actual referrer's API but you are just saying waht is compatible? 

I feel like this shouldn't be included so that you are not expected to keep it up to daet.

arewm (commented):
Just reviewing the text, not any of the actual implementation. I do not know if comments on the text should result in changes in the code.

pkg/chains/storage/oci/legacy.go
anithapriyanatarajan: Thank you for the detailed investigation. This is fixed now and verified with quay and ghcr. Please review and mark it as resolved if you are convinced.

pkg/config/config.go
anithapriyanatarajan: Thank you @jkhelil.  This field no longer exists in the current version of the PR. We removed the separate serialization-format config option entirely. There is now just one key (storage.oci.distribution-method), and the serialization is picked automatically:
For **legacy** serialization would be  **DSSE** (unchanged default)
and for **referrers-api** serialization would be  **protobuf bundle**
We intentionally dropped referrers-api + DSSE because cosign explicitly does not support that combination (sigstore/cosign#4841). Cosign verify-attestation would fail for anyone using it.

On the "experimental" concern: protobuf bundle v0.3 is what cosign uses by default since v2. Got to understand that it is not experimental anymore.

anithapriyanatarajan:
> @anithapriyanatarajan can you add e2e tests ?

Done

pkg/chains/storage/oci/attestation.go
anithapriyanatarajan: Addressed with https://github.com/anithapriyanatarajan/chains/blob/oci-referrer-api-enablement/pkg/chains/storage/oci/attestation.go#L182-L215 for attestation and https://github.com/anithapriyanatarajan/chains/blob/oci-referrer-api-enablement/pkg/chains/storage/oci/simple.go#L161-L190 for signature

docs/config.md
anithapriyanatarajan: Marking this as resolved, since the latest PR is simplified with just one config option. `storage.oci.distribution-method`

docs/oci-artifact-distribution-format-referrers-schema.md
anithapriyanatarajan: Reworded to capture this intent.

docs/oci-artifact-distribution-format-referrers-schema.md
anithapriyanatarajan: https://github.com/anithapriyanatarajan/chains/blob/oci-referrer-api-enablement/docs/oci-artifact-distribution-format-referrers-schema.md?plain=1#L14-L30 - this section captures this aspect now.

docs/oci-artifact-distribution-format-referrers-schema.md
anithapriyanatarajan: Cleaned up.

docs/oci-artifact-distribution-format-referrers-schema.md
anithapriyanatarajan: Tested with latest version of cosign and verify of attestation was possible without this experimental flag.

docs/oci-artifact-distribution-format-referrers-schema.md
anithapriyanatarajan: Done

docs/oci-artifact-distribution-format-referrers-schema.md
anithapriyanatarajan: Have removed this now.

docs/config.md
```diff
@@ -69,6 +69,7 @@ Supported keys include:
...
+| `storage.oci.distribution-method`               | Controls how OCI signatures and attestations are attached to images in the registry, and implicitly the payload encoding: `legacy` uses tag-based storage with DSSE payloads, `referrers-api` uses the OCI 1.1 Referrers API with Sigstore protobuf-bundle attestations. See [OCI Artifact Distribution (Referrers)](oci-artifact-distribution-format-referrers-schema.md) for details.                                                                                                                | `legacy`, `referrers-api`                                                                                                                                                                                                                                                                                                                                                                                                                                           | `legacy` |
```
arewm: I suspect that users will care more about the format whether that is legacy or sigstore bundle formats. It is just that sigstore bundles also happen to require the referrer's api. I don't know how much reframing of the document that would require.

anithapriyanatarajan: `referrers-api` is specific to the OCI storage backend. All other backends (GCS, Firestore, Grafeas, docdb, Archivista, Tekton annotations) continue to use DSSE as-is. This PR does not change them.

As long as cosign `verify-attestation` and cosign `verify-blob-attestation` continue to support DSSE-encoded attestations, the current split is clean. OCI users can opt into the new bundle format via referrers-api, while everything else stays on DSSE. I've raised a clarification with the cosign/sigstore community on  this point: https://sigstore.slack.com/archives/C01PZKDL4DP/p1782872895452699.

However I would like to get views from @tektoncd/chains-collaborators on a broader question: should Chains eventually move all storage backends to the Sigstore protobuf bundle format (which mandates protobuf serialization), or is the current per-backend split the right long-term architecture?

anithapriyanatarajan: @arewm - just to reconfirm, from a user's perspective, choosing `referrers-api` defaults to sigstore bundle. There is no referrer api support DSSE format. So with that clarification, I hope this PR is complete. Rest of the Backends, I prefer to handle in separate PR. Do you have any further concerns with this? Thank you.

arewm: Yes, I agree with that statement but if we expect that users' primary concern for supportability is the format (i.e. bundle vs DSSE) then we should provide that as the toggle. Many registries support the referrer's api v1.1. Again, this is just my assumption. I care about the referrer's API but others might not. It seems reasonable to make the toggle most apparent for the more-likely-to-break change.

anithapriyanatarajan: :smile_cat:  Shall we rename `storage.oci.distribution-method` to something more format-centric, like `storage.oci.bundle-format` with values legacy (DSSE + tag-based) and sigstore-bundle (protobuf bundle + referrers API). The referrers API becomes an implementation consequence of choosing the bundle format, rather than the primary toggle. This also aligns better with how users will reason about cosign v4 compatibility.

Would that naming satisfy your concern? Happy to do the rename in this PR if it gets us to approval.

we could extend a similar configuration for each non-oci backend as we progress with the changes. `storage.gcs.bundle-format`, etc.,

arewm: If you were to name as `bundle-format`, I assume that would just be a `true/false` configuration? If you want the configuration to be clearly legacy vs bundle, a `storage-format` or `encoding-format` might make more sense? Instead of having `legacy/bundle` options, it might also be clearer to have it be `DSSE/bundle` as that is the actual format?

So maybe the suggestion would be `storage.oci.format` with available options of `DSSE` (default for now) and `bundle`?

anithapriyanatarajan:
@jkhelil - I would like to reiterate to see the feasibility of accommodating https://github.com/tektoncd/chains/pull/1691#discussion_r3512951198. Hence marking this as work in progress.

docs/config.md
anithapriyanatarajan: Thank you. One thing to note: `storage.oci.format` would conflict conceptually with the existing `artifacts.oci.format` key (which controls simplesigning vs slsa/v1), so proposing `storage.oci.encoding-format` from your suggestion, with the following values:

**dsse** (default; current behavior, DSSE envelope + tag-based storage)
**sigstore-bundle** (new; protobuf bundle + OCI 1.1 referrers API)

It also naturally extends to storage.gcs.encoding-format etc. in a follow-up.

Hope we can agree upon this name. Once confirmed, Happy to do the rename and rework this PR.

arewm: That works, thanks.

docs/oci-artifact-distribution-format-referrers-schema.md
```diff
@@ -0,0 +1,236 @@
+<!--
+---
+linkTitle: "OCI Artifact Distribution (Referrers)"
+weight: 35
+---
+-->
+
+# OCI Artifact Distribution: the Referrers Schema
```
arewm: Does this document need to be regenerated/reframed based on the configuration option we now have?

anithapriyanatarajan: Done.

anithapriyanatarajan:
## Manual test evidence
 
Tested manually on a `kind` cluster across four configurations of `storage.oci.encoding-format`, using the same [TaskRun](https://gist.github.com/anithapriyanatarajan/9b9e5a0575f570d2b978f1e6b2083bde) for each. Environment: `kind` + registry(quay.io/rh-ee-anataraj), cosign `v3.1.1`, transparency enabled but tlog entries were not made (so `cosign verify*` is run with `--insecure-ignore-tlog=true`; the tlog skip is expected in this env, not a failure).
 
Full write-up with every command and its output (cosign tree/verify/verify-attestation, cosign download, oras discover, crane manifest for the image/signature/attestation) is attached below for reference
[Chains output Analysis - Impact of Referrer's API changes.pdf](https://github.com/user-attachments/files/29788652/Chains.output.Analysis.-.Impact.of.Referrer.s.API.changes.pdf)

### Summary
 
| Scenario | `storage.oci.encoding-format` | Storage layout | Payload format | `.sig`/`.att` tags | `oras discover` |
|---|---|---|---|---|---|
| 1. `main` (pre-PR baseline) | unset | tag-based | DSSE | created | empty |
| 2. PR branch, key unset | unset (implicit) | tag-based | DSSE | created | empty |
| 3. PR branch, explicit | `dsse` | tag-based | DSSE | created | empty |
| 4. PR branch, referrers | `sigstore-bundle` | OCI 1.1 Referrers API | Sigstore protobuf bundle | **none** | **populated** |
 
Scenarios 1–3 confirm the default/legacy path is unchanged (no behavior change when the key is unset or set to `dsse`). Scenario 4 is the new path.

anithapriyanatarajan:
@tektoncd/chains-collaborators - Request review of this. Thank you

arewm (approved):

pkg/chains/storage/oci/attestation.go
```diff
@@ -101,14 +139,110 @@ func (s *AttestationStorer) Store(ctx context.Context, req *api.StoreRequest[nam
...
 	}
-	logger.Infof("Successfully uploaded attestation for %s", req.Artifact.String())
 
+	// req.Bundle.Signature is already a complete DSSE envelope (JSON) produced by
+	// the wrapped signer: its signature is computed over the DSSE PAE. MakeNewBundle
+	// expects exactly this envelope JSON as its `sig` argument - it extracts the
+	// PayloadType and the raw signature from it. Re-wrapping it in another envelope
+	// would place the whole envelope JSON into the inner sig field, producing a
+	// bundle whose signature does not verify ("Found: 0").
+	var rekorEntry *models.LogEntryAnon
```
infernus01: rekorEntry is always nil here so when transparency is enabled and Chains actually got a receipt back from Rekor, it never makes it into the bundle. That means the bundle ships without a tlog entry, and cosign verify-attestation (which checks for one by default since v2) will reject it.

anithapriyanatarajan:
Test prompt used for this PR : [pr1691-test-prompt.md](https://github.com/user-attachments/files/30338584/pr1691-test-prompt.md)
Test Results:  [pr1691-test-results-v2.md](https://github.com/user-attachments/files/30338605/pr1691-test-results-v2.md)

PR #1736, opened 2026-06-29
fix(signing): add writable volume for TUF cache in distroless container

ab-ghosh:
# Changes

When signers.x509.tuf.mirror.url is set to a custom TUF mirror (e.g. one provided by the Trusted Artifact Signer operator), the sigstore TUF library tries to create a local cache directory. In the Chains distroless container, HOME is /home/nonroot (user 65532/nonroot) and readOnlyRootFilesystem: true is set, causing:

`error configuring x509 signer: initialize tuf: creating cached local store: mkdir /.sigstore: read-only file system`

This blocks artifact signing when using private sigstore deployments.

**Fix:** Add a Memory-backed emptyDir volume (sigstore-cache) mounted at /home/nonroot, and set TUF_ROOT=/home/nonroot/.sigstore/root. The sigstore TUF library checks TUF_ROOT before falling back to $HOME/.sigstore/root (see client.go rootCacheDir()).
Using TUF_ROOT scopes the change precisely to TUF cache behaviour without altering the process-wide HOME for any other code path.

The default TUF mirror (https://tuf-repo-cdn.sigstore.dev) is unaffected since it skips local TUF initialization entirely (trust root is embedded in the binary).

**Testing:**
  - Unit test added that reproduces the read-only failure and verifies the fix
  - Manually verified on a kind cluster with a custom TUF mirror URL - confirmed the `read-only file system` error is gone after the fix

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Fix TUF mirror initialization failure (`mkdir /home/nonroot/.sigstore: read-only file system`) when using a custom `signers.x509.tuf.mirror.url` with private sigstore deployments (e.g. Red Hat Trusted Artifact Signer). Added a Memory-backed `emptyDir` volume at `/home/nonroot` and set `TUF_ROOT=/home/nonroot/.sigstore/root` to give the TUF client a writable, ephemeral cache scoped specifically to TUF behaviour. Public sigstore users
(no custom TUF mirror URL) are unaffected.
```

### DIFF
--- a/config/100-deployment.yaml
+++ b/config/100-deployment.yaml
@@ -114,7 +114,11 @@ spec:
               mountPath: /etc/signing-secrets
             - name: oidc-info
               mountPath: /var/run/sigstore/cosign
+            - name: sigstore-cache
+              mountPath: /home/nonroot/.sigstore
           env:
+            - name: TUF_ROOT
+              value: /home/nonroot/.sigstore/root
             - name: SYSTEM_NAMESPACE
               valueFrom:
                 fieldRef:
@@ -141,6 +145,10 @@ spec:
         - name: signing-secrets
           secret:
             secretName: signing-secrets
+        - name: sigstore-cache
+          emptyDir:
+            medium: Memory
+            sizeLimit: 50Mi
         - name: oidc-info
           projected:
             sources:
--- a/pkg/chains/signing/x509/x509_test.go
+++ b/pkg/chains/signing/x509/x509_test.go
@@ -19,6 +19,8 @@ import (
 	"crypto/ed25519"
 	"crypto/sha256"
 	"encoding/json"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"strings"
@@ -144,6 +146,35 @@ func TestCreateSignerFulcioEnabledFilesystemProvider(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
+}
+
+// TestInitializeTUFWithWritableCacheDir verifies that initializeTUF() succeeds
+// past the filesystem step when TUF_ROOT points to a writable directory. This
+// exercises the real code path that fails in the Chains distroless container
+// when readOnlyRootFilesystem is true and no writable cache location is
+// provided (see SRVKP-9439). The mock server serves invalid TUF metadata, so
+// we expect a TUF validation error — but NOT a filesystem error.
+func TestInitializeTUFWithWritableCacheDir(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.Write([]byte(`{"signed":{"_type":"root","version":1},"signatures":[]}`))
+	}))
+	defer ts.Close()
+
+	writableDir := t.TempDir()
+	t.Setenv("TUF_ROOT", filepath.Join(writableDir, ".sigstore", "root"))
+
+	ctx := logtesting.TestContextWithLogger(t)
+	err := initializeTUF(ctx, ts.URL)
+	if err == nil {
+		return
+	}
+	if strings.Contains(err.Error(), "read-only file system") ||
+		strings.Contains(err.Error(), "permission denied") ||
+		strings.Contains(err.Error(), "creating cached local store") {
+		t.Fatalf("Got filesystem error with writable TUF_ROOT: %v", err)
+	}
+	t.Logf("TUF init got past filesystem step (expected metadata error): %v", err)
 }
 
 func TestSigner_SignECDSA(t *testing.T) {


config/100-deployment.yaml
```diff
@@ -141,6 +145,8 @@ spec:
         - name: signing-secrets
           secret:
             secretName: signing-secrets
+        - name: sigstore-cache
+          emptyDir: {}
```
anithapriyanatarajan: Consider using `medium: Memory` with a size cap for this volume:

    - name: sigstore-cache
      emptyDir:
        medium: Memory
        sizeLimit: 50Mi

The TUF cache only holds public metadata (root.json, tuf.db, targets), a few MB at most.
Memory-backed tmpfs means that data never touches the node's disk, so there are no
forensic traces on the node after the pod is deleted. It also evicts instantly on pod
termination rather than waiting for the kubelet GC cycle.

The tradeoff is that it counts against the pod's memory limit, but 50Mi is well within
normal headroom for a controller pod. Not a blocker, just a hardening suggestion.

config/100-deployment.yaml
```diff
@@ -114,7 +114,11 @@ spec:
               mountPath: /etc/signing-secrets
             - name: oidc-info
               mountPath: /var/run/sigstore/cosign
+            - name: sigstore-cache
+              mountPath: /home/nonroot
           env:
+            - name: HOME
```
anithapriyanatarajan: Setting HOME works correctly here since os.UserHomeDir() reads $HOME, but a better alternative would be to set TUF_ROOT directly:

    - name: TUF_ROOT
      value: /home/nonroot/.sigstore/root

The sigstore TUF library checks TUF_ROOT before falling back to $HOME/.sigstore/root
(see vendor/github.com/sigstore/sigstore/pkg/tuf/client.go rootCacheDir()).
Using TUF_ROOT scopes the change precisely to TUF cache behaviour without altering the process-wide HOME for any other library or future code path that might also write to it.

Either approach is safe. Please check this option so that we could limit the impact specific to TUF Cache.

anithapriyanatarajan (changes requested):

anithapriyanatarajan:
@ab-ghosh PR https://github.com/tektoncd/chains/pull/1738 has been merged to main and fixes the E2E CI failures. Please rebase your branch onto main so the E2E tests can pass.

config/100-deployment.yaml
ab-ghosh: Thank you, acknowledged. Added medium: Memory with sizeLimit: 50Mi

config/100-deployment.yaml
ab-ghosh: Thank you for the suggestion, switched to TUF_ROOT, much cleaner since it scopes the change precisely to TUF cache behavior without any process-wide side effects. Updated the test to reflect this as well.

anithapriyanatarajan:
/lgtm
@ab-ghosh Thank you for this work. Happy to merge :heart_eyes_cat: 

Short blog on what was verified feature wise: https://gist.github.com/anithapriyanatarajan/3d3c6b0678bf76faf2f6fa17a1b11f06#file-blog-pr1736-tuf-fix-md.

config/100-deployment.yaml
```diff
@@ -114,7 +114,11 @@ spec:
               mountPath: /etc/signing-secrets
             - name: oidc-info
               mountPath: /var/run/sigstore/cosign
+            - name: sigstore-cache
+              mountPath: /home/nonroot
```
jkhelil: we should mount  only /home/nonroot/.sigstore ? liek in this example https://github.com/tektoncd/chains/blob/main/examples/chains-deployment-keyless.yaml#L115

pkg/chains/signing/x509/x509_test.go
```diff
@@ -146,6 +148,56 @@ func TestCreateSignerFulcioEnabledFilesystemProvider(t *testing.T) {
...
+			t.Fatalf("Failed to make dir read-only: %v", err)
+		}
+		t.Cleanup(func() { os.Chmod(readOnlyDir, 0o755) })
+
+		cacheDir := filepath.Join(readOnlyDir, ".sigstore", "root")
+		err := os.MkdirAll(cacheDir, 0o700)
+		if err == nil {
+			t.Fatal("Expected cache dir creation to fail on read-only directory, but it succeeded")
+		}
+		t.Logf("Bug reproduced — cache dir creation fails: %v", err)
```
jkhelil: HOw does this reproduced the issue, the issue is related to chains container where the path expected by TUF, HOME or TUF_HOME is read only
this test is misleading, it doesnt call init tuf

config/100-deployment.yaml
ab-ghosh: Thank you, updated the mount path to /home/nonroot/.sigstore

pkg/chains/signing/x509/x509_test.go
ab-ghosh: Thank you, fair point. Replaced it with a single test that calls initializeTUF() directly with a mock TUF server and TUF_ROOT pointing to a writable temp directory. It verifies that the real code path gets past the filesystem step - the only error returned is about TUF metadata validation, not a filesystem error.

infernus01 (approved):

Issue #1160, opened 2024-07-18
Promote SLSAv1.0 offering of Chains format to `beta`


chitrangpatel:
Since SLSAv1.0 was released, Chains provided support for it starting with provenance format `slsa/v2alpha2`. At this point, it is `slsa/v2alpha4` (which includes support for StepActions and StepResults).

We propose stabilizing this to a Beta version: `v2beta1` so that there is reduced immediate breaking changes and our users can adopt it more. 

Here, I highlight some missing items:
- Implement support for [Tekton Artifacts](https://github.com/tektoncd/community/blob/main/teps/0147-tekton-artifacts-phase1.md). The final PR is under review and we can introduce support for it in Chains right after.
- Identify the name of the pipelineTask/pipelineRun/taskRun/Step using the remote resource (Pipeline, Task, StepAction) in the resolved dependencies and byProducts. This will help with completeness.
- Capture the origin of a remote StepAction (i.e. uri and digest) in the resolved dependencies: https://github.com/tektoncd/pipeline/pull/8106

chitrangpatel:
cc @lcarva @renzodavid9

chitrangpatel:
@renzodavid9 please add any other items here that I may have missed.

renzodavid9:
With https://github.com/tektoncd/pipeline/pull/8106 merged we can implement last point from https://github.com/tektoncd/chains/issues/1160#issue-2417114302, I'll take a look an open a PR for it

PuneetPunamiya:
@chitrangpatel @renzodavid9 
I'll work on this part 
* Identify the name of the pipelineTask/pipelineRun/taskRun/Step using the remote resource (Pipeline, Task, StepAction) in the resolved dependencies and byProducts. This will help with completeness.

PR #1737, opened 2026-06-30
Capture remote stepAction source in SLSA resolvedDependencies

ngelman1:
This PR aims to close some gaps before advancing to reslove  issue #1160

# Changes

When a TaskRun step uses a remotely resolved stepAction, Pipelines records its origin in `StepState.Provenance.RefSource`. This change wires that data into the SLSA provenance resolvedDependencies as entries named "`stepAction/{stepName}`", enabling full traceability of step definitions.

Currently, Chains captures the origin of remote Tasks and Pipelines in provenance, but ignores the origin of remote StepActions. This means the SLSA provenance has a traceability gap -- a verifier cannot tell where a step's definition came from.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Capture the origin (URI + digest) of remotely resolved StepActions in SLSA resolvedDependencies as entries named "stepAction/{stepName}", enabling full traceability of step definitions in TaskRun and PipelineRun provenance.
Requires artifacts.taskrun.format: slsa/v2alpha4.
```


anithapriyanatarajan:
@ngelman1 - Thank you for the PR. Please consider incorporating the points:
1. **Add e2e test** to handle this scenario, that covers the below:

>1. Submit a TaskRun whose taskSpec.steps[0] references a remote StepAction via the git resolver
> 2. Wait for Chains to sign and annotate it
>3.  Fetch and decode the SLSA attestation
>4. Assert predicate.resolvedDependencies contains {"name": "stepAction/<stepName>", "uri": "<git-url>", "digest": {...}}

  
2. **Update release notes** to reflect something like below:
`Capture the origin (URI + digest) of remotely resolved StepActions in SLSA resolvedDependencies as entries named "stepAction/{stepName}", enabling full traceability of step definitions in TaskRun and PipelineRun provenance.`

ngelman1:
> @ngelman1 - Thank you for the PR. Please consider incorporating the points:
> 
> 1. **Add e2e test** to handle this scenario, that covers the below:
> 
> > 1. Submit a TaskRun whose taskSpec.steps[0] references a remote StepAction via the git resolver
> > 2. Wait for Chains to sign and annotate it
> > 3. Fetch and decode the SLSA attestation
> > 4. Assert predicate.resolvedDependencies contains {"name": "stepAction/", "uri": "", "digest": {...}}
> 
> 2. **Update release notes** to reflect something like below:
>    `Capture the origin (URI + digest) of remotely resolved StepActions in SLSA resolvedDependencies as entries named "stepAction/{stepName}", enabling full traceability of step definitions in TaskRun and PipelineRun provenance.`

Added! thanks for commenting
(btw, I used a stepAction task from pipelines. we do not have anyhthing in this repo. Should I add a simple one under example?)

enarha:
## Code Review

Thanks for the PR — the direction is right and the general approach is sound. Four issues worth addressing before merge:

### 1. `fromRemoteStepActions` bypasses the `WithStepActionsResults` gate (all format versions affected)

`material.FromStepActionsResults` at line 214 of `resolved_dependencies.go` is inside `if opts.WithStepActionsResults {}`. The new `fromRemoteStepActions(tro)` appended after the closing brace is not. `v2alpha3` callers pass `resolveddependencies.ResolveOptions{}` (zero value) — see `v2alpha3/internal/taskrun/taskrun.go:39` and `v2alpha3/internal/pipelinerun/pipelinerun.go:43` — so v2alpha3 attestations now silently receive `stepAction/` entries that were never part of that format's contract. Only v2alpha4 sets `WithStepActionsResults: true`.

Whether to gate this under `WithStepActionsResults` or a new dedicated flag is a design call, but the current unconditional behavior is inconsistent with how every other step-action data point is gated.

### 2. `RemoveDuplicateResolvedDependencies` silently drops entries when two steps share the same remote StepAction revision

The dedup function (line 98) keys on `{Uri, Digest, Content}` — the `Name` field is explicitly stripped before hashing. Its bypass guard is `if !(name == TaskConfigName || name == PipelineConfigName)`. If two steps in the same TaskRun both reference the same StepAction at the same revision (identical URI+Digest), `fromRemoteStepActions` emits two descriptors — e.g. `stepAction/build` and `stepAction/test` — that are distinct only by name. The second entry is silently dropped because `StepActionConfigName` is not in the exception list, so the attestation records only one of the two steps as depending on the remote action.

The test suite only covers the case where each step has a distinct digest — the same-digest scenario is not tested and would currently produce an incomplete attestation.

### 3. E2E test pins to a mutable `main` branch in an external repo

`test/e2e_test.go` — the git resolver params use `revision: "main"` against `https://github.com/tektoncd/pipeline`. The file `examples/v1/taskruns/stepaction.yaml` exists today, but `main` is a moving ref. If that file is renamed, moved, or the resolver behavior changes, the test breaks for reasons unrelated to Chains. Existing Chains e2e tests that use the git resolver pin to a specific commit SHA or tag rather than a branch head — recommend doing the same here.

### 4. Empty URI passes through into the attestation without validation

`GetRemoteStepActions` in `objects.go` filters on `step.Provenance != nil && step.Provenance.RefSource != nil` but does not check `RefSource.URI != ""`. A resolver that populates `RefSource` with only a digest (no URI) produces a `ResourceDescriptor{Name: "stepAction/x", Uri: ""}` — structurally invalid per SLSA. The analogous task-level helper `AddSLSATaskDescriptor` returns `nil` when URI is absent; aligning `GetRemoteStepActions` with that pattern (skip or log-and-skip entries with an empty URI) would be consistent.

ngelman1:
@enarha 
1. 100% agree, fixed. (I added `opt` to the function arguments and updated the calls  (not sure why would we need a new flag for this?)
2. what I added pretty much eliminates the possibility to remove stepAcrion dependencies, but I guess that ok since there cant be 2 in the same step..? (and from differnt steps we would like to keep them both)
3. fixed
4. Im not sure I fully understand the `AddSLSATaskDescriptor` reference, since it assures the RefSource is  not nil just as I did, but anyway, I added another `if` statment to catch all of the URI and the opposite digest case. Is this what you meant?

enarha:
> 4. Im not sure I fully understand the `AddSLSATaskDescriptor` reference, since it assures the RefSource is  not nil just as I did, but anyway, I added another `if` statment to catch all of the URI and the opposite digest case. Is this what you meant?

No. Nothing ensures Uri is not an empty string. A nil/empty `Digest` is less critical (some resolvers may omit it), but an empty URI produces an unresolvable SLSA entry. Let me suggest a check/fix inline.

pkg/chains/formats/slsa/internal/resolved_dependencies/resolved_dependencies.go
```diff
@@ -191,6 +198,20 @@ func fromPipelineTask(logger *zap.SugaredLogger, pro *objects.PipelineRunObjectV
 	return resolvedDependencies, nil
 }
 
+// fromRemoteStepActions converts remote StepAction provenance into resolved dependencies
+func fromRemoteStepActions(tro *objects.TaskRunObjectV1) []*intoto.ResourceDescriptor {
+	var rds []*intoto.ResourceDescriptor
+	for _, stepProv := range tro.GetRemoteStepActions() {
```
enarha, line 204:
```suggestion
	for _, stepProv := range tro.GetRemoteStepActions() {
	    if stepProv.Provenance.RefSource.URI == "" {
	        continue
	    }
```

enarha:
> > 4. Im not sure I fully understand the `AddSLSATaskDescriptor` reference, since it assures the RefSource is  not nil just as I did, but anyway, I added another `if` statment to catch all of the URI and the opposite digest case. Is this what you meant?
> 
> No. Nothing ensures Uri is not an empty string. A nil/empty `Digest` is less critical (some resolvers may omit it), but an empty URI produces an unresolvable SLSA entry. Let me suggest a check/fix inline.

@ngelman1 I actually retract that recommendation. It produces a verifiable attestation, but hides attestation for a step. It'll be better to leave it as is (no empty Uri check) and leave it to the verifier to handle this. Please revert that last change. That's an edge case anyway.

anithapriyanatarajan:
@ngelman1. Thanks again for the awesome work.  A kind cluster test confirms the feature works correctly. Two doc-related asks before merge:

---

**1. Update the release note to include a format qualifier**

The current note says "TaskRun and PipelineRun provenance" without qualification. Users on `slsa/v2alpha3` (a widely used format) will find no `stepAction/` entries and have no way to know it's a format restriction rather than a bug.

Suggested update:

```
Capture the origin (URI + digest) of remotely resolved StepActions in SLSA resolvedDependencies as entries named "stepAction/{stepName}", enabling full traceability of step definitions in TaskRun and PipelineRun provenance.
Requires artifacts.taskrun.format: slsa/v2alpha4.

Note: slsa/v2alpha3 and slsa/v1 are frozen formats and will not receive this data. If you are on v2alpha3 and want StepAction origin traceability, switch to slsa/v2alpha4.
```

---

**2. Add `slsa/v2alpha4` to `pkg/chains/formats/slsa/README.md`**

The format table in `pkg/chains/formats/slsa/README.md` currently stops at `v2alpha3`.  `v2alpha4` is entirely absent. Since `docs/config.md` describes both `v2alpha3` and `v2alpha4` as _"Recommended format for new chains users"_, users have no basis to choose between them. A new row would help:

| Tekton Chains Provenance Format | SLSA predicate | Notes |
|---|---|---|
| **slsa/v2alpha4** | **slsa v1.0** | Extends `v2alpha3` with StepAction-aware provenance: captures type-hinted StepAction results and the remote origin (URI + digest) of each step's definition. Use this format if your TaskRun steps reference remote StepActions via a resolver. |

Without this, the release note's format requirement has no documentation to point users to for context.

PR #1756, opened 2026-07-06
Fix: TaskRuns stuck blocked from deletion due to old chains finaliser

### LINKED ISSUE #1755: TaskRuns stuck on deletion after Chains update
# Summary
TaskRuns started with Chains v0.24.0 are stuck unable to be deleted after update to Chains v0.27.x. Since Chains v0.27.0 started to use service-side-apply method to manipulate finalizers and `chains.tekton.dev` was applied using patch-merge method, the operator cannot remove it. As a result, old pre-update started TaskRuns cannot get deleted without human intervention. 

# Expected Behavior
TaskRuns started on earlier releases of Chains controller than v0.27.x should get deleted successfully when requested. 

# Actual Behavior
TaskRuns starter on Chains v0.24.0 are stuck and cannot be deleted due to presence of the old `chains.tekton.dev` finalizer after the update to Chains v0.27.x

1. Deploy Chains v0.24.0
2. Start a TaskRun . Check that the TaskRun gets `chains.tekton.dev` finalizer in its metadata added.
3. Deploy Chains v0.27.x 
4. Wait a little and describe the TaskRun. Check that a new finalizer `chains.tekton.dev/taskrun` is added to the metadata. At this point, there should be 2 finalizers listed on the TaskRun: `chains.tekton.dev` and `chains.tekton.dev/taskrun`
5. Run `kubectl delete tr <name>` to delete the TaskRun. Witness that the deletion is stuck.  
6. Describe the stuck TaskRun. It will have only `chains.tekton.dev` finalizer listed, since the newer one will get removed. 

# Additional Info

- Kubernetes version:
```
Client Version: v1.35.6
Kustomize Version: v5.7.1
Server Version: v1.28.0
```

- Tekton Pipeline version:
`version: v1.12.0`

Liu-ko:
Chains v.0.27.0 started to use Service-Side-Apply mechanism to manage finalisers. SSA mechanism cannot modify finalisers previously added by merge patch method. As a result, deletion of TaskRuns that were triggered before the Chains update and got old finalisers is blocked forever.

This change is updating the existing `removeFinalizerViaMergePatch` to also cleanup the old finaliser and allow future deletion of a TaskRun.

Fixes: #1755

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Fixes: Unblocked TaskRun deletions stuck on legacy finalizers after updating to v0.27.0.
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -180,6 +180,9 @@ func TestFinalizeKind_SSAMigration(t *testing.T) {
 	mergePatchFieldsV1 := metav1.NewFieldsV1(
 		`{"f:metadata":{"f:finalizers":{".":{},"v:\"chains.tekton.dev/taskrun\"":{}}}}`,
 	)
+	mergePatchFieldsV1Old := metav1.NewFieldsV1(
+		`{"f:metadata":{"f:finalizers":{".":{},"v:\"chains.tekton.dev\"":{}}}}`,
+	)
 	ssaFieldsV1 := metav1.NewFieldsV1(
 		`{"f:metadata":{"f:finalizers":{".":{},"v:\"chains.tekton.dev/taskrun\"":{}}}}`,
 	)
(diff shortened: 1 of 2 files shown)


jkhelil:
@enarha Can you review please ?

enarha (commented):
Thank you for the fix. I'll suggest the following instead. Since finalizer with the old name is guaranteed to be set via merge patch, we can simply use the `removeFinalizerViaMergePatch` function for it. Something like this will be sufficient IMO
```
  func (r *Reconciler) removeFinalizerViaMergePatch(ctx context.Context, tr *v1.TaskRun) error {
      const oldFinalizerName = "chains.tekton.dev" // add the const
      var newFinalizers []string
      for _, f := range tr.Finalizers {
          if f != taskRunFinalizerName && f != oldFinalizerName {  // one extra condition
              newFinalizers = append(newFinalizers, f)
          }
      }
      // ... unchanged
```
This only adds the const and the extra `f != oldFinalizerName` condition. Then we can drop the `removeOldFinalizerIfExists` function and its calls entirely. I would also like to have an extra test for this case.
On a separate note, can you please add a short release note as I would like to see it included in the release notes.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -278,6 +281,62 @@ func TestFinalizeKind_SSAMigration(t *testing.T) {
...
+					Annotations:       map[string]string{annotations.ChainsAnnotation: "true"},
+					ManagedFields: []metav1.ManagedFieldsEntry{
+						{
+							Operation: metav1.ManagedFieldsOperationUpdate,
+							FieldsV1:  mergePatchFieldsV1Old,
+						},
+						{
+							Operation: metav1.ManagedFieldsOperationApply,
+							FieldsV1:  ssaFieldsV1,
+						},
```
enarha, line 301: nit: Is there a reason to include both finalizers here (won't only the old one be enough)? It's otherwise the same test as the next one.

enarha (commented):
Apart from a nit, the rest is LGTM.
Just please update the commit and PR message with the new implementation. We are not updating the existing removeOldFinalizerIfExists function anymore, it's been removed.
Thank you.

pkg/reconciler/taskrun/taskrun_test.go
Liu-ko: I think that from the point of view of the current function under test there is no difference. But if it changes later, this should cover the case to ensure that the new finalizer is not a requirement for the cleanup as well as it is not a blocker for the cleanup of the old finalizer.

enarha (commented):
/lgtm

enarha:
@jkhelil @ngelman1 @infernus01 PTAL

Liu-ko:
@enarha what is the process for deciding whether this should be cherrypicked and fixed on 0.27.x and (presumably) 0.28.x ?

enarha:
> @enarha what is the process for deciding whether this should be cherrypicked and fixed on 0.27.x and (presumably) 0.28.x ?

I do not think there is a established process, at least nothing documented to be best of my knowledge. But for fixes it usually makes sense to cherry-pick. The other thing is to decide on the releases including those cherry-picks which from the effort POV is the "harder" thing. Usually will do for security fixes or critical bugs, otherwise we just advice to update to the version including the fix.

PR #1772, opened 2026-07-08
fix(deps): pin otel/sdk to v1.43.0 to avoid semconv schema conflict

infernus01:
# Changes
  Pin `go.opentelemetry.io/otel/sdk` and `otel/sdk/metric` to v1.43.0 to fix controller panic on startup.

  ### Problem

PR #1768 bumped `otel/sdk` from v1.43.0 to v1.44.0. The SDK v1.44.0 switched its internal semconv import from `v1.40.0` to `v1.41.0`, changing the schema URL returned by `resource.Default()`. Meanwhile, `knative.dev/pkg`'s semconv wrapper is still pinned to `v1.40.0`.

When `knative.dev/pkg/observability/resource.Default()` merges these two resources, `resource.Merge` returns a "conflicting Schema URL" error and the code panics:
```
  panic: conflicting Schema URL:
    https://opentelemetry.io/schemas/1.41.0 and
    https://opentelemetry.io/schemas/1.40.0
```

This caused all 6 e2e tests to fail — the controller enters CrashLoopBackOff immediately after deployment patching.

  ### Why other repos aren't affected

`tektoncd/pruner` has the same `knative.dev/pkg` and `otel v1.44.0`, but its `otel/sdk` stayed at v1.43.0 because `otel/sdk/metric` is an indirect dependency there — dependabot only bumped the direct deps.
In Chains, `otel/sdk/metric` is a direct dependency (imported in metrics test files), so dependabot bumped it to v1.44.0 along with the SDK.

### Fix

Pin `otel/sdk` and `otel/sdk/metric` to v1.43.0 (semconv v1.40.0), matching `knative.dev/pkg`. This is the same version combination that works in `tektoncd/pruner`.
Once `knative.dev/pkg` upstream updates their semconv wrapper to v1.41.0, the SDK can be bumped to v1.44.0+ without conflict.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


infernus01 (commented):
/kind bug

PR #1789, opened 2026-07-09
Move lcarva to emeritus_approvers

lcarva:
Since I have not been involved with the Tekton community recently, and I expect that to continue, I'd like to step down and have my access revoked in the interest of good security hygiene. No sense in keeping permissions I'm not using.

I'm happy to keep my name under emeritus_approvers so folks can more easily ping me if needed.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -6,7 +6,6 @@ approvers:
 - enarha
 - infernus01
 - jkhelil
-- lcarva
 - PuneetPunamiya
 - vdemeester
 - wlynch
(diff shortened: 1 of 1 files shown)


enarha (commented):
/lgtm

anithapriyanatarajan:
:crying_cat_face: @vdemeester - tekton-robot needs a fix I guess. It should not have reacted with :tada: for this PR.  :sob:
@lcarva Thank you very much for mentoring us. we are missing your presence. Hope we could tag you on any TEPs around chains in the community for architectural guidance.

Could you please add a Alumni section to the PR - like here if possible https://github.com/tektoncd/pipeline/blob/main/OWNERS_ALIASES

jkhelil:
@lcarva Louis Thank you for the all the knowledge that you have shared with the team.

PR #1778, opened 2026-07-09
fix(release): bump pinned ko image for go 1.26.4

jkhelil:
# Changes

The patch-release pipeline runs `ko resolve` inside the pinned
`ghcr.io/tektoncd/plumbing/ko` image, which bakes in a specific Go
toolchain. `go.mod` requires `go 1.26.4`, but the pinned `ko` image in
`release/publish.yaml` still shipped `GOLANG_VERSION=1.26.3`. Since
`GOTOOLCHAIN=local`, `ko resolve` refused to auto-fetch the newer
toolchain and the release failed with:

```
go: go.mod requires go >= 1.26.4 (running go 1.26.3; GOTOOLCHAIN=local)
```

This mirrors how `tektoncd/pipeline` avoids this class of failure: it
has a `docker` Dependabot ecosystem entry for its `/tekton` directory
that keeps the pinned `ko`/`koparse` image digests continuously
up to date. Chains had no such entry, so the pinned digests could
silently drift behind `go.mod`.

- Bump `ko` and `koparse` image digests in `release/publish.yaml` to
  versions shipping `GOLANG_VERSION=1.26.4` (verified via
  `crane config`).
- Add a `docker` package-ecosystem entry for `/release` in
  `.github/dependabot.config.yml` and regenerate
  `.github/dependabot.yml` via `./hack/generate-dependabot.sh`, so
  Dependabot keeps these pinned image digests in sync with `go.mod`
  going forward.

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

``` release-note
NONE
```

Made with [Cursor](https://cursor.com)

PR #1791, opened 2026-07-10
fix(release): copy vendor tarball into kodata instead of symlink

jkhelil:
## Changes

`ko` v0.19.0 enforces that `kodata/` symlinks must not resolve outside the
kodata root (ko-build/ko#1619), a deliberate security hardening. This broke
the `run-ko` step in our release pipeline, which symlinked
`cmd/*/kodata/source.tar.gz` to a tarball in an external `mktemp -d`
directory:

```
Error: error processing import paths in ".../config/100-deployment.yaml":
error resolving image references: tarring kodata: kodata symlink
".../cmd/controller/kodata/source.tar.gz" resolves to "/tmp/tmp.gGIAFn/source.tar.gz"
which is outside the kodata root ".../cmd/controller/kodata"
```

This surfaced during a `v0.28.0` patch release after bumping the pinned
`ko` image digest for Go 1.26.4 support (#1778), which pulled in `ko` v0.19.1.

Fix: copy the vendor tarball into each `cmd/*/kodata/` directory instead of
symlinking to it from a shared tmpdir. `ko` already dereferences symlinks
and embeds the resolved file's bytes into the image layer when tarring
kodata, so this has no effect on the final image size — only the symlink
itself was rejected, not the underlying content.

Applied the same fix to the historical `examples/releases/v0.3.0-build-chains-taskrun.yaml`
for consistency, since it has the identical pattern.

Note: `tektoncd/pipeline`'s `tekton/publish.yaml` has the same unfixed
pattern and is pinned to the same `ko` v0.19.1 image family, so it's likely
exposed to the same issue.

## Submitter Checklist

- [x] Has [Tests](https://github.com/tektoncd/community/blob/master/standards.md#tests) included if any functionality added or changed — N/A, release pipeline script fix
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/master/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/master/standards.md)
- [ ] Has a kind label. You can add one by adding a comment on this PR that contains `/kind bug`
- [x] Release notes block below has been updated with any user facing changes

```release-note
Fix release pipeline `ko resolve` failure caused by `ko` >= v0.19.0 rejecting
the `kodata/source.tar.gz` symlink used to bundle vendored source.
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/release/publish.yaml
+++ b/release/publish.yaml
@@ -137,11 +137,15 @@ spec:
       # probably get it wrong), we'll just targz up the whole vendor tree and
       # include it. As of 9/20/2019, this amounts to about 11MB of additional
       # data in each image.
+      # ko >= v0.19.0 rejects kodata symlinks that resolve outside the
+      # kodata root (ko-build/ko#1619), so the tarball is copied into
+      # each kodata/ directory instead of being symlinked from a
+      # shared tmpdir.
       TMPDIR=$(mktemp -d)
       tar cfz ${TMPDIR}/source.tar.gz vendor/
       for d in cmd/*; do
         if [ -d ${d}/kodata/ ]; then
-          ln -s ${TMPDIR}/source.tar.gz ${d}/kodata/
+          cp ${TMPDIR}/source.tar.gz ${d}/kodata/
         fi
       done
 


vdemeester (approved):

jkhelil:
We cant wait for https://github.com/ko-build/ko/pull/1698 to be release

PR #1841, opened 2026-07-28
(feat)Support native Tekton Artifacts API in SLSA provenance

ngelman1:
# Changes
Today, Chains generates SLSA provenance for TaskRuns by extracting input and output artifacts from type-hinted results -specially named results like *IMAGE_URL, *IMAGE_DIGEST, CHAINS-GIT_COMMIT, and *ARTIFACT_OUTPUTS. If a result name doesn't match the expected pattern, Chains silently ignores it.

This PR is wiring the Tekton native artifact API ([TEP-0147](https://github.com/tektoncd/community/blob/main/teps/0147-tekton-artifacts-phase1.md))  into Chains' SLSA provenance generation.
This allows steps to explicitly declare their inputs and outputs by writing a JSON file to $(step.artifacts.path). The Pipelines controller then populates `status.artifacts` and `status.steps[].inputs/outputs` on the TaskRun with structured data including URIs, digests, and a buildOutput flag.
### Artifact mapping
| Artifact field | `buildOutput` | SLSA provenance field |
|---|---|---|
| `inputs[].values` | n/a | `resolvedDependencies` |
| `outputs[].values` | `true` | `subject` |
| `outputs[].values` | `false` | `byproducts` |

This does not replace type-hinting, but adds the ability of chains to fetch inputs and outputs, without forcing the user to exactly match naming convetions.

**Tested on a kind cluster
# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Chains now supports the native Tekton Artifacts API (TEP-0147) for SLSA provenance generation. Task authors can declare build inputs and outputs via `$(step.artifacts.path)` as an alternative to type-hinted results, with no naming conventions required.
```

### DIFF
--- a/docs/slsa-provenance.md
+++ b/docs/slsa-provenance.md
@@ -417,6 +417,61 @@ spec:
 
 Chains Will read `first-ARTIFACT_OUTPUTS` and `second-IMAGE_URL/second-IMAGE_DIGEST` from the StepAction and classify them as a `subject`.
 
+## Native Tekton Artifacts API
+This is a native, structured way to declare build inputs and outputs using the
+[Tekton Artifacts API (TEP-0147)](https://github.com/tektoncd/community/blob/main/teps/0147-tekton-artifacts-phase1.md).
+This is an alternative to type hinting that provides richer, more explicit
+artifact declarations.
+### Prerequisites
+The Artifacts API is gated behind a feature flag in Tekton Pipelines: "enable-artifacts"
+### How it works:
+Instead of writing to specially named results, steps write a structured JSON file to $(step.artifacts.path). The Tekton Pipelines controller reads this file and populates status.artifacts and status.steps[].inputs/outputs on the TaskRun.
+
+Chains then maps these artifacts into the SLSA provenance. 
+Example TaskRun: 
+
+```yaml
+apiVersion: tekton.dev/v1
+kind: TaskRun
+metadata:
+  name: build-with-artifacts
+spec:
+  taskSpec:
+    steps:
+      - name: build-and-push
+        image: gcr.io/kaniko-project/executor
+        script: |
+          #!/usr/bin/env bash
+          # Build and push the image
+          buildah push myimage quay.io/myorg/myapp --digestfile /tmp/digest.txt
+          DIGEST=$(cat /tmp/digest.txt)
+
+          # Declare artifacts
+          cat > $(step.artifacts.path) << EOF
+          {
+            "inputs": [
+              {
+                "name": "source-code",
+                "values": [{
+                  "uri": "git+https://github.com/myorg/myrepo.git",
+                  "digest": {"sha256": "7f2f...."}
+                }]
+              }
+            ],
+            "outputs": [
+              {
+                "name": "container-image",
+                "values": [{
+                  "uri": "quay.io/myorg/myapp",
+                  "digest": {"sha256": "${DIGEST}"}
+                }],
+                "buildOutput": true
+              }
+            ]
+          }
+          EOF
+```
+If a TaskRun uses both type hinting and native artifacts, Chains will include artifacts from both sources. Duplicates are removed automatically.
 
 ## Besides inputs/outputs
 
--- a/pkg/chains/formats/slsa/internal/resolved_dependencies/resolved_dependencies.go
+++ b/pkg/chains/formats/slsa/internal/resolved_dependencies/resolved_dependencies.go
@@ -239,6 +239,7 @@ func taskDependencies(ctx context.Context, opts ResolveOptions, tro *objects.Tas
 	}
 
 	mats = material.FromTaskParamsAndResults(ctx, tro)
+	mats = append(mats, material.FromNativeArtifactInputs(ctx, tro)...)
 	// convert materials to resolved dependencies
 	resolvedDependencies = append(resolvedDependencies, ConvertMaterialsToResolvedDependencies(mats, InputResultName)...)
 
--- a/pkg/chains/formats/slsa/v2alpha4/internal/taskrun/taskrun.go
+++ b/pkg/chains/formats/slsa/v2alpha4/internal/taskrun/taskrun.go
@@ -17,6 +17,7 @@ import (
 	"context"
 
 	intoto "github.com/in-toto/attestation/go/v1"
+	"github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/common"
 	"github.com/tektoncd/chains/pkg/chains/formats/slsa/extract"
 	"github.com/tektoncd/chains/pkg/chains/formats/slsa/internal/artifact"
 	builddefinition "github.com/tektoncd/chains/pkg/chains/formats/slsa/internal/build_definition"
@@ -66,8 +67,43 @@ func ByProducts(tro *objects.TaskRunObjectV1) ([]*intoto.ResourceDescriptor, err
 		return nil, err
 	}
 	byProd = append(byProd, res...)
+	byProd = append(byProd, byProductsFromNativeArtifacts(tro)...)
 
 	return byProd, nil
+}
+
+func byProductsFromNativeArtifacts(tro *objects.TaskRunObjectV1) []*intoto.ResourceDescriptor {
+	var byProds []*intoto.ResourceDescriptor
+
+	addNonBuildOutputs := func(outputs []v1.Artifact) {
+		for _, output := range outputs {
+			if output.BuildOutput {
+				continue
+			}
+			for _, val := range output.Values {
+				if val.Uri == "" || len(val.Digest) == 0 {
+					continue
+				}
+				digestSet := common.DigestSet{}
+				for algo, hex := range val.Digest {
+					digestSet[string(algo)] = hex
+				}
+				byProds = append(byProds, &intoto.ResourceDescriptor{
+					Name:   val.Uri,
+					Digest: digestSet,
+				})
+			}
+		}
+	}
+
+	if arts := tro.GetArtifacts(); arts != nil {
+		addNonBuildOutputs(arts.Outputs)
+	} else {
+		for _, step := range tro.GetStepArtifacts() {
+			addNonBuildOutputs(step.Outputs)
+		}
+	}
+	return byProds
 }
 
 // SubjectDigests returns the subjects detected in the given TaskRun. It takes into account taskrun and step results.
@@ -81,6 +117,7 @@ func SubjectDigests(ctx context.Context, tro *objects.TaskRunObjectV1) []*intoto
 
 	taskSubjects := extract.SubjectsFromBuildArtifact(ctx, tro.GetResults())
 	subjects = artifact.AppendSubjects(subjects, taskSubjects...)
+	subjects = artifact.AppendSubjects(subjects, subjectsFromNativeArtifacts(tro)...)
 
 	return subjects
 }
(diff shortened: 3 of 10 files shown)


pkg/artifacts/signable.go
```diff
@@ -234,6 +236,44 @@ func ExtractOCIImagesFromResults(ctx context.Context, results []objects.Result,
 	return objs
 }
 
+// ExtractOCIImagesFromArtifacts extracts OCI image references from native Tekton Artifacts (TEP-0147) outputs where BuildOutput is true.
```
waveywaves: TEP-0147’s `buildOutput` flag selects `subject` versus `byProduct`; it is not an OCI media-type marker. The API explicitly supports packages, reports, source artifacts, and purl URIs. Recommended forms such as `pkg:docker/...` fail this parser, while some non-OCI URI forms can be accepted as registry references. Could we avoid feeding every build output into `OCIArtifact`, or require a reliable OCI-specific signal first?

pkg/artifacts/signable.go
```diff
@@ -234,6 +236,44 @@ func ExtractOCIImagesFromResults(ctx context.Context, results []objects.Result,
...
+				}
+			}
+		}
+	}
+
+	if arts := obj.GetArtifacts(); arts != nil {
+		addArtifactOutputs(arts.Outputs)
+	}
+
+	for _, step := range obj.GetStepArtifacts() {
```
waveywaves: Pipeline retains step artifacts in `Status.Steps[].Outputs` and also merges them into `Status.Artifacts.Outputs`. Walking both collections therefore emits duplicate signable objects; I reproduced two objects for one artifact. Please consume the aggregate once, with a step-level fallback if it is absent, or deduplicate by canonical digest.

pkg/chains/formats/slsa/v2alpha4/internal/taskrun/taskrun.go
```diff
@@ -66,10 +67,42 @@ func ByProducts(tro *objects.TaskRunObjectV1) ([]*intoto.ResourceDescriptor, err
...
+				}
+				byProds = append(byProds, &intoto.ResourceDescriptor{
+					Name:   val.Uri,
+					Digest: digestSet,
+				})
+			}
+		}
+	}
+
+	for _, step := range tro.GetStepArtifacts() {
```
waveywaves: Pipeline also supports task-level artifacts written through `$(artifacts.path)`. Those appear in `Status.Artifacts` but not under an individual step, so this loop omits their non-build outputs from `byProducts`; I reproduced zero byproducts for one such output. Please read the aggregate here, with a step-level fallback if needed.

test/e2e_test.go
```diff
@@ -1099,3 +1099,93 @@ func TestRemoteStepActionProvenance(t *testing.T) {
...
+	t.Logf("Statement subject count: %d", len(statement.Subject))
+	for _, s := range statement.Subject {
+		t.Logf("  Subject: %s %v", s.Name, s.Digest)
+	}
+
+	if len(statement.Subject) == 0 {
+		t.Error("expected at least one subject in the provenance statement")
+	}
+
+	if signedTR.Status.Artifacts != nil {
```
waveywaves: This test can pass without producing any native artifact. The fixture writes only the legacy `IMAGE_URL`/`IMAGE_DIGEST` results, never writes `$(step.artifacts.path)`, and this test does not arrange for `enable-artifacts` to be enabled. Consequently, artifacts may be nil and skipped or non-nil but empty, while the legacy results satisfy the subject assertion. Please emit native input, build-output, and non-build-output artifacts and assert their exact URIs and digests in `resolvedDependencies`, subjects, and `byProducts`.

docs/slsa-provenance.md
```diff
@@ -417,6 +417,61 @@ spec:
...
+This is a native, structured way to declare build inputs and outputs using the
+[Tekton Artifacts API (TEP-0147)](https://github.com/tektoncd/community/blob/main/teps/0147-tekton-artifacts-phase1.md).
+This is an alternative to type hinting that provides richer, more explicit
+artifact declarations.
+### Prerequisites
+The Artifacts API is gated behind a feature flag in Tekton Pipelines: "enable-artifacts"
+### How it works:
+Instead of writing to specially named results, steps write a structured JSON file to $(step.artifacts.path). The Tekton Pipelines controller reads this file and populates status.artifacts and status.steps[].inputs/outputs on the TaskRun.
+
+Chains then maps these artifacts into the SLSA provenance.
```
waveywaves, line 430: Could we scope this documentation and the release note to `slsa/v2alpha4`? Native subject and byproduct handling is currently wired only into that formatter. Native inputs also reach v2alpha3 through the shared resolved-dependency code, but its outputs do not, and the default `in-toto` format and `slsa/v1` receive neither.

docs/slsa-provenance.md
```diff
@@ -417,6 +417,61 @@ spec:
...
+```yaml
+apiVersion: tekton.dev/v1
+kind: TaskRun
+metadata:
+  name: build-with-artifacts
+spec:
+  taskSpec:
+    steps:
+      - name: build-and-push
+        image: gcr.io/kaniko-project/executor
```
waveywaves, line 442: This example cannot run with the selected image: the Kaniko executor image contains neither `bash` nor `buildah`. Also, Buildah writes `--digestfile` as `algorithm:hex`, so placing `${DIGEST}` under the `sha256` key produces `sha256:sha256:...`. Please use a compatible image and strip the prefix, for example `${DIGEST#sha256:}`.

docs/slsa-provenance.md
```diff
@@ -417,6 +417,61 @@ spec:
...
+                  "uri": "quay.io/myorg/myapp",
+                  "digest": {"sha256": "${DIGEST}"}
+                }],
+                "buildOutput": true
+              }
+            ]
+          }
+          EOF
+```
+If a TaskRun uses both type hinting and native artifacts, Chains will include artifacts from both sources. Duplicates are removed automatically.
```
waveywaves, line 474: “Duplicates are removed automatically” is currently true only for subjects and resolved dependencies. OCI extraction duplicates step artifacts through aggregate and per-step status, while byproducts have no deduplication. Please qualify this statement or deduplicate every affected path.

pkg/chains/formats/slsa/v2alpha4/internal/taskrun/taskrun.go
ngelman1: ok added. not sure I understand why we need the fallback though..? if an artifact appears in a step, than it will always appear in `Status.Artifacts` , correct? in what scenario is Status.Artifacts` empty but we still have step level artifacts?

pkg/artifacts/signable.go
ngelman1: youre right.  this adds more noise than use, removing this.
native artifacts still map correctly into provenance subjects via `subjectsFromNativeArtifacts`

pkg/chains/formats/slsa/v2alpha4/internal/taskrun/taskrun.go
```diff
@@ -66,10 +67,45 @@ func ByProducts(tro *objects.TaskRunObjectV1) ([]*intoto.ResourceDescriptor, err
...
+			for _, val := range output.Values {
+				if val.Uri == "" || len(val.Digest) == 0 {
+					continue
+				}
+				digestSet := common.DigestSet{}
+				for algo, hex := range val.Digest {
+					digestSet[string(algo)] = hex
+				}
+				byProds = append(byProds, &intoto.ResourceDescriptor{
+					Name:   val.Uri,
```
waveywaves, line 92: Here `val.Uri` is stored as the name, leaving the URI field empty. Could we use `output.Name` for `Name` and `val.Uri` for `Uri`?

pkg/chains/formats/slsa/internal/resolved_dependencies/resolved_dependencies.go
```diff
@@ -239,6 +239,7 @@ func taskDependencies(ctx context.Context, opts ResolveOptions, tro *objects.Tas
 	}
 
 	mats = material.FromTaskParamsAndResults(ctx, tro)
+	mats = append(mats, material.FromNativeArtifactInputs(ctx, tro)...)
```
waveywaves, line 242: This adds native inputs only to TaskRun provenance. PipelineRun deep inspection now includes child outputs but not child inputs. Could we include the inputs there as well?

pkg/artifacts/signable.go
waveywaves: Addressed by removing native artifacts from OCI extraction. Thank you.

pkg/artifacts/signable.go
waveywaves: Addressed by reading the aggregate once, with a step-level fallback. Thank you.

pkg/chains/formats/slsa/v2alpha4/internal/taskrun/taskrun.go
waveywaves: Addressed by reading task-level and step-level outputs from the aggregate. Thank you.

test/e2e_test.go
ngelman1: Added a srcond step in `taskrun.json` to write native artifacts to `$(step.artifacts.path)` then enable the flag in the test and assert exact match in `subjects`, `resolvedDependencies`, and `byProducts`

pkg/chains/formats/slsa/internal/resolved_dependencies/resolved_dependencies.go
jkhelil: +1 Native artifact inputs are missing from PipelineRun resolvedDependencies when deep inspection is enabled. FromNativeArtifactInputs is only wired through taskDependencies() (TaskRun path), while child outputs are already picked up via taskrun.SubjectDigests / ByProducts. Could we also call material.FromNativeArtifactInputs(ctx, task) in the deep-inspection loop below (alongside FromStepActionsResults)?
`if slsaconfig.DeepInspectionEnabled && opts.WithStepActionsResults {
    execTasks := pro.GetExecutedTasks()
    for _, task := range execTasks {
        stepActionMat := material.FromStepActionsResults(ctx, task)
        resolvedDependencies = append(resolvedDependencies, ConvertMaterialsToResolvedDependencies(stepActionMat, InputResultName)...)
        nativeInputMat := material.FromNativeArtifactInputs(ctx, task)
        resolvedDependencies = append(resolvedDependencies, ConvertMaterialsToResolvedDependencies(nativeInputMat, InputResultName)...)
    }
}`

jkhelil (commented):
given [TEP-0192](https://github.com/tektoncd/community/pull/1298), I’m not sure we should invest in landing this as-is.

Why defer:

- TEP-0192 replaces the TEP-0147 alpha API ($(step.artifacts.path), buildOutput) with a declarative model (spec.artifacts, subject: true, entrypoint/init, pluggable storage).
- Much of this PR would need rework: docs, e2e fixtures, field semantics, and likely Chains integration design (see TEP-0192 “Integration with Tekton Chains”).
- End-to-end QA for artifacts in Chains is non-trivial (Pipelines enable-artifacts, TaskRun vs PipelineRun, deep inspection, format matrix), effort is hard to justify for an alpha API Tekton plans to supersede.

vdemeester:
@jkhelil good point 👍🏼

ngelman1:
so if I understand correctly we are closing this PR? so does all efforts other under [SRVKP-11142](https://redhat.atlassian.net/issues?filter=-1&selectedIssue=SRVKP-11142)  ?
if the end goal is artifacts api there is no reason to just hold
@jkhelil @vdemeester

PR #1851, opened 2026-08-04
fix(release): stop duplicating GitHub release title

jkhelil:
# Changes

The `create-draft-release` task's `github_release_oci.yaml` pin was
still on the commit before tektoncd/plumbing#3519, so this pipeline
never picked up the upstream fix for the duplicated, quoted release
title.

This PR:

* Bumps the pinned plumbing revision to `main`, which includes that
  fix.
* Drops this pipeline's own `Release ${VERSION}` fallback name in
  `prepare-draft-release`: the plumbing task only avoids the
  duplicated title when `release-name` is empty, so synthesizing a
  non-empty fallback here defeated the fix, producing titles like:

  ```
  Tekton Chains release v0.28.0 "Release v0.28.0"
  ```

* Wires up the new `draft` param (default `"false"`) introduced by
  that same plumbing PR, so releases are published directly instead
  of being left as an unpublished draft that a maintainer has to
  manually publish.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Fix duplicated GitHub release titles and publish releases directly instead of leaving them as unpublished drafts.
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.tekton/release-patch.yaml
+++ b/.tekton/release-patch.yaml
@@ -68,6 +68,8 @@ spec:
       value: ""
     - name: releaseName
       value: ""
+    - name: draft
+      value: "false"
   timeouts:
     pipeline: 3h0m0s
   workspaces:
--- a/.tekton/release.yaml
+++ b/.tekton/release.yaml
@@ -77,6 +77,8 @@ spec:
       value: ""
     - name: releaseName
       value: ""
+    - name: draft
+      value: "false"
   timeouts:
     pipeline: 3h0m0s
   workspaces:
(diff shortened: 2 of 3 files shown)


.tekton/release-patch.yaml
```diff
@@ -68,6 +68,8 @@ spec:
       value: ""
     - name: releaseName
       value: ""
+    - name: draft
+      value: "false"
```
infernus01, line 72: so does it mean releases are immediately visible to GitHub watchers? What if something goes wrong and we want to catch issues (wrong changelog, missing artifacts) before users see it?

infernus01 (approved):
/lgtm

PR #1868, opened 2026-08-10
fix(x509): support ed25519 private keys in the x509 signer

### LINKED ISSUE #1189: panic using ed25519 signing key
# Expected Behavior
Signing taksruns with ed25519 private key shoudl work

 openssl genpkey -algorithm ed25519 -out private.pem
# Actual Behavior
panic
`{"level":"info","ts":"2024-08-27T14:14:54.450Z","logger":"watcher","caller":"x509/x509.go:180","msg":"Found x509 key...","knative.dev/controller":"github.com.tektoncd.chains.pkg.reconciler.taskrun.Reconciler","knative.dev/kind":"tekton.dev.TaskRun","knative.dev/traceid":"589829b2-f182-466b-bc24-561ad0d06f9b","knative.dev/key":"default/pipelinerun-test-6-hello2"}
panic: interface conversion: interface {} is ed25519.PrivateKey, not *ecdsa.PrivateKey

goroutine 152 [running]:
github.com/tektoncd/chains/pkg/chains/signing/x509.x509Signer({0x3d32f48?, 0xc000fc3d40?}, {0xc000d76e00, 0x77, 0x200})
	github.com/tektoncd/chains/pkg/chains/signing/x509/x509.go:190 +0x1b8
github.com/tektoncd/chains/pkg/chains/signing/x509.NewSigner({_, _}, {_, _}, {{{{0x33f6b89, 0xd}, 0xc000fc3ce0, {0x33e6186, 0x4}, 0x0}, ...}, ...})
	github.com/tektoncd/chains/pkg/chains/signing/x509/x509.go:62 +0x205
github.com/tektoncd/chains/pkg/chains.allSigners({_, _}, {_, _}, {{{{0x33f6b89, 0xd}, 0xc000fc3ce0, {0x33e6186, 0x4}, 0x0}, ...}, ...})
	github.com/tektoncd/chains/pkg/chains/signing.go:73 +0x3b8
github.com/tektoncd/chains/pkg/chains.(*ObjectSigner).Sign(0xc000b94900, {0x3d32f48, 0xc000fc3d40}, {0x3d6bc88, 0xc00062a4a8})
	github.com/tektoncd/chains/pkg/chains/signing.go:128 +0x16d
github.com/tektoncd/chains/pkg/reconciler/taskrun.(*Reconciler).FinalizeKind(0xc00082aca0, {0x3d32f48, 0xc000fc3d40}, 0xc000a9c008)
	github.com/tektoncd/chains/pkg/reconciler/taskrun/taskrun.go:67 +0xe9
github.com/tektoncd/chains/pkg/reconciler/taskrun.(*Reconciler).ReconcileKind(0xc000fd18e8?, {0x3d32f48?, 0xc000fc3d40?}, 0x0?)
	github.com/tektoncd/chains/pkg/reconciler/taskrun/taskrun.go:45 +0x1d
github.com/tektoncd/chains/vendor/github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1/taskrun.(*reconcilerImpl).Reconcile(0xc0003b1e00, {0x3d32f48, 0xc000fc3c80}, {0xc00093c210, 0x21})
	github.com/tektoncd/chains/vendor/github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1/taskrun/reconciler.go:236 +0x4f2
github.com/tektoncd/chains/vendor/knative.dev/pkg/controller.(*Impl).processNextWorkItem(0xc00066bb00)
	github.com/tektoncd/chains/vendor/knative.dev/pkg/controller/controller.go:542 +0x508
github.com/tektoncd/chains/vendor/knative.dev/pkg/controller.(*Impl).RunContext.func3()
	github.com/tektoncd/chains/vendor/knative.dev/pkg/controller/controller.go:491 +0x59
created by github.com/tektoncd/chains/vendor/knative.dev/pkg/controller.(*Impl).RunContext in goroutine 146
	github.com/tektoncd/chains/vendor/knative.dev/pkg/controller/controller.go:489 +0x325
`

1. Install chains controller
2. create an ed25519 key pair and secret
 openssl genpkey -algorithm ed25519 -out x509.pem
 openssl pkey -in private.pem -pubout -out x509-pub.pem
 k -n tekton-chains create secret generic signing-secrets --from-file=x509.pem --from-file=x509-pub.pem
3. create a taskrun and check controller log

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.1", GitCommit:"4c9411232e10168d7b050c49a1b59f6df9d7ea4b", GitTreeState:"clean", BuildDate:"2023-04-14T13:14:41Z", GoVersion:"go1.20.3", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"28", GitVersion:"v1.28.0", GitCommit:"855e7c48de7388eb330da0f8d9d2394ee818fb8d", GitTreeState:"clean", BuildDate:"2023-08-15T21:24:51Z", GoVersion:"go1.20.7", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**
tkn version
Client version: 0.37.0
Chains version: v0.22.0
Pipeline version: v0.62.0

pujitha24:
# Changes

The x509 signer's docs (`docs/signing.md`) state that the private key stored
in the `x509.pem` secret may be of type `ed25519` or `ecdsa`, but
`x509Signer()` only ever handled `*ecdsa.PrivateKey`. Any other PKCS8 key
type — including ed25519 — was rejected with
`unsupported private key type ..., only ECDSA keys are supported` (this used
to be an unchecked type assertion that panicked the controller; a prior fix
turned that into a graceful error, but ed25519 keys still could not be used
to sign at all, even though the docs say they should work).

This replaces the single `pk.(*ecdsa.PrivateKey)` assertion with a type
switch that also handles `ed25519.PrivateKey` via
`signature.LoadED25519SignerVerifier` (from the already-vendored
`github.com/sigstore/sigstore/pkg/signature` package used elsewhere in this
file). Any other key type still falls through to the existing descriptive
error, now mentioning both supported types.

The `TestSigner_SignED25519` test, previously skipped with
`t.Skip("skip test until ed25519 signing is implemented")`, is un-skipped
and now passes. `TestNewSignerUnsupportedX509KeyType` (which exercises the
"unsupported key type" error path) was switched from an ed25519 key to a
freshly generated RSA key, since ed25519 is no longer unsupported.

This is scoped to the `x509.pem` load path only; the `cosign.key` path
(`cosignSigner`) already supports ed25519 via cosign's own key loader and is
unaffected.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The x509 signer now supports ed25519 private keys (in addition to ECDSA), matching the documented behavior. Previously, an ed25519 `x509.pem` signing key was rejected with an "unsupported private key type" error.
```

---
AI assistance: this change was drafted with Claude Code.

Fixes #1189

### DIFF
--- a/pkg/chains/signing/x509/x509.go
+++ b/pkg/chains/signing/x509/x509.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/ed25519"
 	cx509 "crypto/x509"
 	"encoding/json"
 	"encoding/pem"
@@ -190,11 +191,16 @@ func x509Signer(ctx context.Context, privateKey []byte) (*Signer, error) {
 	if err != nil {
 		return nil, err
 	}
-	ecdsaKey, ok := pk.(*ecdsa.PrivateKey)
-	if !ok {
-		return nil, fmt.Errorf("unsupported private key type %T, only ECDSA keys are supported", pk)
+
+	var signer signature.SignerVerifier
+	switch key := pk.(type) {
+	case *ecdsa.PrivateKey:
+		signer, err = signature.LoadECDSASignerVerifier(key, crypto.SHA256)
+	case ed25519.PrivateKey:
+		signer, err = signature.LoadED25519SignerVerifier(key)
+	default:
+		return nil, fmt.Errorf("unsupported private key type %T, only ECDSA and ed25519 keys are supported", pk)
 	}
-	signer, err := signature.LoadECDSASignerVerifier(ecdsaKey, crypto.SHA256)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/chains/signing/x509/x509_test.go
+++ b/pkg/chains/signing/x509/x509_test.go
@@ -51,6 +51,38 @@ const ed25519Priv = `-----BEGIN PRIVATE KEY-----
 MC4CAQAwBQYDK2VwBCIEIGQn0bJwshjwuVdnd/FylMk3Gvb89aGgH49bQpgzCY0n
 -----END PRIVATE KEY-----`
 
+// Generated with:
+// openssl genpkey -algorithm RSA -pkeyopt rsa_keygen_bits:2048 -out rsa_private.pem
+// openssl pkcs8 -topk8 -nocrypt -in rsa_private.pem
+const rsaPriv = `-----BEGIN PRIVATE KEY-----
+MIIEvAIBADANBgkqhkiG9w0BAQEFAASCBKYwggSiAgEAAoIBAQDb+UnAxxs/VcaD
+jd5PvjVqzfiiomvZqB25+ZnA8FIyIoCoUKwSvanNRHPTVxbNR5luxuhnOJtD88U7
+s31KHSyjshU9BGIxRP4qxhhk3toq8TnkWlttY4wLd0T/lZxsK1fTIjhvPBtJKMU+
+85Bpc9ZVXR93dQMjWa7prQIKhXVKpj9kQkTyf1sOSkxb0VOQPYdNLzwap+YDOCo0
+0GxYjApjZic6kOM5/+dvEwb0NeTt3sz/YoUjobX5b/RtwYVVM625vhJ9fhTZygG/
+oPbloOKOqf+qj47f6uynLC4VWkWsT7kK8S3VUYfNH4XrJQxs/98lAdh45KLtHKPw
+Pfquf9MtAgMBAAECggEASacpBjbEjUrTmqXMW4f1C8tmZlIa6XhsZ6JG1H7DDs1V
+pcXJL8c4jSXP4GIHHPnNynUoSLN/7Vs4XXqGR2QIV9EfYlxO4m9W6QyGC3RAuWMm
+vqpwdWqA8C/htvApvWAv2l5ZZglKm47eqGrWHjDugYuaJx3TTKlRMyW+CrbP5Iux
+kLR68t+NTVhZNKmqQiUSaUjnuWtumzRJY8oMsFytpo3J58CgETypa6n+OyiEQDn4
+Eco3MvxE7a37S0q4MMh2DrT+4f4q6Aj917i47u1DRRquczBcOxOrtlBDvP33a9G7
+yrH4QAoTFP8frwvf7dScHsP083EziccWDkAX6zkpYwKBgQD5KuPI1h/WDjKgSkvQ
+35ecgmPMeywgCvljDAGn0y1Jl/4Pm/KeJka17XsUrUhD/klZPP4HOZ6KXdaY4Jfv
+V1D2gjp8t8h16B1sK0VFzgPk/Jm3W7ozADpffbeUaXZru7tQmxsJkl8Xi/RheOxJ
+wO6WkZHLFx98IxC8JoF3pRWiZwKBgQDiAXapVkjlf+O0MyI7IwhW5lvHC/AKI+nd
+D+IfUi4KS6e9BjcAap+JjFVZs/5q5QXw9xSGcJnvbXutxcFw2hpsm9aDmiWVcf5F
++MxzaR+iMJYQviTgoyX+MPmFG0cZByV7teDIL1gECXKgK+EBsnLTgvtAmSd7Ovgg
+inBhm+tpSwKBgC/ogD2odhyZRECvqF4z75nHNFsnv7c1hPf3YgYbw5Rn5hCoQoEI
+CQaH7+ds3f080muXH5zSBlrCajWg0XXSix2qsoYybBfHloiq1Tnzv6nyq7emqmmN
+/KtJp9egY4WZZg28lPlFLIWBgm6PapdPwlAvEyJCgupCb8BNgw03L663AoGANIAS
+iJO6q1ViF+Io+YPR1B3/A+YKBNEC6o9d/9ifSVT5yjc/X6FlHhazXPsrBrnc/3Tm
+F7TgjXXpXRyrKwP/T2uEEV4ljOnGH4sEM2sgJhUTRyBkgKplkP7fd8Q2Z+H5GxvM
+87PLxmRLdFm9Ex/Y/LlYlFD/kujH6wc9w+7saLECgYBS7gyUzNIR8W1aOLSyDhuZ
+j8OliPAI7kZr5666L2QlTPEwXAcWJZtNbnsglOnLxJrcHQggfBZJzGROUbYlwVSq
+3OSli8BsO9UXcxqUZknKT24PB+OfBKmXLXcblj/170SRoA/LdjR1mLPqgmysSorU
+B5NAMHSj1ynwROhQdRnLUg==
+-----END PRIVATE KEY-----`
+
 // Generated with RS256 algorithm (required for cosign v2.6.0+)
 // openssl genrsa -out private.pem 2048
 // python3 -c "import jwt; import time; private_key = open('/tmp/private.pem').read(); payload = {'iat': int(time.time()), 'exp': int(time.time()) + 3600 * 24 * 365 * 10, 'iss': 'user123'}; print(jwt.encode(payload, private_key, algorithm='RS256'))"
@@ -218,7 +250,6 @@ func TestSigner_SignECDSA(t *testing.T) {
 }
 
 func TestSigner_SignED25519(t *testing.T) {
-	t.Skip("skip test until ed25519 signing is implemented")
 	ctx := logtesting.TestContextWithLogger(t)
 	d := t.TempDir()
 	p := filepath.Join(d, "x509.pem")
@@ -276,8 +307,8 @@ func TestNewSignerUnsupportedX509KeyType(t *testing.T) {
 	ctx := logtesting.TestContextWithLogger(t)
 	d := t.TempDir()
 	p := filepath.Join(d, "x509.pem")
-	// ed25519 keys are a valid PKCS8 key but not supported by the x509 signer.
-	if err := os.WriteFile(p, []byte(ed25519Priv), 0644); err != nil {
+	// RSA keys are a valid PKCS8 key but not supported by the x509 signer.
+	if err := os.WriteFile(p, []byte(rsaPriv), 0644); err != nil {
 		t.Fatal(err)
 	}
 
--- a/test/clients.go
+++ b/test/clients.go
@@ -29,6 +29,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/ed25519"
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/x509"
@@ -86,10 +87,11 @@ func newClients(t *testing.T, configPath, clusterName string) *clients {
 }
 
 type setupOpts struct {
-	useCosignSigner bool
-	registry        bool
-	kanikoTaskImage string
-	ns              string
+	useCosignSigner  bool
+	useEd25519Signer bool
+	registry         bool
+	kanikoTaskImage  string
+	ns               string
 }
 
 func setup(ctx context.Context, t *testing.T, opts setupOpts) (*clients, string, func()) {
@@ -197,7 +199,7 @@ func createRegistry(ctx context.Context, t *testing.T, namespace string, kubeCli
 }
 
 type secret struct {
-	x509priv   *signature.ECDSASignerVerifier
+	x509priv   signature.SignerVerifier
 	cosignPriv signature.SignerVerifier
 }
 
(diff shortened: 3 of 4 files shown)


anithapriyanatarajan:
@pujitha24 Thank you for this. Could you verify with `transparency.enabled=true` and confirm the rekor uploads and cosign verification work fine E2E for the artifact signature and Provenance attestation?

pujitha24:
Fair ask, but I don't have a live cluster with Fulcio/Rekor set up in my environment, so I haven't run this end-to-end with `transparency.enabled=true` — I don't want to claim that's verified when it isn't.

What I can say from reading the code: this diff only touches key loading in `x509Signer`. Once the key is loaded, both the ECDSA and the new ed25519 branch return a `signature.SignerVerifier` wrapped in the same `Signer` struct, and `rekor.go`'s upload path (`pkg/chains/rekor.go`) operates on that generic interface with no key-type branching, so it should exercise the identical rekor-upload/cosign-verify path that ECDSA already uses today. That's inference from the code, not something I've tested live.

If you can point me to a cluster/kind-e2e setup to run this against, or would rather I add an ed25519 case to the existing e2e suite, happy to do either — let me know which you'd prefer.

anithapriyanatarajan:
> Fair ask, but I don't have a live cluster with Fulcio/Rekor set up in my environment, so I haven't run this end-to-end with `transparency.enabled=true` — I don't want to claim that's verified when it isn't.
> 
> What I can say from reading the code: this diff only touches key loading in `x509Signer`. Once the key is loaded, both the ECDSA and the new ed25519 branch return a `signature.SignerVerifier` wrapped in the same `Signer` struct, and `rekor.go`'s upload path (`pkg/chains/rekor.go`) operates on that generic interface with no key-type branching, so it should exercise the identical rekor-upload/cosign-verify path that ECDSA already uses today. That's inference from the code, not something I've tested live.
> 
> If you can point me to a cluster/kind-e2e setup to run this against, or would rather I add an ed25519 case to the existing e2e suite, happy to do either — let me know which you'd prefer.

Summarizing below possible options:

For key-based signing with transparency, by default Chains uploads to the public Rekor (`https://rekor.sigstore.dev`, the default). So you could verify this with kind cluster and ko apply of your PR.

**1. Create an Ed25519 signing secret**
```bash
openssl genpkey -algorithm ED25519 -out ed25519.pem
kubectl create secret generic signing-secrets -n tekton-chains \
  --from-file=x509.pem=ed25519.pem --dry-run=client -o yaml | kubectl apply -f -
```

**2. Enable transparency (default URL is public Rekor)**
```bash
kubectl patch configmap chains-config -n tekton-chains -p '{"data":{
  "transparency.enabled":"true",
  "artifacts.taskrun.format":"slsa/v2alpha4",
  "artifacts.oci.format":"simplesigning"
}}'
```

**3. Run both cases and watch the controller log around the Rekor upload**
```bash
kubectl logs -n tekton-chains deploy/tekton-chains-controller -f | grep -i "rekor\|tlog\|transparency"
```
**Alternate option:** the repo's `TestRekor` in `test/e2e_test.go` already sets `transparency.enabled: "true"` for both formats. You could swap the signing secret to an Ed25519 key and run it.

pujitha24:
Thanks for the detailed steps. Unfortunately my environment doesn't have docker/kind/kubectl/ko available, so I can't actually stand up a cluster and run either the manual walkthrough or the e2e suite myself — I'd just be guessing at the outcome, which I don't want to do.

I could add an ed25519 variant to `TestRekor` in `test/e2e_test.go` (and the key generation in `test/clients.go`, which currently hardcodes ECDSA for `signing-secrets`) so the case exists for whoever has cluster access to run — but I wouldn't be able to run or verify it myself either. Would that be useful as a starting point, or would you rather run the manual verification on your end since you already have a cluster set up?

anithapriyanatarajan:
>Thanks for the detailed steps. Unfortunately my environment doesn't have docker/kind/kubectl/ko available, so I can't actually stand up a cluster and run either the manual walkthrough or the e2e suite myself — I'd just be guessing at the outcome, which I don't want to do.

@pujitha24  We do need E2E evidence. Per the [Tekton contribution standards](https://github.com/tektoncd/community/blob/main/standards.md#tests), changes like this need to be verified end-to-end, and that bar matters especially for this PR. Once we merge this we are effectively telling users this mode is supported. If it turns out not to work as expected, for example the user reports "I signed my artifact and now I can't verify it", which is a much worse failure for users than a slower review.

On the tooling: standing up a local environment is simple, and it's a one-time setup that will pay off for future contributions too. DEVELOPMENT.md walks through it, but roughly:

- Install `kind`, `ko`, and `kubectl` (all single-binary installs, no privileged infra needed)
- `kind create cluster` to get a local cluster
- Install Tekton Pipelines, then deploy your build of Chains with `ko apply -f config/`
- Run a TaskRun, then verify the resulting signature/attestation

Happy to help if you are stuck anywhere with these steps

What I am looking for on the PR before approving:

1. Confirmation the controller comes up cleanly with your change deployed
2. The manual walkthrough run against a real TaskRun, with the signing **and** verification output pasted in (this is the part that proves the mode actually works)
3. The E2E suite passing, or a note on which cases you couldn't cover and why

Once that's on the PR I'm happy to approve. Thank you.

pujitha24:
Since I still don't have kind/kubectl/ko in this sandbox, I can't produce the manual walkthrough or confirm the controller comes up cleanly — I don't want to fake that. What I could do is add a `taskrun-ed25519` case to `TestRekor` in `test/e2e_test.go` (commit 1035926), which swaps the `signing-secrets` x509.pem for an ed25519 key via a new `useEd25519Signer` option in `test/clients.go`, so the real e2e job exercises the new code path end to end — sign, Rekor upload, verify — the same way the existing ECDSA case does. That required widening `secret.x509priv` from `*signature.ECDSASignerVerifier` to the `signature.SignerVerifier` interface and switching two `.Public()` call sites to `.PublicKey()`.

This is item 3 from your list (E2E suite coverage), but items 1 and 2 (controller-up confirmation, pasted signing/verification output) still need someone with cluster access — I can't produce those here. If you're able to run the e2e suite (or the manual steps you outlined) against this branch, that would close the gap; otherwise let me know if there's another way you'd like this verified.

anithapriyanatarajan:
/hold 

Marking this on-hold to complete the test myself and report the status her

anithapriyanatarajan:
> /hold
> 
> Marking this on-hold to complete the test myself and report the status her

 ## Testing with kind cluster targeting default public rekor shows the following caveat: ed25519 keys fail Rekor upload

**Summary:** When `transparency.enabled: true`, an ed25519 key cannot sign everything Chains uploads to Rekor. Per run Chains signs two payloads — an **attestation** and an **image signature** — and ed25519 needs a *different* signing mode for each. A single signer can only do one mode, so one leg always fails and the run ends `signed=failed`.

### Background

- **Pure ed25519** signs the whole message.
- **ed25519ph** signs a hash of the message first. Same key, different mode.
- Chains picks the mode from the secret file it finds (`NewSigner`, `pkg/chains/signing/x509/x509.go` L56-66):
  - **`x509.pem`** → `x509Signer` → **pure** ed25519 (`LoadED25519SignerVerifier`, L200).
  - **`cosign.key`** → `cosignSigner` → `cosign.LoadPrivateKey(..., nil)` (L218), which defaults to **ed25519ph** (`GetDefaultLoadOptions`, cosign `pkg/cosign/keys.go` L290).
- Rekor stores the two payloads in different entry types: **dsse** (attestation) and **hashedrekord** (image signature). Each accepts only one ed25519 mode.

### What we tested (kind cluster, same ed25519 key)

| Secret file | Loaded as | Attestation (dsse) | Image sig (hashedrekord) |
|---|---|---|---|
| `x509.pem` | pure ed25519 | ✅ accepted | ❌ `unsupported hash algorithm: "SHA-256" not in [SHA-512]` |
| `cosign.key` | ed25519ph | ❌ `could not verify envelope: accepted signatures do not match threshold, Found: 0, Expected 1` | ✅ hash accepted |

Same key both runs (the `cosign.key` was the same ed25519 key re-encoded via `cosign import-key-pair`; identical public key). Either file leaves one leg failing → `signed=failed`. Image-sig result is identical in both OCI modes (`dsse` tag and `sigstore-bundle` referrer) — they share one upload path.

### Why

- **hashedrekord** requires **ed25519ph over SHA-512** for ed25519. The `x509.pem` path sends a pure signature, and the upload hash is hardcoded **SHA-256** — rejected.
- **dsse** verifies the envelope with a **pure** ed25519 verifier. The `cosign.key` path's ph signature yields 0 valid signatures — rejected.

### Where in code

- Mode chosen by secret file: `pkg/chains/signing/x509/x509.go` L56-66; pure loader at L200, cosign/ph path at L218.
- One signer signs both legs: `pkg/chains/signing.go` — `signer.SignMessage(...)` L197 (signer looked up once at L172-173, DSSE-wrapped for attestations at L178-185).
- Upload split by payload type: `pkg/chains/rekor.go` — dsse at L51 (`TLogUploadDSSEEnvelope`), hashedrekord at L54-58 (`sha256.New()` → `TLogUpload`).

### Fix direction

Not a one-liner: because a single signer signs both payloads, you cannot just change the upload hash. A real fix must select the ed25519 mode **per entry type**:

- **hashedrekord (image sig):** sign with **ed25519ph** and upload with **SHA-512** (`cosign.TLogUploadWithCustomHash`, already vendored).
- **dsse (attestation):** sign with **pure** ed25519 (current behaviour).

This means threading the mode through so the image-sig leg and the attestation leg use different signer options.

### Refs

- rekor#1945 — ed25519ph support on the Rekor server
- sigstore/sigstore#1595 — ed25519ph signer/verifier helpers (vendored)

anithapriyanatarajan:
> > /hold
> > Marking this on-hold to complete the test myself and report the status her
> 
> ## Testing with kind cluster targeting default public rekor shows the following caveat: ed25519 keys fail Rekor upload
> **Summary:** When `transparency.enabled: true`, an ed25519 key cannot sign everything Chains uploads to Rekor. Per run Chains signs two payloads — an **attestation** and an **image signature** — and ed25519 needs a _different_ signing mode for each. A single signer can only do one mode, so one leg always fails and the run ends `signed=failed`.
> 
> ### Background
> * **Pure ed25519** signs the whole message.
> * **ed25519ph** signs a hash of the message first. Same key, different mode.
> * Chains picks the mode from the secret file it finds (`NewSigner`, `pkg/chains/signing/x509/x509.go` L56-66):
>   
>   * **`x509.pem`** → `x509Signer` → **pure** ed25519 (`LoadED25519SignerVerifier`, L200).
>   * **`cosign.key`** → `cosignSigner` → `cosign.LoadPrivateKey(..., nil)` (L218), which defaults to **ed25519ph** (`GetDefaultLoadOptions`, cosign `pkg/cosign/keys.go` L290).
> * Rekor stores the two payloads in different entry types: **dsse** (attestation) and **hashedrekord** (image signature). Each accepts only one ed25519 mode.
> 
> ### What we tested (kind cluster, same ed25519 key)
> Secret file	Loaded as	Attestation (dsse)	Image sig (hashedrekord)
> `x509.pem`	pure ed25519	✅ accepted	❌ `unsupported hash algorithm: "SHA-256" not in [SHA-512]`
> `cosign.key`	ed25519ph	❌ `could not verify envelope: accepted signatures do not match threshold, Found: 0, Expected 1`	✅ hash accepted
> Same key both runs (the `cosign.key` was the same ed25519 key re-encoded via `cosign import-key-pair`; identical public key). Either file leaves one leg failing → `signed=failed`. Image-sig result is identical in both OCI modes (`dsse` tag and `sigstore-bundle` referrer) — they share one upload path.
> 
> ### Why
> * **hashedrekord** requires **ed25519ph over SHA-512** for ed25519. The `x509.pem` path sends a pure signature, and the upload hash is hardcoded **SHA-256** — rejected.
> * **dsse** verifies the envelope with a **pure** ed25519 verifier. The `cosign.key` path's ph signature yields 0 valid signatures — rejected.
> 
> ### Where in code
> * Mode chosen by secret file: `pkg/chains/signing/x509/x509.go` L56-66; pure loader at L200, cosign/ph path at L218.
> * One signer signs both legs: `pkg/chains/signing.go` — `signer.SignMessage(...)` L197 (signer looked up once at L172-173, DSSE-wrapped for attestations at L178-185).
> * Upload split by payload type: `pkg/chains/rekor.go` — dsse at L51 (`TLogUploadDSSEEnvelope`), hashedrekord at L54-58 (`sha256.New()` → `TLogUpload`).
> 
> ### Fix direction
> Not a one-liner: because a single signer signs both payloads, you cannot just change the upload hash. A real fix must select the ed25519 mode **per entry type**:
> 
> * **hashedrekord (image sig):** sign with **ed25519ph** and upload with **SHA-512** (`cosign.TLogUploadWithCustomHash`, already vendored).
> * **dsse (attestation):** sign with **pure** ed25519 (current behaviour).
> 
> This means threading the mode through so the image-sig leg and the attestation leg use different signer options.
> 
> ### Refs
> * rekor#1945 — ed25519ph support on the Rekor server
> * [Add support for Ed25519ph Signer/Verifier sigstore/sigstore#1595](https://github.com/sigstore/sigstore/pull/1595) — ed25519ph signer/verifier helpers (vendored)

@pujitha24 would you have the bandwidth to address this finding or shall I proceed to get this fixed. cc: @jkhelil

pujitha24:
Thank you for actually running this down on a cluster — that's exactly the kind of gap I couldn't have found without one.

I traced through why this happens and I think there's a wrinkle in the fix direction worth flagging before I touch the signing code. `signer.SignMessage()` produces the same bytes that go both to Rekor's hashedrekord entry and (for the image-signature leg) into the stored OCI signature. If I switch the default ed25519 signer for `x509.pem` to ed25519ph so hashedrekord upload succeeds, that also changes what gets stored as the image signature — and every public-key resolution path I can find in vendored cosign/sigstore (`cosign.LoadPublicKey` -> `VerifierForKeyRef` -> `signature.LoadVerifier`, in `vendor/github.com/sigstore/cosign/v2/pkg/signature/keys.go`) bottoms out in `signature.LoadVerifierWithOpts`, which only returns an ed25519ph verifier if the caller explicitly passes the `ED25519ph` option — nothing in that chain does. `cosign verify --key`'s actual CLI wiring isn't vendored here so I can't fully trace it, but it's documented as going through this same key-loading machinery, and I couldn't find any flag that flips it to ph mode. So an ed25519-signed image would upload to Rekor fine but plausibly fail standard `cosign verify --key` — trading the E2E failure you found for a different one, arguably worse since it's silent until someone tries to verify.

I don't have a cluster to confirm this myself, so I'd rather flag it than guess. Given that, would you rather I: (a) still implement the per-entry-type mode threading you outlined and you validate `cosign verify` against the resulting image signature, or (b) scope this PR down to what's actually verified working today — ed25519 for the DSSE/attestation leg only, and leave the image-signature leg (hashedrekord) documented as unsupported for ed25519 until the verify-side story is sorted out?

anithapriyanatarajan:
@pujitha24 Let's close the loop on this. I prefer to consider the implementation approach that I have outlined in my earlier comment. Please do the following in order along with the fix:

1. Stand up a kind cluster and verify E2E. Install kind, ko, kubectl (single-binary installs, no privileged infra), kind create cluster, install Tekton Pipelines, then ko apply -f config/ with this branch. Run a TaskRun with transparency.enabled=true and an ed25519 x509.pem signing secret, and paste back:
 
> (a) the controller coming up cleanly, 

> (b) the signing/attestation output, and 

> (c) cosign verify / Rekor upload result for both the attestation (dsse) and the image-signature (hashedrekord) legs.

2. Harden the tests. Based on what the run actually shows, add/adjust tests so the supported path is covered and the known ed25519+hashedrekord limitation is explicitly asserted or guarded, not left implicit.

3.  If your environment genuinely cannot run this  i.e. you're in a locked-down sandbox with no docker/kind/kubectl/ko and no way to install them, then don't keep iterating on inference. Recommend closing this PR, so it can be picked up by someone with cluster access rather than sitting in a review loop that can't be completed from your side.

Pick 1+2 or 3, but let's not stay in the "reasoning from the code without a live run" state any longer.

pujitha24:
You're right that we shouldn't keep iterating from inference. I checked: this sandbox has no docker, kind, kubectl, or ko, and while kubectl/kind/ko are simple binaries I could fetch, there's no container runtime available and no way to stand one up here without a VM-based workaround (Lima/QEMU) that's well outside what I can responsibly do in this environment. So per your option 3 — closing this out so it can be picked up by someone with real cluster access.

Thanks for actually tracking down the ed25519ph/hashedrekord split — that's a real gap in this approach that reasoning from the code alone wouldn't have caught.

PR #1882, opened 2026-08-12
feat(reconciler): filter TaskRun/PipelineRun by spec.managedBy

khrm:
Skip signing runs managed by external controllers by filtering on the spec.managedBy field at the informer level. Runs with nil, empty, or "tekton.dev/pipeline" managedBy are always accepted. Additional values can be allowed via the "filter.managed-by" key in chains-config.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


khrm (commented):
/kind feature

pkg/config/config.go
```diff
@@ -280,11 +291,12 @@ func defaultConfig() *Config {
 				TUFMirrorURL:     tuf.DefaultRemoteRoot,
 			},
 		},
-		Storage: StorageConfigs{OCI: OCIStorageConfig{
-			EncodingFormat: OCIEncodingFormatDSSE,
-		}, Grafeas: GrafeasConfig{
-			NoteHint: "This attestation note was generated by Tekton Chains",
-		},
+		Storage: StorageConfigs{
```
khrm, line 294: This change is due to gofumt.

khrm (commented):
@infernus01 Can you please review and merge this?

pkg/reconciler/taskrun/controller.go
```diff
@@ -53,35 +53,33 @@ func NewNamespacesScopedController(namespaces []string) func(ctx context.Context
...
-			// setup watches for the config names provided by client
-			cfgStore.WatchConfigs(cmw)
+			if err := storage.WatchBackends(ctx, watcherStop, tsSigner.Backends, cfg); err != nil {
+				logger.Error(err)
+			}
+		})
+		cfgStore.WatchConfigs(cmw)
 
+		impl := taskrunreconciler.NewImpl(ctx, c, func(_ *controller.Impl) controller.Options {
 			return controller.Options{
```
infernus01, line 83: The informer filter at line 94 (TaskRunInformerFilterFunc(namespaces, cfgStore)) protects against steady-state events, but it doesn't cover leader promotion. Here's what happens on every controller restart:

The generated Knative reconciler at vendor/.../taskrun/controller.go:66 initializes promoteFilterFunc as nil. When the controller wins leader election, PromoteFunc (line 71-91) calls lister.List(labels.Everything()) and enqueues every TaskRun — the if promoteFilterFunc != nil check at line 81 is skipped because this Options block doesn't set PromoteFilterFunc.

That means externally-managed TaskRuns reach Reconcile, where setFinalizerIfFinalizer (reconciler.go:248) adds the chains.tekton.dev/taskrun finalizer via SSA. From that point on, the informer filter blocks all future events for that run — including deletion. FinalizeKind never fires, the finalizer is never removed, and the object is stuck in Terminating forever.

The framework already provides the escape hatch — PromoteFilterFunc in controller.Options (options.go:61-64):
   return controller.Options{
       SkipStatusUpdates:               true,
       ConfigStore:                     cfgStore,
       FinalizerName:                   "chains.tekton.dev/taskrun",
       UseServerSideApplyForFinalizers: true,
       FinalizerFieldManager:           "tekton-chains-controller/finalizers",
       PromoteFilterFunc:               reconciler.TaskRunInformerFilterFunc(namespaces, cfgStore),
   }

Same fix needed in pipelinerun/controller.go line 86 — add PromoteFilterFunc: reconciler.PipelineRunInformerFilterFunc(namespaces, cfgStore). The pipelinerun controller's second informer (taskrun with ownership, line 106) is also affected since it shares the same impl.

 ---Review by Claude Code (Opus 4.6)

pkg/reconciler/filter.go
```diff
@@ -14,62 +14,79 @@ limitations under the License.
...
+// and list of namespaces membership
+func PipelineRunInformerFilterFunc(namespaces []string, cfgStore *config.ConfigStore) func(obj interface{}) bool {
 	return func(obj interface{}) bool {
-		// Namespace filter
+		pr, ok := obj.(*v1.PipelineRun)
+		if !ok {
+			return false
+		}
+		if !isManagedByAllowed(pr.Spec.ManagedBy, cfgStore.Load().Filter.ManagedByValues) {
+			return false
```
infernus01, line 43: Should we add some debug-level log here?

pkg/reconciler/taskrun/controller.go
khrm: I know about this. And I deliberately didn't add that. It's a pre-existing behavior. I would raise separate pr for this. Since it's a bug effecting namespace filter also.

pkg/reconciler/filter.go
khrm: There's no context available to fetch log. We don't do that for existing namespace filters, so I think it's fine to do that.

pkg/reconciler/taskrun/controller.go
infernus01: But before this PR, there were no externally-managed runs being filtered, so the PromoteFunc bypass never caused stuck finalizers. This PR creates the scenario where a run passes the promote path but then gets permanently blocked by the informer filter. So am I getting it wrong that this kinda bug is introduced in this PR only.

khrm: Actually, there was a filter. The namespace filter was already there. It had the same bug. Only reason, I am didn't include promotefunc is that I need to write e2e for that also.

infernus01 (commented):
/lgtm

PR #1883, opened 2026-08-13
fix(storage): drop imagePullSecrets from OCI auth

### LINKED ISSUE #1336: Chains controller uses `imagePullSecrets` for artifact upload, causing failure due to read-only registry access
# Expected Behavior

If the ServiceAccount used in the TaskRun or PipelineRun has both `imagePullSecrets` and `secrets`, chains controller is expected to use the mounted `Secrets` which would have push permission to the given registry rather than relying on `imagePullSecrets` which will have only READ permission to the targeted oci registry for signed artifact upload.

# Actual Behavior

If the ServiceAccount used in the TaskRun or PipelineRun has both `imagePullSecrets` and `secrets`, chains controller is always using the `imagePullSecrets` to upload the signed artifacts resulting in AUTH error.

1. Install pipelines,chains in a kind cluster and complete config and execute a TaskRun following the script: https://gist.github.com/anithapriyanatarajan/ffa14efcc0274f7c22ec599df092ba21
2. verify that the annotation `chains.tekton.dev/signed: "false"` on TR  after 3 retries

4. Remove the imagePullSecret attached the serviceAccount and start the TaskRun again.
    `kubectl patch serviceaccount $SERVICE_ACCOUNT_NAME -p '{"imagePullSecrets": null}' -n $NAMESPACE`
6. verify that the annotation `chains.tekton.dev/signed: "true"`

# Additional Info
- Kubernetes version:

 ```
Client Version: v1.31.1
Kustomize Version: v5.4.2
Server Version: v1.31.0 
```

- Tekton Pipeline version:
```
pipeline version: 0.70.0
```
- Tekton Chain version:
```
chains version: 0.24.0
```
# Relevance to another issue

https://github.com/tektoncd/chains/issues/700 - As part of this we have enabled usage of `secrets` from service Account. 

https://github.com/lcarva/chains/blob/843c6b3477fee1154093d5a8660630f3ebdaef31/pkg/chains/storage/oci/oci.go#L58-L60 - here we do not seem to specify which secret to use for oci storage in options and hence as per the library both imagePullSecrets and Secrets are consolidated into a single slice and controller repeatedly fails always using the imagePullSecrets. This issue will be evident only if the oci registry is same for pull and push but the auth secret for pull is configured in `imagePullSecrets` and that for push is kept in `secrets`  patched to the service account used by the chains controller

@jkhelil - Thankyou for detailed inputs regarding the library details that is causing this concern.

@lcarva - could you suggest a way forward for this issue? Is it fine to explicity include logic in oci backend storage to use only `secrets` but never `imagePullSecrets` for signed artifact upload.

### PR DESCRIPTION
**Pull request by pujitha24**
# Changes

When a TaskRun/PipelineRun's ServiceAccount has both `imagePullSecrets`
and mounted `secrets` targeting the same OCI registry, the vendored
go-containerregistry keychain consolidates both sets and returns the
first match, which is always the read-only `imagePullSecrets`
credential. The OCI storage backend (`pkg/chains/storage/oci/legacy.go`,
still the active auth path wired in `pkg/chains/storage/storage.go`
despite its "Deprecated" doc-comment on the `Backend` type) only ever
pushes signed attestations, so resolving a read-only credential makes
the push fail with an auth error even though a push-capable credential
is available in `secrets`.

This addresses the ServiceAccount-level case described in the issue;
the PodTemplate-level case was already fixed in a prior, separate
change.

**Approach**
- Bump the vendored `github.com/google/go-containerregistry`
  `pkg/authn/kubernetes` module to the commit that merged upstream's
  `IgnorePullSecrets` option
  (https://github.com/google/go-containerregistry/pull/2315), which
  lets the keychain skip both the ServiceAccount's `imagePullSecrets`
  and any explicit pull secrets.
- Set `IgnorePullSecrets: true` when building the k8schain options used
  by the OCI backend, since this keychain is only ever used to push,
  never to pull.

**Validation**
- `go build ./...` passes.
- `make test-unit` passes for all packages.
- `make golangci-lint PKG=./pkg/chains/storage/oci/...` reports 0 issues.
- Added `TestK8schainOptions_IgnoresImagePullSecrets`, confirmed to fail
  if `IgnorePullSecrets: true` is removed and to pass with the fix,
  asserting the OCI backend's k8schain options are built with
  `IgnorePullSecrets` set.
- Not run: no live cluster is available in this environment, so the fix
  was not reproduced end-to-end against a real registry. Correctness
  instead relies on the upstream library's own test coverage for
  `IgnorePullSecrets` plus the added unit test confirming this backend
  threads the option through.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
When 'oci' is configured as the storage option for signed artifacts,
the tekton-chains-controller no longer uses the ServiceAccount's
`imagePullSecrets` to authenticate with the target OCI registry when
that ServiceAccount also has push-capable `secrets` for the same
registry.
```

---
AI assistance: this change was drafted with Claude Code.

Fixes #1336

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -259,7 +259,7 @@ require (
 	github.com/google/cel-go v0.29.2 // indirect
 	github.com/google/certificate-transparency-go v1.3.3 // indirect
 	github.com/google/gnostic-models v0.7.0 // indirect
-	github.com/google/go-containerregistry/pkg/authn/kubernetes v0.0.0-20250225234217-098045d5e61f // indirect
+	github.com/google/go-containerregistry/pkg/authn/kubernetes v0.0.0-20260602200943-e4a78951c3eb // indirect
 	github.com/google/go-github/v73 v73.0.0 // indirect
 	github.com/google/go-querystring v1.1.0 // indirect
 	github.com/google/licenseclassifier v0.0.0-20210722185704-3043a050f148 // indirect
--- a/go.sum
+++ b/go.sum
@@ -701,8 +701,8 @@ github.com/google/go-containerregistry v0.21.9 h1:F+D4uZ3iA3DLMJLfhaqMdHJbzeqm/2
 github.com/google/go-containerregistry v0.21.9/go.mod h1:dP5XNKcL7kMFF/TB3LfvWmVhAcv7iqkHb3oDK8aauTo=
 github.com/google/go-containerregistry/pkg/authn/k8schain v0.0.0-20260414223304-7a662782a11f h1:31syNEUHTz7kt9AmH9SyXgV+Y6LMd62ZjaEcPWCy9z0=
 github.com/google/go-containerregistry/pkg/authn/k8schain v0.0.0-20260414223304-7a662782a11f/go.mod h1:iMEl9wsO8BHbAcfXcIFCN03G1odxdJXkdAaskB4pHxg=
-github.com/google/go-containerregistry/pkg/authn/kubernetes v0.0.0-20250225234217-098045d5e61f h1:GJRzEBoJv/A/E7JbTekq1Q0jFtAfY7TIxUFAK89Mmic=
-github.com/google/go-containerregistry/pkg/authn/kubernetes v0.0.0-20250225234217-098045d5e61f/go.mod h1:ZT74/OE6eosKneM9/LQItNxIMBV6CI5S46EXAnvkTBI=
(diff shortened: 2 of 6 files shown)


pujitha24:
/assign @wlynch

This has been approved and is green on CI — assigning for an lgtm when you have a moment.

infernus01 (commented):
/lgtm

PR #1967, opened 2026-09-16
chore: group sigstore dependency updates in dependabot

jkhelil:
## Summary
- Group `github.com/sigstore/sigstore` and its KMS submodules in Dependabot so a single sigstore release produces one PR per branch instead of five.
- Rename the gomod catch-all group to `ungrouped` so it sorts after `sigstore` in generated `dependabot.yml` (required for Dependabot first-match grouping).

Mirrors https://github.com/tektoncd/pipeline/pull/10761.

## Test plan
- [x] `./hack/generate-dependabot.sh`
- [ ] Verify Dependabot opens a single grouped PR on the next sigstore release


infernus01 (commented):
/lgtm

